PR #59, opened 2018-10-13
Feature: Threshold for line charts

JohnPeng47:
### Updates
- added the capability to add user provided thresholds to line graphs
- added the capability to specify circle border colors for line graph points lying inside the threshold ranges

### Future Work
- mouseout event reverts circle stroke color back to default color 

- [ ] Demos all features
- [ ] Documented/annotated
- [ ] Matches UI/UX specs
- [ ] Accessible


packages/core/src/base-axis-chart.ts
```diff
@@ -421,17 +422,21 @@ export class BaseAxisChart extends BaseChart {
...
 		this.cleanGrid(g);
 		this.fillTickThresholds(g);
 	}
 
 	fillTickThresholds(yGrid) {
 		const { thresholds } = this.options;
+		if (typeof thresholds === "undefined") {
+			return;
+		}
+
```
theiliad: This would probably be more effective going into the function above where you're calling `fillTickThresholds`

```js
if (this.options.thresholds) {
  this.fillTickThresholds(g);
}
```

theiliad: Also, we should add a few checks to make sure the `thresholds` config provided is formatted properly or does not contain bad values. If so, we can use `console.error` to let the developer know

JohnPeng47: You said that we're gonna be provided a set of colors instead of specifying them in options.thresholds right? Will options.thresholds be a flat array of numbers then?

theiliad: This could be the format:
image

theiliad: Although, I think style would make more sense than color

zvonimirfras: Or type.

theiliad: Good point Z

JohnPeng47:
Sorry for the large number of commits i kinda messed up but the actual changes are in this one:
https://github.com/IBM/carbon-charts/pull/59/commits/17da6d9d971585f47c5eadc9f03ee5412203f88b

packages/core/src/base-axis-chart.ts
```diff
@@ -433,31 +434,86 @@ export class BaseAxisChart extends BaseChart {
...
+		let prevBase = this.y(0);
 
+		const width = yGrid.node().getBBox().width;
+		console.log(thresholds);
 		yGrid.selectAll(".tick")
 			.each(function(d, i) {
 				const y = parseFloat(select(this).attr("transform")
 					.replace(")", "")
 					.split(",")[1]
 				);
```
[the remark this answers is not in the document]
theiliad: I'm talking about
```js
const y = parseFloat(select(this).attr("transform")
	.replace(")", "")
	.split(",")[1]
);
```

It's very unreadable, if it does something that can be re-used or complex, we can move it into Tools.ts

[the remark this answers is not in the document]
JohnPeng47: Ok noted I put this into a function in the latest commit, should be more readable now

PR #81, opened 2018-11-08
feat($core): adds support for tension, beta, alpha in curves

### LINKED ISSUE #39: Support d3 curves' beta, alpha & tensions
Currently we support all of the D3 curves, except for the `beta`, `alpha` & `tension` parameters for 3 of the curves.

image

t-mullen:
Adds support for D3 curve modification methods like tension, alpha and
beta. Re-done because of the broken Git forks.

Closes #39

### Updates

- Adds in support for all D3 curve modifying methods (`.tension()`, `.alpha()`, `.beta()`, any any other similar methods).

The demo `curveNatural` has been replaced with the `curveCardinal` with a modified tension (it is visually identical.)

- [ ] Demos all features
- [ ] Documented/annotated
- [ ] Matches UI/UX specs
- [ ] Accessible

### DIFF
--- a/packages/core/demo/demo-data/line.ts
+++ b/packages/core/demo/demo-data/line.ts
@@ -55,7 +55,9 @@ export const curvedLineOptions = {
 			}
 		}
 	},
-	curve: "curveNatural",
+	curve: {
+		name: "curveNatural"
+	},
 	legendClickable: true,
 	containerResizable: true
 };
--- a/packages/core/src/line-chart.ts
+++ b/packages/core/src/line-chart.ts
@@ -51,11 +51,23 @@ export class LineChart extends BaseAxisChart {
 
 		this.innerWrap.attr("transform", `translate(${margins.left}, ${margins.top})`);
 
+		let curveName;
+		let curveOptions;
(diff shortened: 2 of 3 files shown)


t-mullen:
The idea to support the dual `curve` option instead of of two options gets pretty weird with the typings. I'm not sure if it's the best call.

theiliad:
I think we need to have a discussion here.

**There are 2 ways to do this:**
1.
image

2.
image
image

@t-mullen @zvonimirfras @cal-smith which approach looks more ideal to you?

Aside from that the implementation works great, although I would suggest we throw a `console.error` when user enters curve options that don't exist.

cal-smith:
I think 2 overall ... it's a little more complex, but it does a better job of keeping everything localized.

That said, does the rest of the library use `x` + `xOptions`? Or would the plan be to refactor any of those instances to the second style as well?

theiliad:
@cal-smith I agree about approach number 2. Nope, usually we don't do `x` + `xOptions`.

theiliad:
Any updates here @t-mullen?

PR #84, opened 2018-11-09
Fix(core): Fix label colours on donut charts

### LINKED ISSUE #68: Pie & Donut Label Colors can be wrong
In Pie & Donut if some of your values in the new data provided are larger than before, they get re-ordered since Pie & Donut always show the larger slices first (which is expected).

Now a bug in that functionality shows up where when the order is changed, labels take on different colors than before. If a label is re-used within the **lifetime** of a chart, it should always show the same color it was initially assigned.

JohnPeng47:
On every chart re-render, color values gets mapped to data labels using d3.OrdinalScale.range().
Because donut chart reorders its label data due to sorting, the indices of the label changes while
the ordering of the colours remain static, resulting in labels without a fixed colour. So instead of
passing the sorted labels to be mapped to colours, instead I take a copy of the whole set of labesl
during chart initalization, and use this instead. On every data update, I also check for the
addition of new labels not present in the copy, and I push these to the copy array.

Closes #68


packages/core/src/base-chart.ts
```diff
@@ -162,6 +164,16 @@ export class BaseChart {
 			}
 		});
 
+		if (typeof this.fixedDataLabels === "undefined") {
```
cal-smith: I'd either flip this if statement around, or do `if(!this.fixedDataLabels)`, unless we need to check for specifically "undefined".

packages/core/src/base-chart.ts
```diff
@@ -41,6 +41,8 @@ export class BaseChart {
 		tooltips: null
 	};
 
+	fixedDataLabels;
```
theiliad: This should move up a few lines to:
```ts
	// Data
	data: any;
	displayData: any;
	fixedDataLabels;
```

PR #91, opened 2018-11-15
Combo Chart

JohnPeng47:
… chart; add tslint fix to core'

When combo charts initalizes its sub-charts, a line of code in the base-chart constructor removes
the "chart-svg" element drawn by the combo constructor. This affected the calculation of Yscale on
data updates, resulting in wrong scale values.

3

### Updates

- list
- out
- updates

- [ ] Demos all features
- [ ] Documented/annotated
- [ ] Matches UI/UX specs
- [ ] Accessible


packages/core/package.json
```diff
@@ -13,6 +13,7 @@
     "build:watch": "rm -rf dist && webpack -w --config webpack.build.js",
     "bundle:analyzer": "rm -rf dist && webpack --config webpack.build.js --env.bundleAnalyzer",
     "lint": "tslint -p tsconfig.json -c tslint.json",
+	"fix": "tslint --fix -p tsconfig.json -c tslint.json",
```
itsAlexNguyen: For style, can we tab this in to keep things consistent?

theiliad: Inconsistent spacing here

packages/core/src/base-chart.ts
```diff
@@ -53,7 +53,6 @@ export class BaseChart {
 		const {chartId, container} = this.setChartIDContainer();
 		this.container = container;
 		this.chartContainerID = chartId;
-
```
theiliad: Should still have a line break here for readability

packages/core/src/bar-chart.ts
```diff
@@ -193,7 +193,7 @@ export class BarChart extends BaseAxisChart {
 		this.dispatchEvent("update");
 	}
 
-	updateElements(animate: boolean, rect?: any, g?: any) {
+	updateElements(animate: boolean, rect?: any) {
```
theiliad: Is there a reason to keep `rect?: any`?

JohnPeng47: Potentially we could but I don't think it's necessary. The only time the rect parameter is used is when new data is bound to the bars, and a reference to this is passed into UpdateElements. So potentially, we could move this into UpdateElement, but then later on in the code, the same reference is used to define an exit transition, which we would also have to move into UpdateElement unless we want to keep 2 references to the bars with newly bound data.
`		const rect = g.selectAll("rect.bar")
			.data((d, index) => this.addLabelsToDataPoints(d, index));

		this.updateElements(true, rect);
                ....
		rect.exit()  // We would either have to move this inside UpdateElement as well
                                 // or keep two references to rect
			.transition(this.getDefaultTransition())
			.style("opacity", 0)
			.remove();
`

theiliad: ok, if we do sometimes pass in things for rect then we can keep it, but you don't need to pass in `null` for it, since it's an optional parameter

zvonimirfras:
There are conflicts so let's start with that :)

JohnPeng47:
there is actually still a bug with combo charts where if the update data/changing the legend behaves weirdly (ie. bars get rendered out misaligned and legend items disappear randomly).
Steps to reproduce in the demo: 
- tick off a legend item
- click update data a few 

Repeat above steps a couple of times and you should observe the unexpected behaviour, working on a fix today

theiliad:
@JohnPeng47 In your opinion is this an issue in `changeDemoData()` or the chart?

JohnPeng47:
I think the problem is with updateDisplayData being called for each sub chart in

JohnPeng47:
Umm I’m going to be busy for the next 1-2 days, and after that as well so
I’m not sure when I will get a chance to work on this

theiliad:
Closing this in favour of #275

PR #96, opened 2018-11-16
[WIP] Accessibility

licarijd:
### Updates

- createSlider function called on initial draw, attempts to append a slider element to each chart's action bar (see TODO to see how I'm getting the action bar's name - not good)

- Since the only this.options.type is line, 3 sliders are created in the line chart's access bar (there should only be one for each chart), and no sliders created in other charts such as step-line charts

- The sliders look ugly but I'll fix that eventually
screen shot 2018-11-16 at 1 43 12 pm

- [ ] Demos all features
- [ ] Documented/annotated
- [ ] Matches UI/UX specs
- [ ] Accessible

### DIFF
--- a/packages/core/src/bar-chart.ts
+++ b/packages/core/src/bar-chart.ts
@@ -106,7 +106,7 @@ export class BarChart extends BaseAxisChart {
 						.attr("fill", d => this.getFillScale()[d.datasetLabel](d.label))
 						.attr("stroke", d => this.options.accessibility ? this.colorScale[d.datasetLabel](d.label) : null)
 						.attr("stroke-width", Configuration.bars.default.strokeWidth)
-						.attr("stroke-opacity", d => this.options.accessibility ? 1 : 0)						
+						.attr("stroke-opacity", d => this.options.accessibility ? 1 : 0)
 						.attr("tabindex", 0);
 
 		// Hide the overlay
--- a/packages/core/src/bar-chart.ts
+++ b/packages/core/src/bar-chart.ts
@@ -106,7 +106,8 @@ export class BarChart extends BaseAxisChart {
 						.attr("fill", d => this.getFillScale()[d.datasetLabel](d.label))
 						.attr("stroke", d => this.options.accessibility ? this.colorScale[d.datasetLabel](d.label) : null)
 						.attr("stroke-width", Configuration.bars.default.strokeWidth)
-						.attr("stroke-opacity", d => this.options.accessibility ? 1 : 0);
+						.attr("stroke-opacity", d => this.options.accessibility ? 1 : 0)						
+						.attr("tabindex", 0);
 
 		// Hide the overlay
 		this.updateOverlay().hide();
--- a/packages/core/src/line-chart.ts
+++ b/packages/core/src/line-chart.ts
@@ -89,7 +89,8 @@ export class LineChart extends BaseAxisChart {
 				.attr("cx", d => this.x(d.label) + margins.left)
 				.attr("cy", d => this.y(d.value))
 				.attr("r", Configuration.charts.pointCircles.radius)
-				.attr("stroke", d => this.colorScale[d.datasetLabel](d.label));
+				.attr("stroke", d => this.colorScale[d.datasetLabel](d.label))
+				.attr("tabindex", 0);
 
 		// Hide the overlay
 		this.updateOverlay().hide();
--- a/packages/core/src/pie-chart.ts
+++ b/packages/core/src/pie-chart.ts
@@ -124,6 +124,7 @@ export class PieChart extends BaseChart {
 			.attr("stroke", d => this.colorScale[this.displayData.datasets[0].label](d.data.label))
 			.attr("stroke-width", Configuration.pie.default.strokeWidth)
 			.attr("stroke-opacity", d => this.options.accessibility ? 1 : 0)
+			.attr("tabindex", 0)
 			.each(function(d) { this._current = d; });
 
 		// Draw the slice labels


packages/core/src/base-axis-chart.ts
```diff
@@ -12,6 +12,10 @@ import { BaseChart } from "./base-chart";
 import * as Configuration from "./configuration";
 import { Tools } from "./tools";
 
+import { zoom } from "d3-zoom";
+import { drag } from "d3-drag";
+import { marcon } from "d3-marcon"
```
theiliad: Missing semicolon here

packages/core/src/base-axis-chart.ts
```diff
@@ -62,6 +66,8 @@ export class BaseAxisChart extends BaseChart {
 			this.drawXGrid();
 			this.drawYGrid();
 
+			this.createSlider();
```
theiliad: Function names should be more readable and understandable. Also, the function should only be called if a certain option exists in the chart configurations. e.g.:
```js
if (this.options.scales.zoom) {
  this.makeChartZoomable();
}
```

packages/core/src/base-axis-chart.ts
```diff
@@ -136,6 +142,107 @@ export class BaseAxisChart extends BaseChart {
 		}));
 	}
 
+	createSlider(){
+
```
theiliad: Empty line here

packages/core/src/base-axis-chart.ts
```diff
@@ -136,6 +142,107 @@ export class BaseAxisChart extends BaseChart {
 		}));
 	}
 
+	createSlider(){
```
theiliad: There should be a space after `)`

packages/core/src/base-axis-chart.ts
```diff
@@ -136,6 +142,107 @@ export class BaseAxisChart extends BaseChart {
...
+
+		//TODO get the name of the action bar for the chart
+		var divToInsert = "#actions-" + this.options.type
+		console.log(divToInsert)
+
```
theiliad: Use of `var` is not recommended. Use `const` or `let`. Preferably `const` everywhere

packages/core/src/base-axis-chart.ts
```diff
@@ -136,6 +142,107 @@ export class BaseAxisChart extends BaseChart {
...
+			axis_height = 20;
+
+		//insert the slider in the action bar element
+		var setup = marcon()
+				.width(window.innerWidth)
+				.height(handle_radius * 2 + slider_height + handle_padding + axis_height)
+				.left(handle_radius)
+				.right(handle_radius)
+				.bottom(axis_height)
+				.element(divToInsert);
```
theiliad: We should include some comments here as to what some of these lines mean, and what they do

packages/core/src/base-axis-chart.ts
```diff
@@ -136,6 +142,107 @@ export class BaseAxisChart extends BaseChart {
...
+		console.log(divToInsert)
+
+		var handle_radius = 50,
+			handle_start_val = 50,
+			handle_padding = 10,
+			slider_height = 5,
+			axis_height = 20;
+
+		//insert the slider in the action bar element
+		var setup = marcon()
```
theiliad: Better naming here needed

packages/core/src/base-axis-chart.ts
```diff
@@ -136,6 +142,107 @@ export class BaseAxisChart extends BaseChart {
...
+
+		//TODO get the name of the action bar for the chart
+		var divToInsert = "#actions-" + this.options.type
+		console.log(divToInsert)
+
+		var handle_radius = 50,
+			handle_start_val = 50,
+			handle_padding = 10,
+			slider_height = 5,
+			axis_height = 20;
```
theiliad: Any magic numbers and configurations should be added to `Configuration.ts` and loaded in from there

packages/core/src/base-axis-chart.ts
```diff
@@ -136,6 +142,107 @@ export class BaseAxisChart extends BaseChart {
...
+
+		// the pointer
+		range_svg.append("polygon")
+				.attr("class", "range-dragger range-pointer")
+				.attr("points", calcPointerPoints(handle_start_val))
+				.call(drag()
+					.on("drag", dragged)
+				);
+
+		function calcPointerPoints(handle_val){
```
theiliad: Better naming needed. Also for internal functions we should use this format:
```js
const myFunc = (myVar) => {
   ...
};
```

packages/core/src/base-axis-chart.ts
```diff
@@ -136,6 +142,107 @@ export class BaseAxisChart extends BaseChart {
...
+			var point_c = range_x(handle_val) + "," + (range_height - slider_height);
+			var point_a = (range_x(handle_val) - (handle_radius / 4)) + "," + (range_height - slider_height - handle_padding - (handle_radius / 10));
+			var point_b = (range_x(handle_val) + (handle_radius / 4)) + "," + (range_height - slider_height - handle_padding - (handle_radius / 10));
```
theiliad: Lots of magical and unreadable calculations here. There should be an easier way, but if after lots of investigation there didn't turn out to be 1, these line should be heavily commented

packages/core/src/base-axis-chart.ts
```diff
@@ -136,6 +142,107 @@ export class BaseAxisChart extends BaseChart {
...
+			return point_a + " " + point_b + " " + point_c;
```
theiliad:
```js
return `${point_a} ${point_b} ${point_c}`;
```

packages/core/src/base-axis-chart.ts
```diff
@@ -136,6 +142,107 @@ export class BaseAxisChart extends BaseChart {
...
+		}
+
+		function dragged(){
```
theiliad:
```js
const myFunc = (myVar) => {
   ...
};
```

packages/core/src/base-axis-chart.ts
```diff
@@ -136,6 +142,107 @@ export class BaseAxisChart extends BaseChart {
...
+
+			var coordinates = [0, 0];
+			coordinates = mouse(this);
+			var x = coordinates[0];
+			x = x > range_width ? range_width :
+				x < 0 ? 0 :
+				x;
```
theiliad: Unreadable code

packages/core/src/base-axis-chart.ts
theiliad: Also, the `d3-marcon` package doesn't seem to be a reputable package, or part of the `d3` community. Do we really need to use it?

packages/core/src/base-axis-chart.ts
licarijd: Won't let me use the 'this' reference ( coordinates = mouse(this) ).

Message: Argument of type 'this' is not assignable to parameter of type 'ContainerElement

PR #112, opened 2018-11-26
fix(react): Fix react wrappers resizing with animation

### LINKED ISSUE #111: Animated initial render in frameworks (e.g. React, angular etc.)
Suspect this is an issue with `requestAnimationFrame`, which should be replaced with a more efficient method such as `ResizeObserver`

theiliad:
Closes #111 

### Updates
- Add width & height props to **react** `BaseChart`
- Support width & height option values for `core` package

### DIFF
--- a/packages/core/src/base-chart.ts
+++ b/packages/core/src/base-chart.ts
@@ -51,25 +51,53 @@ export class BaseChart {
 	constructor(holder: Element, configs: any) {
 		this.id = `chart-${BaseChart.chartCount++}`;
 
-		(holder as HTMLElement).style.position = "relative";
+		if (configs.options) {
+			this.options = Object.assign({}, this.options, configs.options);
+		}
+
+		// Save holder element reference, and initialize it by applying appropriate styling
 		this.holder = holder;
+		this.styleHolderElement();
 
 		const {chartId, container} = this.setChartIDContainer();
 		this.container = container;
 		this.chartContainerID = chartId;
 
-		if (configs.options) {
-			this.options = Object.assign({}, this.options, configs.options);
-
-			if (this.options.containerResizable) {
-				this.resizeWhenContainerChange();
-			}
+		if (this.options && this.options.containerResizable) {
+			this.resizeWhenContainerChange();
 		}
 
 		this.events = document.createDocumentFragment();
 
 		if (configs.data) {
 			this.setData(configs.data);
+		}
+	}
+
+	styleHolderElement() {
+		const holderElement = this.holder as HTMLElement;
+		const { width, height } = this.options;
+
+		// If width exists in options
+		if (width) {
+			// Validate it's value
+			if (Tools.validateWidthHeightValues(width)) {
+				// Apply formatted width attribute to chart
+				holderElement.style.width = Tools.formatWidthHeightValues(width);
+			} else {
+				console.error("`width` is badly formatted");
+			}
+		}
+
+		// If height exists in options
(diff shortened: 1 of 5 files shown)


packages/core/src/base-chart.ts
```diff
@@ -51,19 +51,20 @@ export class BaseChart {
...
 		this.container = container;
 		this.chartContainerID = chartId;
 
-		if (configs.options) {
-			this.options = Object.assign({}, this.options, configs.options);
-
-			if (this.options.containerResizable) {
-				this.resizeWhenContainerChange();
-			}
+		if (this.options && this.options.containerResizable) {
```
zvonimirfras: Based on what I see here `this.options` should always exist and checking for it is redundant.

packages/core/src/base-chart.ts
```diff
@@ -73,6 +74,33 @@ export class BaseChart {
 		}
 	}
 
+	styleHolderElement() {
+		const holderElement = this.holder as HTMLElement;
+		const { width, height } = this.options;
+
+		// If width exists in options
+		if (width) {
+			// Validate it's value
```
zvonimirfras: Should this be
```suggestion
			// Validate its value
```

packages/core/src/base-chart.ts
```diff
@@ -73,6 +74,33 @@ export class BaseChart {
...
+				// Apply formatted width attribute to chart
+				holderElement.style.width = Tools.formatWidthHeightValues(width);
+			} else {
+				console.error("`width` is badly formatted");
+			}
+		}
+
+		// If height exists in options
+		if (height) {
+			// Validate it's value
```
zvonimirfras: Should this be
```suggestion
			// Validate its value
```

packages/core/src/base-chart.ts
```diff
@@ -73,6 +74,33 @@ export class BaseChart {
...
+	styleHolderElement() {
+		const holderElement = this.holder as HTMLElement;
+		const { width, height } = this.options;
+
+		// If width exists in options
+		if (width) {
+			// Validate it's value
+			if (Tools.validateWidthHeightValues(width)) {
+				// Apply formatted width attribute to chart
+				holderElement.style.width = Tools.formatWidthHeightValues(width);
```
zvonimirfras: I see you're trying to enforce using sizes in % or px but I'm having a hard time understanding why are those only viable options.

packages/react/src/base-chart.js
```diff
@@ -7,6 +7,20 @@ export default class BaseChart extends React.Component {
 		this.data = props.data;
 		this.options = props.options;
 
+		// Width prop is mandatory for the wrappers
```
zvonimirfras: Why are width and height "mandatory for the wrappers"?

packages/core/src/base-chart.ts
theiliad: Good catch 👍

packages/react/src/base-chart.js
theiliad: This is an issue with frameworks such as React & Angular. The chart holder does not start at its ideal size initially, and re-renders multiple times to get to the final chart size, an as a result of that, `resizeChart()` is called multiple times, making the initial render seem animated.
image

We can use `react-sizeme` solves this by initially rendering the component invisibly, then re-rendering to the full size, or another approach that **most** charting frameworks use in React is to make width & height mandatory.

PR #129, opened 2019-01-12
[WIP] Accessiblity

licarijd:
### Updates

- makeAccessiblity() in base-chart.ts called from each chart data component
- right now makeAccessiblity() just adds each element to the tabindex

- [ ] Demos all features
- [ ] Documented/annotated
- [ ] Matches UI/UX specs
- [ ] Accessible

### DIFF
--- a/packages/core/src/bar-chart.ts
+++ b/packages/core/src/bar-chart.ts
@@ -106,7 +106,7 @@ export class BarChart extends BaseAxisChart {
 						.attr("fill", d => this.getFillScale()[d.datasetLabel](d.label))
 						.attr("stroke", d => this.options.accessibility ? this.colorScale[d.datasetLabel](d.label) : null)
 						.attr("stroke-width", Configuration.bars.default.strokeWidth)
-						.attr("stroke-opacity", d => this.options.accessibility ? 1 : 0)						
+						.attr("stroke-opacity", d => this.options.accessibility ? 1 : 0)
 						.attr("tabindex", 0);
 
 		// Hide the overlay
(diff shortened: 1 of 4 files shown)


packages/core/src/base-chart.ts
```diff
@@ -951,4 +951,12 @@ export class BaseChart {
 	getBBox(selector: any) {
 		return this.innerWrap.select(selector).node().getBBox();
 	}
+
+	// ================================================================================
+	// Makes chart data components more screen reader friendly
+	// ================================================================================
```
theiliad: Good call with the comment 👍

packages/core/src/base-chart.ts
```diff
@@ -951,4 +951,12 @@ export class BaseChart {
...
+	makeAccessible (innerWrap: any) {
+		// Make chart data components tabbable
+		innerWrap.attr("tabindex", 0);
```
theiliad: This function should be able to handle all sorts of chart components. Keep in mind the d3 `.call` function passes in the selection, therefore you can use that info in this function.

packages/core/src/bar-chart.ts
```diff
@@ -106,7 +106,9 @@ export class BarChart extends BaseAxisChart {
 						.attr("fill", d => this.getFillScale()[d.datasetLabel](d.label))
 						.attr("stroke", d => this.options.accessibility ? this.colorScale[d.datasetLabel](d.label) : null)
 						.attr("stroke-width", Configuration.bars.default.strokeWidth)
-						.attr("stroke-opacity", d => this.options.accessibility ? 1 : 0);
+						.attr("stroke-opacity", d => this.options.accessibility ? 1 : 0)
+						.attr("aria-label", d => this.x1(d.datasetLabel))
```
theiliad: Would need to be move to `makeAccessible()`, although for bar it might have to stay. Needs investigating

licarijd:
Created an identical request from 'accessibility' branch

PR #130, opened 2019-01-13
Accessibility

licarijd:
### Updates

- tabindex each chart data element
- aria label each chart data element as "Label: [x value], Value: [yval], belongs to the dataset [dataset name]"
- mark every chart container as a live region with role=alert so users get notified when charts update

- [ ] Demos all features
- [ ] Documented/annotated
- [ ] Matches UI/UX specs
- [ ] Accessible


packages/core/src/bar-chart.ts
```diff
@@ -106,7 +106,11 @@ export class BarChart extends BaseAxisChart {
 						.attr("fill", d => this.getFillScale()[d.datasetLabel](d.label))
 						.attr("stroke", d => this.options.accessibility ? this.colorScale[d.datasetLabel](d.label) : null)
 						.attr("stroke-width", Configuration.bars.default.strokeWidth)
-						.attr("stroke-opacity", d => this.options.accessibility ? 1 : 0);
+						.attr("stroke-opacity", d => this.options.accessibility ? 1 : 0)
+						// Make the data element accessible and pass x and y values for the data point, and the dataset label
+						.call(this.makeAccessible,
+							(d, index) => `Label: ${d.label}, Value: ${d.value}, belongs to ${d.datasetLabel}`,
+							d => this.y(Math.max(0, d.value)));
```
theiliad: the above 3 lines are very hard to read/understand. Only `makeAccessible` should be called here, and `makeAccessible` should know what to do with the data passed in.

If separate logic is needed for `bar-chart`, then `makeAccessible` should be overriden in `bar-chart`.

packages/core/src/base-chart.ts
```diff
@@ -394,6 +394,8 @@ export class BaseChart {
 			chartId = this.id;
 			container = parent.append("div");
 			container.attr("chart-id", chartId)
+				// Alert user when the contents of the chart container changes
+				.attr("role", "alert")
```
theiliad: Not sure if this is the greatest choice. If multiple charts are on the page, then the user could potentially be receiving too many alerts. @zvonimirfras @cal-smith what do you think?

packages/core/src/base-chart.ts
```diff
@@ -951,4 +953,14 @@ export class BaseChart {
 	getBBox(selector: any) {
 		return this.innerWrap.select(selector).node().getBBox();
 	}
+
+	// ================================================================================
+	// Makes chart data components more screen reader friendly
+	// ================================================================================
+	makeAccessible (innerWrap: any, label?: any) {
+		// Make chart data components tabbable
+		innerWrap.attr("tabindex", 0)
```
theiliad: innerWrap should not be passed into `makeAccesible`. Can be accessed using `this.innerWrap`.

packages/core/src/base-chart.ts
```diff
@@ -951,4 +953,14 @@ export class BaseChart {
...
+		// Aria label contains the x value, y value, and dataset for the data element
+		.attr("aria-label", label);
```
theiliad: Same as my comment above, label should be generated in here.

packages/core/src/line-chart.ts
```diff
@@ -89,7 +89,10 @@ export class LineChart extends BaseAxisChart {
 				.attr("cx", d => this.x(d.label) + margins.left)
 				.attr("cy", d => this.y(d.value))
 				.attr("r", Configuration.charts.pointCircles.radius)
-				.attr("stroke", d => this.colorScale[d.datasetLabel](d.label));
+				.attr("stroke", d => this.colorScale[d.datasetLabel](d.label))
+				.call(this.makeAccessible, d => "Label: " + d.label
+						+ ", Value: " + d.value
+						+ ", belongs to " + d.datasetLabel, d => this.y(Math.max(0, d.value)));
```
theiliad: Still using string concatenation here.

packages/core/src/bar-chart.ts
licarijd: Just passing a string containing the label, value, and dataset into makeAccessible Separate logic isn't needed for bar chart, every chart will also be passing in a string with those values.

licarijd: Nevermind, I understand after seeing your comment below

theiliad: Ok great. This function is still hard to read though. If you just call `this.makeAccesible`, `d` will be passed into it, and you can do all of this in there

theiliad: ok

packages/core/src/base-chart.ts
licarijd: I don't have to pass innerWrap in, but leaving innerWrap?:any as a parameter does the trick

packages/core/src/base-chart.ts
```diff
@@ -951,4 +954,28 @@ export class BaseChart {
...
+
+	// ================================================================================
+	// Makes chart data components more screen reader friendly
+	// ================================================================================
+	makeAccessible (element?: any, type?: any, dataList?: any) {
+		// Make chart data components tabbable
+		element.attr("tabindex", 0);
+		// Aria label contains the x value, y value, and dataset for the data element
+		if (type === "axis") {
+			element.attr("aria-label", (d) => `Label: ${d.label}, Value: ${d.value}, belongs to ${d.datasetLabel}`);
```
zvonimirfras: How do we translate these strings?

zvonimirfras: We don't use optional parenthesis in arrow functions.

packages/core/src/base-chart.ts
cal-smith: [based on MDN](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_alert_role) this definitely wouldn't be a good use of `role="alert"`

theiliad: Great thanks Cal

PR #135, opened 2019-01-15
fix(core): Add y-axis title

licarijd:
… titles

### Updates

- Add a 'headers' element to scales, containing the chart's title
- Display headers.title at the top of every chart
- x axis title remains in place

- [ ] Demos all features
- [ ] Documented/annotated
- [ ] Matches UI/UX specs
- [ ] Accessible


packages/core/demo/demo-data/bar.ts
```diff
@@ -52,10 +52,14 @@ export const groupedBarData = {
 
 export const groupedBarOptions = {
 	scales: {
+		headers:{
+			title:  "2018 Annual Sales Figures"
+		},
 		x: {
-			title: "2018 Annual Sales Figures",
+			title: "Asset Category",
```
theiliad: Good call 👍

packages/core/demo/demo-data/bar.ts
```diff
@@ -52,10 +52,14 @@ export const groupedBarData = {
 
 export const groupedBarOptions = {
 	scales: {
+		headers:{
+			title:  "2018 Annual Sales Figures"
+		},
```
theiliad: `title` is really an option for the entire chart. doesn't have anything to do with the scales.

I would move `title` to the root layer of `groupedBarOptions`

packages/core/demo/demo-data/bar.ts
```diff
@@ -92,10 +96,14 @@ export const simpleBarData = {
 export const simpleBarOptions = {
 	accessibility: false,
 	scales: {
+		headers:{
+			title:  "2018 Annual Sales Figures"
+		},
```
theiliad: Same as above.

> `title` is really an option for the entire chart. doesn't have anything to do with the scales.
> 
> I would move `title` to the root layer of `groupedBarOptions`

packages/core/demo/demo-data/bar.ts
```diff
@@ -159,10 +167,14 @@ export const stackedBarData = {
 export const stackedBarOptions = {
 	accessibility: false,
 	scales: {
+		headers:{
+			title:  "2018 Annual Sales Figures"
+		},
```
theiliad: Same as above.

> `title` is really an option for the entire chart. doesn't have anything to do with the scales.
> 
> I would move `title` to the root layer of `groupedBarOptions`

packages/core/demo/demo-data/line.ts
```diff
@@ -42,10 +42,14 @@ export const curvedLineData = {
 export const curvedLineOptions = {
 	accessibility: false,
 	scales: {
+		headers:{
+			title:  "2018 Annual Sales Figures"
+		},
```
theiliad: Same as above.

> `title` is really an option for the entire chart. doesn't have anything to do with the scales.
> 
> I would move `title` to the root layer of `groupedBarOptions`

packages/core/demo/demo-data/line.ts
```diff
@@ -107,10 +111,14 @@ export const lineData = {
 export const lineOptions = {
 	accessibility: false,
 	scales: {
+		headers:{
+			title:  "2018 Annual Sales Figures"
+		},
```
theiliad: Same as above.

> `title` is really an option for the entire chart. doesn't have anything to do with the scales.
> 
> I would move `title` to the root layer of `groupedBarOptions`

packages/core/src/base-axis-chart.ts
```diff
@@ -244,6 +245,15 @@ export class BaseAxisChart extends BaseChart {
 				.text(this.options.scales.x.title);
 		}
 
+		// Add chart title
+		if (this.options.scales.headers.title) {
```
theiliad: Adding the chart title is not relevant to `setXAxis`. Should be moved into a different function/area.

packages/core/src/base-axis-chart.ts
```diff
@@ -244,6 +245,15 @@ export class BaseAxisChart extends BaseChart {
...
+			this.innerWrap.append("text")
+				.attr("class", "title")
+				.attr("text-anchor", "middle")
+				.attr("transform", `translate(${xAxisRef.node().getBBox().width / 2}, 0)`)
```
theiliad: According to the design specs, titles are aligned to the top-left side of the chart.

packages/core/src/base-axis-chart.ts
```diff
@@ -244,6 +245,15 @@ export class BaseAxisChart extends BaseChart {
...
+				.text(this.options.scales.headers.title);
```
theiliad: The chart would also need to be pushed down when user adds a title. Currently the chart stays in the same position. Design specs need to be reviewed for the aesthetic implementation.

packages/core/src/base-axis-chart.ts
```diff
@@ -54,6 +55,7 @@ export class BaseAxisChart extends BaseChart {
 			this.setXAxis();
 			this.setYScale();
 			this.setYAxis();
+			// this.setTitle();
```
theiliad: we can take this line out if we don't need it

packages/core/src/base-axis-chart.ts
```diff
@@ -176,6 +178,10 @@ export class BaseAxisChart extends BaseChart {
 			this.repositionXAxisTitle();
 		}
 
+		if (this.innerWrap.select(".axis-label.y").nodes().length > 0 && this.options.scales.y.title) {
+			this.repositionYAxisTitle();
```
theiliad: we should do an `||` here. In case you had no title, but you decide to add one with an update. Would a change as simple as that do the trick?

packages/core/src/base-axis-chart.ts
```diff
@@ -198,6 +204,16 @@ export class BaseAxisChart extends BaseChart {
...
+	/*setTitle() {
+		// Add chart title
+		if (this.options.title) {
+			this.innerWrap.append("text")
+				.attr("class", "title")
+				.attr("text-anchor", "left")
+				.text(this.options.title);
+		}
+	}*/
+
```
theiliad: Let's remove this if we don't need it

packages/core/src/base-axis-chart.ts
```diff
@@ -260,6 +276,18 @@ export class BaseAxisChart extends BaseChart {
...
 	}
 
+	repositionYAxisTitle() {
+		const yAxisRef = this.svg.select("g.y.axis");
+		const tickHeight = this.getLargestTickHeight(yAxisRef.selectAll(".tick"));
+
+		const yAxisTitleRef = this.svg.select("g.y.axis text.y.axis-label");
+		yAxisTitleRef.attr("class", "y axis-label")
+		.attr("text-align", "center")
+			.attr("transform", `rotate(-90) translate(${ - (yAxisRef.node().getBBox().height / 2)},
```
theiliad: Mismatched indentation on the 2 lines above

packages/core/src/base-axis-chart.ts
```diff
@@ -260,6 +276,18 @@ export class BaseAxisChart extends BaseChart {
...
+			${ - (tickHeight + Configuration.scales.tick.heightAddition) * 1.5})`)
```
theiliad: too much going on here, let's make this piece more readable

packages/core/src/base-axis-chart.ts
```diff
@@ -362,6 +390,18 @@ export class BaseAxisChart extends BaseChart {
...
+		// Add y-axis title
+		if (this.innerWrap.select(".axis-label.y").nodes().length === 0 && this.options.scales.y.title) {
+			yAxisRef.append("text")
+				.attr("class", "y axis-label")
+				.attr("transform", `rotate(-90) translate(${ - (yAxisRef.node().getBBox().height / 2)},
+				${ - (tickHeight + Configuration.scales.tick.heightAddition) * 1.5})`)
+				.attr("text-align", "center")
+				.text(this.options.scales.y.title);
+			this.svg.attr("transform", `translate(${this.innerWrap.select(".axis-label.y").node().getBBox().height}, 0)`);
+		}
```
theiliad: same comments here as the other instance of this code. Also can we reuse this somehow. We can probably use the d3 `.data()` to handle this

theiliad:
@zvonimirfras

packages/core/src/base-axis-chart.ts
```diff
@@ -45,6 +45,7 @@ export class BaseAxisChart extends BaseChart {
 
 		// If an axis exists
 		const xAxisRef = select(this.holder).select(".axis.x");
+		const yAxisRef = select(this.holder).select(".axis.y");
```
theiliad: Seems like this is also not used. Let's remove it then

packages/core/demo/demo-data/bar.ts
```diff
@@ -51,11 +51,13 @@ export const groupedBarData = {
 };
 
 export const groupedBarOptions = {
+	title:  "2018 Annual Sales Figures",
```
theiliad: If the PR doesn't include chart titles anymore, then we should get rid of these

packages/core/src/base-axis-chart.ts
licarijd: Sure, I will change the x axis title in that case as well since it is doing the same thing

packages/core/src/base-axis-chart.ts
licarijd: Sounds good, followed same convention as what's done for x axis

packages/core/src/base-axis-chart.ts
licarijd: Sounds good, followed same convention as what's done for x axis

packages/core/src/base-axis-chart.ts
```diff
@@ -362,6 +378,20 @@ export class BaseAxisChart extends BaseChart {
...
 
+		const tickHeight = this.getLargestTickHeight(yAxisRef.selectAll(".tick"));
+		const yAxisTitleXPosition = - (tickHeight + Configuration.scales.tick.heightAddition) * 1.5;
+
+		// Add y-axis title
+		if (this.innerWrap.select(".axis-label.y").nodes().length === 0 && this.options.scales.y.title) {
+			yAxisRef.append("text")
+				.attr("class", "y axis-label")
+				.attr("transform", `rotate(-90) translate(${ - (yAxisRef.node().getBBox().height / 2)},
+				${ yAxisTitleXPosition})`)
```
theiliad: same as above, also unnecessary _space_ here

packages/core/src/base-axis-chart.ts
```diff
@@ -260,6 +264,18 @@ export class BaseAxisChart extends BaseChart {
...
 	}
 
+	repositionYAxisTitle() {
+		const yAxisRef = this.svg.select("g.y.axis");
+		const tickHeight = this.getLargestTickHeight(yAxisRef.selectAll(".tick"));
+
+		const yAxisTitleRef = this.svg.select("g.y.axis text.y.axis-label");
+		yAxisTitleRef.attr("class", "y axis-label")
+		.attr("text-align", "center")
+		.attr("transform", `rotate(-90) translate(${ - (yAxisRef.node().getBBox().height / 2)},
```
theiliad: This should all be on the same line

licarijd: All of that will exceed that max line length

theiliad: which means that the logic should be broken down into different variables...

packages/core/src/base-axis-chart.ts
```diff
@@ -362,6 +383,23 @@ export class BaseAxisChart extends BaseChart {
...
 
+		const tickHeight = this.getLargestTickHeight(yAxisRef.selectAll(".tick"));
+
+		const yAxisTitleTranslate = {
+			x: - (yAxisRef.node().getBBox().height / 2),
+			y: - (tickHeight + Configuration.scales.tick.heightAddition) * 1.5
+		};
+
+		// Add y-axis title
+		if (this.innerWrap.select(".axis-label.y").nodes().length === 0 && this.options.scales.y.title) {
```
theiliad: wouldn't `this.innerWrap.select(".axis-label.y").nodes().length` always be equal to zero here?

PR #154, opened 2019-01-17
Fix x-axis alignment on resize 

### LINKED ISSUE #43: Line chart values do not perfectly align with x-axis in smaller dimensions
image

t-mullen:
### Updates

Fixes #43. Data points are always aligned with the x-axis now.

Previous code assumed left margin (or more precisely, the distance from the y-axis to the first tick) was always 40, which isn't true on resize. This uses half the tick value, which is correct. Also removes an extra padding.

screen shot 2019-01-17 at 3 55 56 pm

- [ ] Demos all features
- [ ] Documented/annotated
- [ ] Matches UI/UX specs
- [ ] Accessible


packages/core/src/base-axis-chart.ts
```diff
@@ -193,7 +193,7 @@ export class BaseAxisChart extends BaseChart {
 			const chartSize = this.getChartSize();
 			const width = chartSize.width - margins.left - margins.right;
 
-			this.x = scaleBand().rangeRound([0, width]).padding(Configuration.scales.x.padding);
```
theiliad: This seems to be breaking the spacing in barChart

packages/core/src/line-chart.ts
```diff
@@ -218,6 +218,11 @@ export class LineChart extends BaseAxisChart {
 		super.resizeChart();
 	}
 
+	setXScale () {
+		super.setXScale();
+		this.x.padding(0);
```
theiliad: This is good, but can we add a comment to reference back to baseAxis?

theiliad: And an empty line between the 2 calls, to distinguish

t-mullen: What kind of comment do you mean? Just one saying that super refers to BaseAxis?

theiliad: no a comment saying why we're overriding the padding, and that it is set in the `setXScale()` in BaseAxis

PR #155, opened 2019-01-17
Fix pie labels becoming misaligned when chart is scaled down

### LINKED ISSUE #151: Pie chart labels
image

When the Pie chart shrinks the labels don't draw in the correct order.

Related to #144/#145 ... There may be a design decision required here?

t-mullen:
### Updates

Fixes #151 

- Makes both the slices and labels use the same `computeRadius` function. Moves margining into the labels' `deriveTransformString` function.
- Replaces quadrant detection with a method that uses the angle and text size to evenly space all labels, regardless of label length or font size.

screen shot 2019-01-17 at 5 14 50 pm

- [ ] Demos all features
- [ ] Documented/annotated
- [ ] Matches UI/UX specs
- [ ] Accessible


packages/core/src/configuration.ts
```diff
@@ -193,8 +193,8 @@ export const pie = {
 	sliceLimit: 6,
 	label: {
 		dy: ".32em",
-		margin: 15,
-		other: "Other"
+		margin: 8,
+		other: "Other",
```
cal-smith: seems like we generally avoid trailing commas
```suggestion
		other: "Other"
```

PR #172, opened 2019-01-31
Fixes back-to-back updates not fully transitioning.

### LINKED ISSUE #37: Back to back transitions can result in unfinished transitions
image

Not too critical as you really need to quickly press things to reproduce the bug, and also considering that animations can be disabled in chart configs.

t-mullen:
### Updates
D3 transitions are canceled when a new transition is registered. The "opacity" attribute was being stopped mid-transition, leaving some elements transparent. The fix is to have the updateElements transition continue the opacity animation. 

Fixes #37

Also fixes an inconsistency where the SVG opacity attr was being used instead of the CSS opacity style (the latter is preferred). 

- [ ] Demos all features
- [ ] Documented/annotated
- [ ] Matches UI/UX specs
- [ ] Accessible


t-mullen:
I just realized using the opacity attribute causes some problems - will fix it.

t-mullen:
CSS style work everywhere - fixed.

theiliad:
@t-mullen While all other instances seem to have been fixed, there's still 1 bug left:
image

If you toggle `Dataset 1` off then on really quickly, it will disappear.

Is this related to transitions or a different bug?

Do we need a separate PR?

zvonimirfras:
Does the approval mean that the answer was offline yes for a separate PR and this is good to go?

cal-smith:
I'll echo @zvonimirfras though, is there a fix incoming in this PR or another for the one bug left?

t-mullen:
I'm still looking into the issue.

t-mullen:
The issue was elements were being scheduled for removal in a transition, then being reused before actually getting removed. The newest commit marks elements for removal with a class to make sure they aren't reused.

There's likely a better way to do transition scheduling with d3 - specifying every attribute for the end of every transition them feels hacky. But changing that would require a major refactor.

theiliad:
@t-mullen Great job on fixing this bug.

Could you please apply the requested changes so we can merge this asap and fix the experience for user? 🙂

theiliad:
Bumping this, is there an update?

theiliad:
@zvonimirfras @cal-smith

PR #187, opened 2019-02-08
Custom line circle radius

### LINKED ISSUE #150: Line chart - single continuous plot instead of points
#### I am submitting a...

- [X] Feature request
- [ ] Design defect
- [ ] Source code defect
- [ ] Other

#### charts version:
0.8.3
### Issue description
graph needs to have 30 plot points to indicate last 30 days (below does not indicate 30days, but hopefully illustrates the point). We need a single continuous plot instead of points.
a

This is what is desired:
b

theiliad:
Closes #150


packages/core/demo/demo-data/line.ts
```diff
@@ -130,6 +130,9 @@ export const lineOptions = {
 			]
 		}
 	},
+	points: {
```
zvonimirfras: We shouldn't replace our demo with this. We can have a separate demo if needed.

packages/core/src/configuration.ts
```diff
@@ -12,6 +12,10 @@ const baseOptions: any = {
 	tooltip: {
 		formatter: null
 	},
+	// Only used for line chart
```
zvonimirfras: They shouldn't be in baseOptions then.

packages/core/src/line-chart.ts
```diff
@@ -35,6 +35,15 @@ export class LineChart extends BaseAxisChart {
 		}));
 	}
 
+	getCircleRadius() {
+		return this.options.points.radius || Configuration.charts.pointCircles.radius;
```
zvonimirfras: Why different naming for the same thing in options and configuration? points vs pointCircles?

packages/core/demo/demo-data/line.ts
theiliad: Why wouldn't we? We allow custom radius of the points now, how else would someone find out?

zvonimirfras: Not that we shouldn't demo the functionality, we shouldn't _replace_ defaults with a custom value without making it obvious it's a customization.
Having an interactive circle size demo or a separate demo for the functionality.
Documentation.

PR #196, opened 2019-02-21
feat(bar): bar chart should support a max bar width

### LINKED ISSUE #139: Bar chart - Responsive mode
#### I am submitting a...

- [X] Feature request
- [ ] Design defect
- [ ] Source code defect
- [ ] Other

#### charts version:
0.8.3
### Issue description
**Expected:**  In responsive mode chart weight is expected to adjust to a widget width (expand along with the widget).  Expected to be aligned within the widget.
**Observed:** Component has fixed size and won't adjust as widget expands. Not aligned within widget.

--------------------------------------------------------------------------------------

**2.  Single bar width**
**Expected:** On the Neutrino site there are examples where charts have different bar width (consistent within the same chart).  Below is the example used as a reference for UI design.
**Observed:**  OMSI bars appear thicker than on the mockup. Single bar width changes in responsive mode (becomes even thicker). Our developer team says that bar appearance has to do with the component.

### LINKED ISSUE #140: Bar chart - Single bar width
#### I am submitting a...

- [X] Feature request
- [ ] Design defect
- [ ] Source code defect
- [ ] Other

#### charts version:
0.8.3

### Issue description
**Expected:** On the Neutrino site there are examples where charts have different bar width (consistent within the same chart).  Below is the example used as a reference for UI design.
**Observed:**  OMSI bars appear thicker than on the mockup. Single bar width changes in responsive mode (becomes even thicker). Our developer team says that bar appearance has to do with the component.

cal-smith:
closes #139 
closes #140 

Right now it's just applied to normal bars, but the process should be the same for stacked bars.

Grouped bars will be a bit of a different beast - I'm not sure this entirely applies to them anyway.

### DIFF
--- a/packages/core/demo/demo-data/bar.ts
+++ b/packages/core/demo/demo-data/bar.ts
@@ -108,7 +108,9 @@ export const simpleBarOptions = {
 	},
 	legendClickable: true,
 	containerResizable: true,
-	maxWidth: 50
+	bars: {
+		maxWidth: 50
+	}
 };
 
 // Stacked bar
--- a/packages/core/demo/demo-data/bar.ts
+++ b/packages/core/demo/demo-data/bar.ts
@@ -107,7 +107,8 @@ export const simpleBarOptions = {
 		}
 	},
 	legendClickable: true,
-	containerResizable: true
+	containerResizable: true,
+	maxWidth: 50
 };
 
 // Stacked bar
--- a/packages/core/src/configuration.ts
+++ b/packages/core/src/configuration.ts
@@ -167,7 +167,9 @@ export const bars = {
 		bars: 0.2,
 		datasets: 0.25
 	},
-	maxWidth: -1
+	bars: {
+		maxWidth: null
+	}
 };
 
 export const lines = {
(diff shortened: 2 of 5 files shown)


packages/core/demo/demo-data/bar.ts
```diff
@@ -107,7 +107,8 @@ export const simpleBarOptions = {
 		}
 	},
 	legendClickable: true,
-	containerResizable: true
+	containerResizable: true,
+	maxWidth: 50
```
theiliad: Is this for each bar?

This should become:
```js
{
   bars: {
	  maxWidth: 50
   }
}
```

packages/core/src/configuration.ts
```diff
@@ -166,7 +166,8 @@ export const bars = {
 	spacing: {
 		bars: 0.2,
 		datasets: 0.25
-	}
+	},
+	maxWidth: -1
```
theiliad: Why -1? Why not `null`?

cal-smith: Just needed an empty value, we can use `null` for sure 👍

packages/core/demo/demo-data/bar.ts
cal-smith: sure

cal-smith:
Part of #139 seems to be about some of the extra space reserved for the legend ... It's not immediately obvious to me how the chart is sized, so I'm going to consider that out of scope for this PR ... we should have a separate issue/PR for legend placement and chart sizing 👍

PR #199, opened 2019-02-26
feat(core): Keyboard accessible and screen reader friendly components

licarijd:
Aria labels for all chart data elements and legend elements, all chart data elements and legend
elements are added to the page's tabindex

### Updates
- list
- out
- updates

- [ ] Demos all features
- [ ] Documented/annotated
- [ ] Matches UI/UX specs
- [ ] Accessible


cal-smith:
Related #130

packages/core/src/base-chart.ts
```diff
@@ -951,4 +954,28 @@ export class BaseChart {
...
 		return this.innerWrap.select(selector).node().getBBox();
 	}
+// ================================================================================
+	// Makes chart data components more screen reader friendly
+	// ================================================================================
+	makeAccessible (element?: any, type?: any, dataList?: any) {
+		// Make chart data components tabbable
+		element.attr("tabindex", 0);
+		// Aria label contains the x value, y value, and dataset for the data element
+		if (type === "axis") {
```
theiliad: Is there any way we can find out this information just from `element`? Not too happy with passing strings into this function

packages/core/src/base-chart.ts
```diff
@@ -951,4 +954,28 @@ export class BaseChart {
...
+			element.attr("aria-label", (d) => `Label: ${d.label}, Value: ${d.value}, belongs to ${d.datasetLabel}`);
+		} else if (type === "slice") {
+			element.attr("aria-label", (d) => `Label: ${d.data.label}, Value: ${d.value},
+			percentage is ${Tools.convertValueToPercentage(d.data.value, dataList)} percent`);
```
theiliad: Seems like the presence of `dataList` means that `type === "slice"`, so that's a step towards removing the string being passed into this function

theiliad:
@zvonimirfras @cal-smith Please chime in for the comment above, if you have an opinion.

My thinking is that it's not ideal to pass strings into the `makeAccessible` functions, as the function is capable of finding out what sort of element it's dealing with, and in the comments for every conditional we can describe the type of element it'll correspond to...

cal-smith:
Yeah, if each element has some defining feature (classname, attribute, whatever) you could totally eliminate the `type` argument (or at least make it optional for the common usecase). It should be reasonably clear what each check is doing.

packages/core/src/base-chart.ts
```diff
@@ -967,4 +968,30 @@ export class BaseChart {
...
+			percentage is ${Tools.convertValueToPercentage(d.data.value, dataList)}`);
+		// Legend items
+		} else if (element.attr("class") === "legend-circle") {
+			element.attr("aria-label", (d) => `Legend Item: ${d.key}, Status: ${d.value ? "enabled" : "disabled"},
+			Click to ${d.value ? "disable" : "enable"}`);
+		// Chart containers
+		} else if (element.attr("chart-id")) {
+			element.attr("aria-label", `${element.attr("chart-id")}`);
+		// Legend containers
+		} else if (element.attr("class") === "legend right-legend") {
```
theiliad: This should be a better check. What if the legend is left-aligned?

licarijd:
> Almost there 💯
> 
> Also I'm still not able to toggle legend items when tabbing through them. Is that a WIP?

Yes actually I didn't consider toggling while tabbing but that is important. Should we use the 'Enter' key to toggle while a legend item is highighted?

theiliad:
> > Almost there 💯
> > Also I'm still not able to toggle legend items when tabbing through them. Is that a WIP?
> 
> Yes actually I didn't consider toggling while tabbing but that is important. Should we use the 'Enter' key to toggle while a legend item is highighted?

Yeah I think `Enter` is a good choice for now. Great 👍

packages/core/src/base-chart.ts
```diff
@@ -383,6 +383,31 @@ export class BaseChart {
 					c.selectAll(".legend-btn.not-allowed").classed("not-allowed", false);
 				}
 			});
+
+			select(this).on("keypress", function() {
```
theiliad: What does this piece do?

packages/core/src/base-chart.ts
```diff
@@ -954,4 +980,30 @@ export class BaseChart {
...
 		return this.innerWrap.select(selector).node().getBBox();
 	}
+// ================================================================================
+	// Makes chart data components more screen reader friendly
+	// ================================================================================
+	makeAccessible (element?: any, dataList?: any) {
+		// Make chart data components tabbable
+		element.attr("tabindex", 0);
+		// Aria label contains the x value, y value, and dataset for the data element
+		if (element.attr("class") === "bar" || element.attr("class") === "dot") {
```
theiliad: What if we decide to change the `class` for bars or dots down the road? We need a better way here to distinguish the elements

theiliad:
Closing this due to inactivity. Will revisit this topic soon

PR #201, opened 2019-02-27
Implements custom color functions.

### LINKED ISSUE #136: There should be an option to supply a custom color function
So colors for lines/bars/pie/etc can be determined based on the data or any other metric.

A potential signature could be:
```typescript
function(data: any, chart: any): string
```

Where `data` is the current data being displayed, and `chart` is either a reference to the chart instance, or a reference to the config (whichever is sufficiently powerful).

t-mullen:
### Updates
Adds two new function options `getStrokeColor` and `getFillColor`. These allow the developer to override colors based on dataset, label, or value.

Closes #136.

```javascript
new BarChart(container, {
  options: {
    getStrokeColor: function (datasetLabel, label, value) {
      return 'red'
      return false // return falsey values for default colors
    },
    getFillColor: function (datasetLabel, label, value) {
      return 'red'
      return false // return falsey values for default colors
    }
  }
})
```

Bar chart coloring bars red when value is over 35000:
screen shot 2019-02-27 at 12 39 58 pm

- [ ] Demos all features
- [ ] Documented/annotated
- [ ] Matches UI/UX specs
- [ ] Accessible


t-mullen:
This feature currently isn't accessible - there is no way to specify custom accessibility patterns in addition to custom colors.

packages/core/src/base-chart.ts
```diff
@@ -238,6 +238,22 @@ export class BaseChart {
 		}
 	}
 
+	getFillColor(datasetLabel: any, label?: any, value?: any) {
+		if (this.options.getFillColor && !this.options.accessibility) {
```
theiliad: what if accessibility is on?

t-mullen: Then this feature isn't available - using color to signify data is not accessible. We'll need a way to generate a unique pattern for each custom color.

PR #212, opened 2019-03-12
feat(core): specify config as interfaces

cal-smith:
- add interfaces for each chart types specific config
- add interface for the data format
- clean up unused variables


theiliad:
@cal-smith Code is fine, demos are broken:
image

theiliad:
Also, when resizing now, labels move

cal-smith:
Interesting! I'll take a closer look

packages/core/src/base-chart.ts
```diff
@@ -49,9 +49,9 @@ export class BaseChart {
 
 	constructor(holder: Element, configs: ChartConfig<BaseChartOptions>) {
 		this.id = `chart-${BaseChart.chartCount++}`;
-
+		console.log(this.options);
```
theiliad: probably just missed this

cal-smith: yep

cal-smith:
Nothing breaking, it just codifies the current API

theiliad:
Tooltips are broken now:
image

As well as border-opacity:
image

image

image

cal-smith:
My bad, `merge` wasn't checking for falsey values when deciding what to merge ... fixed 👍

packages/core/src/line-chart.ts
```diff
@@ -17,6 +17,15 @@ export class LineChart extends ScatterChart {
 	constructor(holder: Element, configs: ChartConfig<LineChartOptions>) {
 		super(holder, configs);
 
+		console.log(configs.options, Configuration.options.LINE);
```
theiliad: 🙂

cal-smith: I'll get it right sooner or later 😬

PR #219, opened 2019-03-21
Fix Vue build break

moores2:
### Fix Vue build break
- index.js for Vue wrappers has wrong context root
- eslint has problems locating config file when lerna creates symlinks (not fully understood) 

- [ ] Check netlify build no longer shows:
```
2:31:52 PM:  ERROR  Failed to compile with 1 errors14:31:52
2:31:52 PM: This relative module was not found:
2:31:52 PM: * ./components/ in ./src/index.js
```

### DIFF
--- a/.travis/after_success.sh
+++ b/.travis/after_success.sh
@@ -3,6 +3,9 @@ set -e
 
 npm config set //registry.npmjs.org/:_authToken=$NPM_TOKEN -q
 
+# This script builds all package bundles (for NPM) & demos (for gh-pages)
+./scripts/build-packages-and-demos.sh
+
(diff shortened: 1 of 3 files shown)


theiliad:
Hi @moores2. Could you please move this line in `.travis/after-success.sh`

```sh
# This script builds all package bundles (for NPM) & demos (for gh-pages)
./scripts/build-packages-and-demos.sh
```

below this:
```sh
npm config set //registry.npmjs.org/:_authToken=$NPM_TOKEN -q
```

That way travis will build the packages & demos for the PR and we'll make sure it works

PR #220, opened 2019-03-21
Refactor the build system and deploy process

cal-smith:
- everything builds into `dist` directories
- all `.d.ts` files should now have accompanying `.js` files
- simplified the folder structures
- angular should build correctly
- added `.umd.js` where appropriate
- built `style.scss` to `style.css` for core
- removed `style.scss` from `index.ts` for core
- other, small changes


packages/vue/package.json
```diff
@@ -4,13 +4,14 @@
   "description": "Carbon charting components for Vue",
   "main": "index.js",
   "files": [
-    "dist",
-    "index.js",
+    "charts-view.common.js",
+    "charts-view.umd.js",
+    "charts-view.umd.min.js",
```
theiliad: ...-vue?

cal-smith: ahahaha, yeah I actually meant `-vue`

cal-smith: You would not believe the number of times I've don that

theiliad: 😄

packages/angular/package.json
```diff
@@ -19,39 +19,31 @@
...
+		"@angular/core": "6.1.0",
+		"@angular/platform-browser": "6.1.0",
+		"@angular/platform-browser-dynamic": "6.1.0",
+		"@babel/core": "7.3.4",
+		"@storybook/angular": "5.0.3",
 		"@types/node": "8.5.2",
-		"ng-packagr": "4.4.0",
-		"typescript": "2.7.2"
+		"babel-loader": "8.0.5",
+		"d3": "4.11.0",
```
theiliad: d3 & babel-loader are need for the angular package?

packages/core/src/polyfills.ts
```diff
@@ -1,12 +1,12 @@
 (function() {
-	if (typeof window["CustomEvent"] === "function") return false;
+	if (typeof window["CustomEvent"] === "function") { return false; }
 
 	function CustomEvent(event, params) {
 		params = params || { bubbles: false, cancelable: false, detail: undefined };
-		var evt = document.createEvent("CustomEvent");
+		const evt = document.createEvent("CustomEvent");
```
theiliad: 👍

packages/react/rollup.config.js
```diff
@@ -1,17 +1,20 @@
 import babel from "rollup-plugin-babel";
-import commonjs from "rollup-plugin-commonjs";
 
 module.exports = {
     input: "./src/index.js",
     output: {
-        file: "./dist/bundle.js",
-        format: "cjs" 
+        file: "./dist/bundle/bundle.js",
```
theiliad: is there an extra space here? the indenting doesn't look right on github

scripts/build-packages-and-demos.sh
```diff
@@ -1,35 +1,36 @@
...
+# React build
+cd ../react
 npm run build
 npm run build-storybook
 cp -a storybook-dist/. ../../pages/react
 
-# Build Vue demos
+# Vue build
 cd ../vue
 npm run build
```
theiliad: If we're going to have a `build` script in every package, then we can run `lerna run build` and it'll build all, and so on for the other standard commands as well

cal-smith:
Fixed 👍

scripts/build-packages-and-demos.sh
cal-smith: That would definitely simplify things

theiliad: 👍

cal-smith: The demo build is `build-storybook` and `demo:build` ... if we standardize on one (`demo:build` I think) that'll work too

cal-smith: Each demo should be able to hoist itself to the top level pages directory too 🤔

theiliad: for sure, this whole file can be shrunk to the a few lines

cal-smith: Or that can just be a separate collect step in the deploy script ... that's probably cleaner

packages/react/rollup.config.js
cal-smith: Probably ... I'm guessing the formatting was hit with some weird copypasta

packages/core/src/polyfills.ts
cal-smith: Thank tslint!

packages/angular/package.json
cal-smith: babel-loader for storybook iirc ... let me check on d3 🤔

scripts/build-packages-and-demos.sh
theiliad: Pretty sure using `lerna run` you can have a script in root run in every package directory, that'll just regex all these folders and copy them over to `root/pages`

cal-smith: Yep! `lerna exec` + some mild demo dir changes and it works super well 🎉

packages/angular/package.json
cal-smith: It seems to be fine ... but it's better to explicitly declare our dependencies (d3 is a peer for core)

cal-smith:
fixed!

scripts/build-packages-and-demos.sh
```diff
@@ -1,39 +1,23 @@
...
-npm run build-storybook
-cp -a storybook-dist/. ../../pages/vue
 
-# Go back to project root folder
-cd ../..
+# run the build and demo:build script in all packages
+lerna run --parallel build
+lerna run --parallel demo:build
+# copy all the demos/{package name here} folders to the pages deploy directory
+lerna exec -- cp -a demo/bundle/. \$LERNA_ROOT_PATH/pages
```
theiliad: 🥇

cal-smith:
I'm pretty sure `netlify.sh` is up to date? I swapped it to run `npm run build-all`

theiliad:
Yeah that's what I wanted, the `npm` command

PR #227, opened 2019-03-26
fix(core): export patterns as strings of svgs

cal-smith:
- export patterns as strings in ts files (simplifies distribution, potentially simplifies adding patterns in the future)
- add d3@4 as a hard dep
- remove `after_success` from travis.yml ... it's not needed at this point

### DIFF
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -23,7 +23,8 @@
   "author": "IBM",
   "license": "Apache-2.0",
   "dependencies": {
-    "resize-observer-polyfill": "1.5.0"
+	"resize-observer-polyfill": "1.5.0",
+	"d3": "4.13.0"
   },
   "peerDependencies": {
     "d3": ">=4.11.0 <=5.7.0"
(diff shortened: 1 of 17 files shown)


packages/core/package.json
```diff
@@ -23,7 +23,8 @@
   "author": "IBM",
   "license": "Apache-2.0",
   "dependencies": {
-    "resize-observer-polyfill": "1.5.0"
+	"resize-observer-polyfill": "1.5.0",
```
theiliad: Has this also been taken out of the bundle?

cal-smith: not that I'm aware ... I haven't touched any of the package.jsons, so that's how it was before my changes

packages/vue/package.json
```diff
@@ -35,6 +35,9 @@
     "prettier": "1.16.4",
     "vue-template-compiler": "2.5.21"
   },
+  "publishConfig": {
```
theiliad: Right, for scoped packages, I wonder why this wasn't included in our initial vue PRs

cal-smith: I didn't notice it, and the vue cli didn't generate it 😕 in the future It'd be great to have some tooling take care of that kind of project syncing

PR #229, opened 2019-03-26
fix(core): remove suffixes to use regular module resolution, and force publish all packages with every release

cal-smith:
- oops
- also removed the unused patterns and re-numbered them


packages/core/demo/index.ts
```diff
@@ -10,7 +10,7 @@ import {
 // Styles
 import "./index.scss";
 import "./../src/style.scss";
-import { ChartData } from "./configuration";
+import { ChartData } from "./../src/configuration";
```
theiliad: I also would separate this from the `// Styles` group of imports

theiliad: We can do `// Interfaces`

PR #234, opened 2019-03-27
Update package READMEs and package.json keywords

theiliad:


packages/angular/README.md
```diff
@@ -1,6 +1,8 @@
-# `@carbon/charts-angular`
+# `@carbon/charts-vue`
```
cal-smith: Pretty sure this should stay `-angular` 😉 
```suggestion
# `@carbon/charts-angular`
```

PR #235, opened 2019-03-27
[WIP] - Save charts to SVG

theiliad:
### Updates
- Exports an SVG file that's editable in Sketch, Illustrator etc.
- Add `xlmns` attribute to the SVG element so it can be loaded as an image
- Add patterns that belong to the chart to final SVG
- Download file

### TODO
- Export styles
- Export font data

### DIFF
--- a/packages/core/demo/index.html
+++ b/packages/core/demo/index.html
@@ -123,6 +123,10 @@
 		<button class="bx--btn bx--btn--primary change-data-promise" type="button" data-promise-delay="2000">
 			Promise
 		</button>
+
+		<button class="bx--btn bx--btn--primary save-to-image" type="button">
+			Save to Image
+		</button>
 	</div>
 
 	<!-- Bar Chart -->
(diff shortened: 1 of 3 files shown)


theiliad:
Changes im making right now to this is to get a POC going for our demo,
might not necessarily be what I want to merge

cal-smith:
Okay, lets make sure any demo/POC things are noted with a comment and added to an issue(s) ...

theiliad:
Z & I usually don't review anything that says `WIP`. Creating a new label is kind of pointless since this is the last POC we'll build for the capstone

theiliad:
> Small suggestion - could we split out the blob generation and the download prompt?
> 
> Maybe `toString` and `toBlob`?
> 
> Also, is prompting for download something we want to handle in base-chart? It feels like something that would be better handled by a utility function outside of the chart instance - it doesn't need anything from the chart instance to function

Yeah so for this functionality I'm thinking we should only hand the developer a string including all the HTML, then they can process if they need to, or just download for the user. Optionally we can probably keep the download function

theiliad:
Closing this PR as we're rethinking some of the ideas around this functionality

PR #236, opened 2019-03-27
Fix Vue wrapper packaging issue

### LINKED ISSUE #231: WIP: Vue package does not have an index.js
We are currently publishing this tree to NPM for our `vue` package:
image

It is missing a fallback `index.js`. Therefore in cases where you do this:
image

You would encounter this error:
image

We should create an index.js that'll export out modules from a specific bundle, so you won't have to do this:
image

@moores2, @ZrianinaMariia

moores2:
### Fixes #231 
- Set `main` for package correctly
- Export \src and \dist to package, so that individual components can be imported by clients (like the React wrappers do).

Note: the vue-cli library build generates and extraneous dist/demo.html that does not work. There is an issue for this here: https://github.com/vuejs/vue-cli/issues/3291 

- [ ] Demos all features
- [ ] Documented/annotated
- [ ] Matches UI/UX specs
- [ ] Accessible

### DIFF
--- a/packages/vue/package.json
+++ b/packages/vue/package.json
@@ -2,7 +2,7 @@
   "name": "@carbon/charts-vue",
   "version": "0.11.10",
   "description": "Carbon charting components for Vue",
-  "main": "index.js",
+  "main": "dist/charts-vue.umd.min.js",
   "scripts": {
     "build": "bash build.sh",
     "lint": "vue-cli-service lint ./src/index.js",
@@ -38,6 +38,10 @@
   "publishConfig": {
     "access": "public"
   },
+  "files": [
+    "/dist",
+    "/src"
+  ],
   "lint-staged": {
     "*.{scss,css,js,md,vue}": [
       "yarn format",
--- a/packages/vue/package.json
+++ b/packages/vue/package.json
@@ -2,7 +2,7 @@
   "name": "@carbon/charts-vue",
   "version": "0.11.10",
   "description": "Carbon charting components for Vue",
-  "main": "dist/charts-vue.umd.min.js",
+  "main": "charts-vue.umd.min.js",
   "scripts": {
     "build": "bash build.sh",
     "lint": "vue-cli-service lint ./src/index.js",
@@ -38,10 +38,6 @@
   "publishConfig": {
     "access": "public"
   },
-  "files": [
-    "/dist",
-    "/src"
-  ],
   "lint-staged": {
     "*.{scss,css,js,md,vue}": [
       "yarn format",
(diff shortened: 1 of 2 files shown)


packages/vue/package.json
```diff
@@ -38,6 +38,10 @@
   "publishConfig": {
     "access": "public"
   },
+  "files": [
+    "/dist",
+    "/src"
```
cal-smith: So, with some changes we made, dist and src aren't automatically part of the build ... you can see this by checking `build.sh` and the ouput by running `npm run build` (we deploy the dist directory) ... if we want to include src (maybe not required?) we'll need to copy it into dist as part of `build.sh`

packages/vue/package.json
```diff
@@ -2,7 +2,7 @@
   "name": "@carbon/charts-vue",
   "version": "0.11.8",
   "description": "Carbon charting components for Vue",
-  "main": "index.js",
+  "main": "dist/charts-vue.umd.min.js",
```
cal-smith: This should just be 
```suggestion
  "main": "charts-vue.umd.min.js",
```

Since we deploy the contents of the dist directory now

theiliad:
@moores2 We need to modify `build.sh` for the `vue` package now to include `src`:
https://github.com/carbon-design-system/carbon-charts/blob/master/packages/vue/build.sh

cal-smith:
from my comment here: https://github.com/carbon-design-system/carbon-charts/issues/231#issuecomment-478570856

do we need so ship `src/`? I figure it best to just let consumers deal with the compiled JS ... especially with Vue, consumers really don't get anything by having it available.

theiliad:
Yeah it would be preferable if we have a bundle that exports modules separately, and one that includes all where dev can use it through CDNs

moores2:
I have reworked index.js with a simpler approach, and added exports so that the source is not needed in the shipped package. This update should allow clients to load the wrappers in 2 ways: 

Firstly, to just load all of the chart components globally into Vue by using a Vue plug-in, like this:
```
import Vue from 'vue';
import chartsVue from '@carbon/charts-vue';
...
Vue.use(chartsVue);
...
export default {
...
  template:
    '<ccv-bar-chart :data="barData" :options="barOptions"></ccv-bar-chart>'
```
The second option is to allow consumers to selectively import just the wrappers/components that they want, like this:
```
<template>
  ...
  <ccv-bar-chart :data="barData" :options="barOptions"></ccv-bar-chart>
  ...
</template>

<script>
import CcvBarChart from "@carbon/charts-vue";

export default {
  ...
  components: { CcvBarChart },
  ...
};
</script>
```

theiliad:
@moores2 @cal-smith This looks good 💯
Do we need any changes in the deploy scripts for the correct bundles/files/directories to be picked up?

cal-smith:
Nope we're good to go :+1:

cal-smith:
Merging

PR #237, opened 2019-03-27
Scatter demo fix

darsi-an:
### Updates
- Fixed axis labels for scatter-chart 
- Separate scatter data and options from line demo data
- Update initialize options for scatter

- [ ] Demos all features
- [ ] Documented/annotated
- [ ] Matches UI/UX specs
- [ ] Accessible


theiliad:
@darsi-an There seems to be a conflict for `packages/core/demo/index.ts`.

Could you please resolve this so we can merge?

darsi-an:
@theiliad Okay will do.

packages/core/src/scatter-chart.ts
```diff
@@ -7,10 +7,16 @@ import { ChartConfig, ScatterChartOptions, ChartType } from "./configuration";
...
+	options: ScatterChartOptions;
 
 	constructor(holder: Element, configs: ChartConfig<ScatterChartOptions>) {
 		super(holder, configs);
+		// initialize options
+		if (configs.options) {
+			this.options = Tools.merge({}, Configuration.options.SCATTER, configs.options);
+		} else {
+			this.options = Tools.merge({}, Configuration.options.SCATTER);
+		}
```
theiliad: This seems not necessary.

Doing:
```ts
options: ScatterChartOptions = Tools.merge({}, Configuration.options.SCATTER);
```

and then doing:
```ts
this.options = Tools.merge({}, Configuration.options.SCATTER, configs.options)
```

would be better

packages/core/src/scatter-chart.ts
```diff
@@ -7,10 +7,16 @@ import { ChartConfig, ScatterChartOptions, ChartType } from "./configuration";
 import { Tools } from "./tools";
 
 export class ScatterChart extends BaseAxisChart {
-	options: ScatterChartOptions = Tools.merge({}, Configuration.options.SCATTER);
+	options: ScatterChartOptions;
 
 	constructor(holder: Element, configs: ChartConfig<ScatterChartOptions>) {
 		super(holder, configs);
+		// initialize options
```
theiliad: Can we add an empty line here right after:
```ts
super(holder, configs);
```

packages/core/demo/index.ts
```diff
@@ -1,10 +1,10 @@
 import {
 	BarChart,
+	ScatterChart,
 	LineChart,
 	PieChart,
 	DonutChart,
-	ComboChart,
-	ScatterChart
+	ComboChart
```
theiliad: Why are we moving these imports around?

packages/core/src/scatter-chart.ts
darsi-an: The code follows how other charts initialize options, so i would have to change those as well. I'll open another PR for that.

packages/core/demo/index.ts
darsi-an: I just moved combo to the bottom as its sub-charts include the charts above and, scatter before line. It doesn't really matter, you want me to revert this?

theiliad: I see. We can keep it, it doesn't really matter

packages/core/src/scatter-chart.ts
theiliad: I see. Ok we can merge this and fix these in a different PR.

If we use the `getProperty` function in `tools.ts`:
```ts
this.options = Tools.merge({}, Configuration.options.SCATTER, getProperty(configs, "options"));
```

It would just return `null` for `configs.options` if it doesn't exist, and we'd get what we want.

packages/core/demo/chart-types.ts
```diff
@@ -15,7 +15,9 @@ const {
 	curvedLineData,
 	lineData,
 	lineOptions,
+	// Scatter
```
cal-smith: Unused?
```suggestion
```

packages/core/src/scatter-chart.ts
darsi-an: Okay, noted. Good catch btw.

packages/core/demo/chart-types.ts
darsi-an: Just a comment, ever other chart-type has a comment as well. Line and scatter are two different implementations.

PR #241, opened 2019-04-12
fix(react): use es6 style exports in index.js

cal-smith:


cal-smith:
Didn't mean to close this 😅

theiliad:
@cal-smith is netlify failing because it took so long?

cal-smith:
Yeah, netlify seems to have some weird time of day build limits - I've noticed midday builds are more likely to fail than morning or afternoon ones. We may have to look into some build optimizations anyway.

Issue #163, opened 2019-01-24
Bar chart Legend 'expand' should be able to be hidden / disabled


asahlin:
I am submitting a...

- [ ] Feature request
- [ ] Design defect
- [x] Source code defect
- [ ] Demo/documentation defect
- [ ] Other

charts version:
0.8.6

Issue description
I have 2 bar charts side by side on a page, I only have one key in my legend, but it appears based on chart width (600px or under) the legend '+' expand is automatically shown.

Steps to produce the issue
Bar chart that is 600px or under

Current behavior
Legend expand '+' is being shown

Expected behavior
Ability to turn that feature off, or have it based on number of keys in legend & size

Screenshot or recording
![image](https://user-images.githubusercontent.com/13006634/51696526-523c0c80-1fcb-11e9-9648-b9f700b256a5.png)

theiliad:
Good catch

PR #242, opened 2019-04-18
Legend issues

licarijd:
### Updates
Fix issue #163 https://github.com/carbon-design-system/carbon-charts/issues/163

Legend tooltip only becomes active when 1 or more legend items are hidden. When all legend items are visible, the tooltip disappears.

https://www.loom.com/share/7e11d1e497934534821b4b06d6e11a74

- [ ] Demos all features
- [ ] Documented/annotated
- [ ] Matches UI/UX specs
- [ ] Accessible


theiliad:
I'm a little bit confused.

In the deploy preview I can see that the first chart has the expand button disabled, however the legend items are still cut off and I won't be able to see the ones getting cropped out (as a user).

I think there are some design decisions to be made here.

packages/core/src/base-chart.ts
```diff
@@ -954,4 +1000,30 @@ export class BaseChart {
...
 	}
+// ================================================================================
+	// Makes chart data components more screen reader friendly
+	// ================================================================================
+	makeAccessible (element?: any, dataList?: any) {
+		// Make chart data components tabbable
+		element.attr("tabindex", 0);
+		// Aria label contains the x value, y value, and dataset for the data element
+		if (element.attr("class") === "bar" || element.attr("class") === "dot") {
+			element.attr("aria-label", (d) => `Label: ${d.label}, Value: ${d.value}, belongs to ${d.datasetLabel}`);
```
zvonimirfras: No braces if not necessary.

```suggestion
			element.attr("aria-label", d => `Label: ${d.label}, Value: ${d.value}, belongs to ${d.datasetLabel}`);
```

packages/core/src/base-chart.ts
```diff
@@ -954,4 +1000,30 @@ export class BaseChart {
...
+		// Slices
+		} else if (dataList !== undefined) {
+			element.attr("aria-label", (d) => `Label: ${d.data.label}, Value: ${d.value},
```
zvonimirfras:
```suggestion
			element.attr("aria-label", d => `Label: ${d.data.label}, Value: ${d.value},
```

packages/core/src/base-chart.ts
```diff
@@ -954,4 +1000,30 @@ export class BaseChart {
...
+			percentage is ${Tools.convertValueToPercentage(d.data.value, dataList)}`);
+		// Legend items
+		} else if (element.attr("class") === "legend-circle") {
+			element.attr("aria-label", (d) => `Legend Item: ${d.key}, Status: ${d.value ? "enabled" : "disabled"},
```
zvonimirfras:
```suggestion
			element.attr("aria-label", d => `Legend Item: ${d.key}, Status: ${d.value ? "enabled" : "disabled"},
```

packages/core/src/base-chart.ts
```diff
@@ -954,4 +1000,30 @@ export class BaseChart {
...
+			Click to ${d.value ? "disable" : "enable"}`);
+		// Chart containers
+		} else if (element.attr("chart-id")) {
+			element.attr("aria-label", `${element.attr("chart-id")}`);
+		// Legend containers
+		} else if (element.attr("class") === "legend-wrapper") {
+			element.attr("aria-label", "Legend - click items to add or remove them from the chart");
```
zvonimirfras: We need a way to make these hard-coded labels translatable into other languages.
In carbon-components-angular, we do that (to simplify) by providing inputs with defaults read from json files. Here, those could be config/options params.

Probably not a do-or-die for this PR. Definitely something we need to address. cc @theiliad @cal-smith

packages/core/src/base-chart.ts
```diff
@@ -554,9 +579,30 @@ export class BaseChart {
...
+		let totalLegendItems = 0;
+
+		// Count visible legend items
+		this.container.selectAll(".legend-btn.clickable").each(function(d) {
+
+			totalLegendItems++;
+
+			const node = select(this);
+
+			if (node.style("display") === "inline-block") {
```
zvonimirfras: Is there a better way for testing this?
If the intention is "not visually hidden" maybe "not none" would allow us to use other display values in future without breaking this piece.

theiliad:
Legend is getting re-implemented in SVG, therefore will close this PR

Issue #101, opened 2018-11-19
Curved line chart overflows onto axis on data change


t-mullen:
#### I am submitting a...

- [x] Source code defect

#### charts version:
core@0.7.4

### Issue description
Curved lines can overflow the chart and cover axis.

#### Steps to produce the issue
1. Create a line chart with `curve: 'curveNatural'`.
2. Provide a steep data set. [0, 100, 0]
3. Change data set to [0, 100, 0, 100]

#### Current behavior
Line overflows beneath x-axis and covers labels.

#### Expected behavior
Line disappears below x-axis, or chart resizes to fit curve.

### Screenshot or recording
<img width="607" alt="screen shot 2018-11-19 at 5 25 50 pm" src="https://user-images.githubusercontent.com/14932492/48738837-2e9dc400-ec20-11e8-8f06-7011c364045c.png">

t-mullen:
This is related to the cropping functionality needed for zoom/scroll.

PR #251, opened 2019-04-20
Line underflow 2

licarijd:
### Updates
add clip path to x grid
fixes underflow issues such as https://github.com/carbon-design-system/carbon-charts/issues/101

https://www.loom.com/share/42b8982fc5c9470a92a3e3daff877ef8

- [ ] Demos all features
- [ ] Documented/annotated
- [ ] Matches UI/UX specs
- [ ] Accessible


cal-smith:
I'm not sure this is entirely the right solution ... this looks broken:

image

But the line extending past the x axis is also non-ideal ... I'm just not sure which looks _less_ broken

theiliad:
Thanks @licarijd. This is something I suggest design to take a look at first.

@cameroncalder @shixiedesign

shixiedesign:
Can i get a bit more context/description on use case around this? Why is the path clipped and is it desirable?

theiliad:
> Can i get a bit more context/description on use case around this? Why is the path clipped and is it desirable?

So this is a somewhat of an edge case where you might have a few points on the very bottom of your y-axis, and lots of fluctuations. In such cases, a small part of the curve will go out of the chart, which can look a bit off.

theiliad:
@licarijd  I closed this PR for now since it's been inactive for a long time. Let me know if you'd still like to work on it and we can re-open.

PR #259, opened 2019-04-23
feat(core): first pass at updating colors

cal-smith:
First pass at swapping out the color palettes and some other small changes (pie/donut stroke, line datapoint size)


cal-smith:
Using #261 instead 👍

PR #267, opened 2019-05-07
Yarn workspaces & separate Netlify builds per package

theiliad:
**Updates:**
1. Update lerna to latest
2. Update storybook to latest
3. Switch to Yarn
4. Enable Yarn workspaces
5. Add yarn-offline-mirror
6. Add build script to build packages separately in Netlify

**Post-merge Checks:**
1. Whether in all packages, the dependency to `@carbon/charts` will update to the latest version


packages/core/package.json
```diff
@@ -37,7 +37,6 @@
     "babel-polyfill": "6.26.0",
     "copy-webpack-plugin": "4.5.2",
     "css-loader": "0.28.7",
-    "d3": "4.11.0",
```
cal-smith: Don't we still need a copy of `d3` in our devDependencies?

cal-smith: We should also probably look at removing the copy from `dependencies` since we have it listed in `peerDependencies` ... 

Or remove it from `peerDependencies`, but I don't think that's super viable

theiliad: I'm still not sure about what we should do with `d3`...

We can do peer and dev, but that'd make sense if `d3` doesn't end up in our bundle...

theiliad: @zvonimirfras

cal-smith: peer and dev is generally what we do for shared dependencies (angular, react, etc) ... I think we should bring this up in the slack channel - see what everyone thinks, and make sure they're aware we're (probably) changing to not install it by default

zvonimirfras: peer + dev makes sense
npm/yarn should warn that there's a not-installed peer dependency and additionally informing people of this change is useful and appreciated. 👍

theiliad: You guys are right. I wasn't thinking clearly about dev dependencies

cal-smith:
We can probably avoid the netlify timeout issue if we split up the demo builds across multiple instances ... for example `carbon-charts`, `carbon-charts-angular`, `carbon-charts-react`, `carbon-charts-vue`

theiliad:
> We can probably avoid the netlify timeout issue if we split up the demo builds across multiple instances ... for example `carbon-charts`, `carbon-charts-angular`, `carbon-charts-react`, `carbon-charts-vue`

Yeah I need to look at how that's done. Also adding offline-mirror should help

cal-smith:
Should just be as simple as pointing the build command to each package's `demo:build`

cal-smith:
_should_ being the key word 😛

cal-smith:
> Switch back to d3 v4.11.0 as 4.13 breaks charts

Do we know the cause? If it's truly incompatible we should make our upper limit `<4.13.0` in the peer dependency

theiliad:
> > Switch back to d3 v4.11.0 as 4.13 breaks charts
> 
> Do we know the cause? If it's truly incompatible we should make our upper limit `<4.13.0` in the peer dependency

So we normally do 4.11.0, in this PR only I upgraded to 4.13 and a few things broke. Seems to be the `d3-transition` package. We can update the range for now then look into higher versions later

cal-smith:
> > > Switch back to d3 v4.11.0 as 4.13 breaks charts
> > 
> > 
> > Do we know the cause? If it's truly incompatible we should make our upper limit `<4.13.0` in the peer dependency
> 
> So we normally do 4.11.0, in this PR only I upgraded to 4.13 and a few things broke. Seems to be the `d3-transition` package. We can update the range for now then look into higher versions later

Okay cool, once that's done I think this is good to go

theiliad:
> Looks good 👍
> 
> You'll also want to announce the change of D3 from dependency -> peerDependency and the upgrade to >5 only

For sure

PR #271, opened 2019-05-09
feat(core): Color palettes

theiliad:
In the previous PR some files were ignored accidentally. Adding those here.


cal-smith:
Awesome 👍 strange how the PR build didn't catch that though

Issue #254, opened 2019-04-23
Restyle: donut & pie 


shixiedesign:
Overall thinning the rings to `24px`, adding `1px` white divider line between sections, when label is short & few, display labels in chart instead of showing a legend. 

- [ ] Donut chart
- [ ] Pie chart

![(4) Donut   pie](https://user-images.githubusercontent.com/15144993/56553709-f2ff8100-6554-11e9-866f-2e537ff7bb5c.png)

theiliad:
Hi @shixiedesign, couple of pointers here:
1. What defines "when possible" for slice labels?
2. Everything else completed

shixiedesign:
> What defines "when possible" for slice labels?

For the sake of not keep checking and compromise performance, how about **when there are only 2 slices**? We know when there are 2 slices, there's no way labels would overlap / or be too close. When it gets up to 3 slices, there's a chance 1 slice is too small and there will be no room for label.

theiliad:
@shixiedesign Seems somewhat inconsistent doesn't it? Also not sure how we'll handle longer labels

shixiedesign:
I see your point. Let's tap into the larger community for some more takes.

shixiedesign:
Structure spec for pie & donut charts
https://github.com/carbon-design-system/carbon-charts/issues/257

shixiedesign:
## Additional spec for donut / pie chart states styles: 
- [ ] Hover state (tooltip spec in __issue # tbd__ )
- [ ] Small value
- [ ] Error state 1 & 2
- [ ] Empty state

FYI @theiliad. Nothing too controversial here I think, but I will go over these again during our Wednesday's call in case someone have issues with these. 

[(4) Donut & pie2.pdf](https://github.com/carbon-design-system/carbon-charts/files/3319102/4.Donut.pie2.pdf)

theiliad:
Thanks @shixiedesign.

I don't see the strokes on hover for slices, should we remove those?
![image](https://user-images.githubusercontent.com/14989804/60030941-d6291c00-9671-11e9-8e86-184efe034979.png)

shixiedesign:
Yes please! The width / thickness-growth alone would be the styling change for hover.

natashadecoste:
![image](https://user-images.githubusercontent.com/14351335/60046635-d804d680-9695-11e9-8bb4-d3e0186f7a4a.png)

[...]

PR #288, opened 2019-06-24
feat(core): Pie & Donut restyles

theiliad:
**Updates so far:**
- Remove strokes onHover for pie & donut
- Expand slice arcs onHover

Partially addresses #254


theiliad:
@shixiedesign For sure! Is that our easing for all transitions? Because I can just apply it for all.

Also let me know whether the durations you specified apply to all or just this interaction...

shixiedesign:
Yes for all the hover style changes. I think tooltip's should get it's own motion spec – would be similar to stay in sync with style changes, but an exit easing when they fade out. 

Any other state changes, like solo / legend hover, I will create specific motion spec too. Lemme know if this make sense!

packages/core/src/pie-chart.ts
```diff
@@ -366,27 +369,106 @@ export class PieChart extends BaseChart {
...
+	 */
+	private drawCallout(startPos, endPos, dir) {
+		// Clean up the label callouts
+		const callout = Tools.appendOrSelect(this.innerWrap, `g.callout-lines-${dir}`);
+		const midpointX = (endPos.x + startPos.x) / 2 ;
+
+		// draw vertical line
+		const verticalLine = Tools.appendOrSelect(callout, "line.vertical-line");
+		verticalLine
+		.style("stroke-width", "1px")
```
theiliad: Should be in `Configuration`

packages/core/src/pie-chart.ts
```diff
@@ -366,27 +369,106 @@ export class PieChart extends BaseChart {
...
+			.attr("x1", startPos.x)
+			.attr("y1", startPos.y)
+			.attr("x2", midpointX)
+			.attr("y2", endPos.y);
+
+		// draw horizontal line
+		const horizontalLine = Tools.appendOrSelect(callout, "line.horizontal-line");
+		horizontalLine
+		.style("stroke-width", "1px")
```
theiliad: Same as above

packages/core/src/pie-chart.ts
```diff
@@ -366,27 +369,106 @@ export class PieChart extends BaseChart {
...
+					x: xPosition,
+					y: yPosition + textOffsetY
+				};
+				// end position for the callout line
+				const endPos = {
+					x: xPosition - Configuration.pie.label.calloutOffsetX + textOffsetX + Configuration.pie.label.calloutTextMargin,
+					y: yPosition - Configuration.pie.label.calloutOffsetY
+				};
+				this.drawCallout(startPos, endPos, "left");
+				return `translate(${xPosition - Configuration.pie.label.calloutOffsetX}, ${yPosition - Configuration.pie.label.calloutOffsetY})`;
```
theiliad: Can we store the 2 in `xTranslate` and `yTranslate` variables?

packages/core/src/pie-chart.ts
natashadecoste: putting in styles.scss

packages/core/src/pie-chart.ts
natashadecoste: putting in styles.scss

packages/core/src/pie-chart.ts
natashadecoste: renamed it to be more concise, will still return the translate string (needed for adding to attribute of d3 selection).

theiliad:
can we add the piece here where you would move the pie chart down after having callouts so it shows right below the title. The refactor would address this, but we still should merge a fully functional version of this here.

Also, please check that once you do that, even in cases without callout the chart is aligned well.

packages/core/src/pie-chart.ts
```diff
@@ -232,6 +208,30 @@ export class PieChart extends BaseChart {
 					.style("opacity", 1);
 			});
 
+		// fade out left callout
+		this.innerWrap.select("g.callout-lines-left")
+			.transition()
+			.duration(Configuration.transitions.default.duration / 2)
```
theiliad: is this a new duration? Should be added to `Configurations` like so:
image

and this function would get the transition:
image

packages/core/src/pie-chart.ts
```diff
@@ -232,6 +208,30 @@ export class PieChart extends BaseChart {
...
+			.style("opacity", 0)
+			.on("end", function(d) {
+				select(this)
+					.transition()
+					.duration(Configuration.transitions.default.duration / 2)
+					.style("opacity", 1);
```
theiliad: Same as above

theiliad: Also, I don't think you need `.on("end")` here, you can just chain:
```js
.transition()
	.duration(Configuration.transitions.default.duration / 2)
	.style("opacity", 1);
```

to `.style("opacity", 0)`

packages/core/src/pie-chart.ts
```diff
@@ -232,6 +208,30 @@ export class PieChart extends BaseChart {
...
+		this.innerWrap.select("g.callout-lines-right")
+			.transition()
+			.duration(Configuration.transitions.default.duration / 2)
+			.style("opacity", 0)
+			.on("end", function(d) {
+				select(this)
+					.transition()
+					.duration(Configuration.transitions.default.duration / 2)
+					.style("opacity", 1);
+		});
```
theiliad: Same as above

packages/core/src/pie-chart.ts
```diff
@@ -289,27 +288,23 @@ export class PieChart extends BaseChart {
...
 		this.innerWrap.selectAll("path")
 			.on("click", d => self.dispatchEvent("pie-slice-onClick", d))
 			.on("mouseover", function(d) {
 				const sliceElement = select(this);
 				Tools.moveToFront(sliceElement);
 
-				sliceElement.attr("stroke-width", Configuration.pie.mouseover.strokeWidth)
-					.attr("stroke-opacity", Configuration.pie.mouseover.strokeOpacity)
-					.attr("stroke", self.getStrokeColor(self.displayData.datasets[0].label, d.data.label, d.data.value));
+				sliceElement.transition(self.getDefaultTransition("pie_slice_hover"))
```
theiliad: is the indentation correct here?

packages/core/src/pie-chart.ts
```diff
@@ -366,27 +369,123 @@ export class PieChart extends BaseChart {
...
+					y: yPosition - Configuration.pie.callout.calloutOffsetY + textOffsetY
+				};
+
+				// last slice always gets callout to the right side
+				this.drawCallout(startPos, endPos, Configuration.pie.callout.direction.RIGHT);
+				return `translate(${endPos.x + Configuration.pie.callout.calloutTextMargin + textOffsetX},
+					${yPosition - Configuration.pie.callout.calloutOffsetY})`;
+			}
+			// remove any unneeded callout for last slice
+			this.removeCallout(Configuration.pie.callout.direction.RIGHT);
```
theiliad: trailing whitespace?

packages/core/src/pie-chart.ts
```diff
@@ -366,27 +369,123 @@ export class PieChart extends BaseChart {
...
+					x: xPosition - Configuration.pie.callout.calloutOffsetX,
+					y: yPosition - Configuration.pie.callout.calloutOffsetY + textOffsetY
+				};
+
+				this.drawCallout(startPos, endPos, Configuration.pie.callout.direction.LEFT);
+				return `translate(${endPos.x - textOffsetX - Configuration.pie.callout.calloutTextMargin},
+					${yPosition - Configuration.pie.callout.calloutOffsetY})`;
+			}
+			// remove any leftover unneeded callout
+			this.removeCallout(Configuration.pie.callout.direction.LEFT);
```
theiliad: trailing whitespace?

packages/core/src/tools.ts
```diff
@@ -163,15 +163,18 @@ export namespace Tools {
...
 	 * @param {any} item
 	 * @param {any} fullData
-	 * @returns The percentage in the form of a string "87%"
+	 * @returns The percentage in the form of a number
 	 */
 	export function convertValueToPercentage(item, fullData) {
-		return Math.floor(item / fullData.reduce((accum, val) => accum + val.value, 0) * 100) + "%";
+		const percentage = item / fullData.reduce((accum, val) => accum + val.value, 0) * 100;
+		return percentage < 1 ?
+			(item / fullData.reduce((accum, val) => accum + val.value, 0) * 100).toPrecision(1) :
```
theiliad: what about:
```js
return percentage < 1 ? percentage.toPrecision(1) : percentage;
```

packages/core/src/configuration.ts
```diff
@@ -745,6 +745,9 @@ export const transitions = {
 	pie_slice_hover: {
 		duration: 100
 	},
+	pie_chart_titles: {
```
theiliad: aren't they just callouts?

packages/core/src/pie-chart.ts
```diff
@@ -295,15 +279,17 @@ export class PieChart extends BaseChart {
 				const sliceElement = select(this);
 				Tools.moveToFront(sliceElement);
 
-				sliceElement.transition(self.getDefaultTransition("pie_slice_hover"))
+				sliceElement
+					.transition(self.getDefaultTransition("pie_slice_hover"))
```
theiliad: trailing whitespace?

packages/core/src/pie-chart.ts
```diff
@@ -295,15 +279,17 @@ export class PieChart extends BaseChart {
...
 					.attr("d", self.hoverArc);
 
 				self.showTooltip(d);
 				self.reduceOpacity(this);
 			})
 			.on("mousemove", d => self.tooltip.positionTooltip())
 			.on("mouseout", function(d) {
-				select(this).transition(self.getDefaultTransition("pie_slice_hover"))
+				select(this)
+					.transition(self.getDefaultTransition("pie_slice_hover"))
```
theiliad: whitespace

packages/core/src/configuration.ts
natashadecoste: the same transition needs to be on the callouts as the chart titles. I'm grouping together callouts and titles (since callout cannot exist without a corresponding title). this gets applied to the title texts and the lines

theiliad: ok great

packages/core/src/configuration.ts
```diff
@@ -647,9 +647,21 @@ export const pie = {
...
+		direction: {
+			LEFT: "left",
+			RIGHT: "right"
+		}
+	},
 	default: {
 		strokeWidth: 2
-	}
+	},
+	chartPadding: 20
```
theiliad: the parent is already a chart, could `padding` be a more appropriate name?

cal-smith:
@natashadecoste @theiliad I'm seeing this on firefox, just want to confirm it's not an issue?

image

natashadecoste:
@theiliad I'm not seeing any font problems on Chrome or Firefox.
I've checked the netlify builds (all of them) in Chrome as well as Mozilla 69.0b16.
For Safari I can see the fonts are messed up, seems like Safari isnt properly loading **IBM Plex Sans Condensed** but has no problem with **IBM Plex Sans**. I added some more font-family alternatives to ensure we get a san-serif since it does not use inherited parent font-family styles. Safari seems to think it has loaded in Plex Sans Condensed properly and does not take parent styles.

cal-smith:
Seems like everything's working now :+1: not sure what was up earlier

Issue #244, opened 2019-04-19
Update layout & typography to follow new language


shixiedesign:
## Layout, spacing, alignment

Spacing and alignment spec below 

![image](https://user-images.githubusercontent.com/15144993/56442143-2e7c2000-62b5-11e9-8762-74717b7cc0a8.png)

## Typography, grid, & color

Please note the addition of `Plex sans condensed` font. `12/16 px` refers to 12 px type size, and 16px line height. 

### Light UI

```css
Chart title: Plex sans semibold, 14/18 px, Gray 100
Legend name: Plex sans condensed, 12/16 px, Gray 60
background-color: White #ffffff
Grid lines: 1px Gray 10
Axis: 1px Gray 50 
Axis names: Plex sans regular, 12/16 px, Gray 100
Axis labels (string): Plex sans condensed, 12/16, Gray 60
Axis labels (num): Plex mono, 12/16, Gray 60

```

### Dark UI

```css
Chart title: Plex sans semibold, 14/18 px, Gray 100
Legend name: Plex sans condensed, 12/16 px, Gray 60
background-color: Black #000000
Grid lines: 1px Gray 90
Axis: 1px Gray 60 
Axis names: Plex sans regular, 12/16 px, Gray 10
Axis labels (string): Plex sans condensed, 12/16, Gray 50
Axis labels (num): Plex mono, 12/16, Gray 50
```

![image](https://user-images.githubusercontent.com/15144993/56442220-874bb880-62b5-11e9-8174-48f48c5fe209.png)

## Pie & Donut charts

![image](https://user-images.githubusercontent.com/15144993/60040603-00350b00-967f-11e9-96d9-9499ea3db057.png)

natashadecoste:
@shixiedesign I just want to confirm that the Chart Title font is **semibold** for all charts. Under pie/donut chart specs it is listed as _Plex sans regular_ and in Light UI and Dark UI it is listed as _Plex sans semibold_

shixiedesign:
@natashadecoste Yes! All Chart titles should be **semibold** Sorry about the mistake! I will corrected it for future sketchfiles.

natashadecoste:
[...]

PR #292, opened 2019-06-27
Chart Titles

natashadecoste:
### Updates
- Addition of optional titles 
- Added titles to demo page for some charts (photo below)
- Not all style specs are present
- https://github.com/carbon-design-system/carbon-charts/issues/244

image

- [x] Demos all features
- [x] Documented/annotated
- [x] Matches UI/UX specs
- [ ] Accessible


packages/core/demo/demo-data/bar.ts
```diff
@@ -72,7 +72,8 @@ export const groupedBarOptions = {
 		size: "compact"
 	},
 	legendClickable: true,
-	containerResizable: true
+	containerResizable: true,
+	chartTitle: "Bar Chart"
```
theiliad: we can use a simpler key here like `title`, the root level of the options JSON corresponds to the chart anyways

packages/core/src/base-chart.ts
```diff
@@ -166,6 +166,7 @@ export class BaseChart {
 				// Perform the draw or update chart
 				if (initialDraw) {
 					this.initialDraw();
+					this.drawTitle(this.options.chartTitle);
```
theiliad: drawTitle itself has access to `this.options.chartTitle`, no reason to pass it in

packages/core/src/base-chart.ts
```diff
@@ -363,6 +367,32 @@ export class BaseChart {
...
+			.append("text")
+			.classed("chart-title", true)
+			.text(title);
+		}
+		const titleMargin = Configuration.charts.marginForChartTitle;
+
+
+		// retrieve the current transformations (specific to each chart type)
+		const transform = this.innerWrap.attr("transform");
+		const translateArr = transform.substring(transform.indexOf("(") + 1, transform.indexOf(")")).split(",");
```
theiliad: **Do we absolutely need this?**
**If no =>** what's a more readable way to do this?
**If yes =>** this should become a function in `tools.ts` that can parse transform strings, and give you certain values back

packages/core/src/base-chart.ts
```diff
@@ -363,6 +367,32 @@ export class BaseChart {
...
+
+
+		// add padding for title, keep other translations
+		this.innerWrap
+		.attr("transform", `translate(${translateArr[0]}, ${+translateArr[1] + titleMargin})`);
+
+
```
theiliad: empty lines here^^

packages/core/src/base-chart.ts
```diff
@@ -363,6 +367,32 @@ export class BaseChart {
 		resizeObserver.observe(this.holder);
 	}
 
+	/**
+	 * Draws title within the Chart's svg element.
+	 * @param title
+	 */
+	drawTitle(title: string) {
```
theiliad: Overall the way this function works isn't too ideal:
1. What if `title` is updated? Seems like we're only adding a title here and not updating
2. A function that deals with titles is updating the `transform` string for the entire graph

packages/core/src/configuration.ts
```diff
@@ -501,6 +505,7 @@ export const charts = {
 	minWidth: 150,
 	widthBreak: 600,
 	marginForLegendTop: 40,
+	marginForChartTitle: 24,
```
theiliad: We need more readability and extendability here:
e.g. `Configuration.charts.title.marginBottom`

We'll also have to change things like `marginForLegendTop` down the road, that was inherited from the old charts codebase...

packages/core/src/pie-chart.ts
```diff
@@ -127,7 +127,7 @@ export class PieChart extends BaseChart {
 			.attr("height", `${diameter}px`);
 
 		this.innerWrap
-			.style("transform", `translate(${radius}px,${radius}px)`)
+			.attr("transform", `translate(${radius},${radius})`)
```
theiliad: What was the issue here?

packages/core/src/style.scss
```diff
@@ -17,10 +17,22 @@ div.chart-holder {
...
 
+
 	// Chart elements
 	svg.chart-svg {
 		overflow: visible !important;
 
+		/*
+		Chart titles
+		*/
+		.chart-title {
```
theiliad: Could be more specific, `text.chart-title`

packages/core/src/base-chart.ts
natashadecoste: Yes, so true, this should be in tools.  We need this because specific chart types are translated already (donut/pie) since they are drawn from the center. Adding an optional title we need to translate the chart more but not undo the current translations that center the chart.

packages/core/src/pie-chart.ts
natashadecoste: its just different than the other charts extending the basechart and changing it allows it to be more consistent with the other charts for spacing purposes. Returning an array with the translations will work effectively for all chart types now.

packages/core/demo/demo-data/pie-donut.ts
```diff
@@ -4,7 +4,9 @@ export const pieOptions = {
 	accessibility: false,
 	legendClickable: true,
 	containerResizable: true,
-	colors
+	colors,
+	title: "Pie Chart",
+	theme: "dark"
```
theiliad: How many themes are we required to support?
If there are multiple, themes should be exported in the library.

e.g.
```js
import { themes } from "@carbon/charts";

const chartOptions = {
  theme: themes.G90,
  theme: themes.G30,
  theme: themes.G20,
  theme: themes.P90,
  theme: themes.DARK
}
```

Similar to that^^

packages/core/src/pie-chart.ts
theiliad: ok

packages/core/src/style.scss
```diff
@@ -17,9 +17,20 @@ div.chart-holder {
 	position: relative;
 	@include fonts();
 
+
```
theiliad: 2 empty lines here. should be 1

packages/core/src/tools.ts
```diff
@@ -233,6 +233,16 @@ export namespace Tools {
 		return transformMatrixArray[4];
 	}
 
+	/**
+	 * Returns an array of the translations from the element's attributes.
+	 * @param elementRef
+	 */
+	export function getTranslations(elementRef: any) {
```
theiliad: I don't think this function will work as you expect it to.

For instance, if elementRef was:
```html
<text transform="translate(50, 30) rotate(31, 80)"></text>
```

Then it'd grab the wrong thing.

Also, the function should be a little more general and less specific to `translate`. It'd be good if it separated all the transforms into an object or array, then depending on the use case we'd take the specific transform out of that.

packages/core/src/base-chart.ts
natashadecoste: 1. I will move it outisde the initializing of the chart so it runs anytime the setData is called 

2. is there a better way to move the entire chart rather than translations on the inner-wrap? 

I do think it could be nicer to remove the layout spacing code to its own function and have this **only** draw the title.

packages/core/demo/demo-data/pie-donut.ts
natashadecoste: im going tol remove the theme in favour of it being in its own PR

packages/core/src/tools.ts
natashadecoste: yeah its useful in the long run to get all the transforms

packages/core/demo/demo-data/pie-donut.ts
theiliad: 👍

packages/core/src/tools.ts
```diff
@@ -81,6 +81,17 @@ export namespace Tools {
 			height: parseFloat(el.style.height.replace("px", "") || el.offsetHeight)
 		};
 	}
+	/**
+	 * Returns an elements's x and y translations from its computed style
+	 * @param {HTMLElement} element
+	 * @returns an object containing the x and y translations
+	 */
+	export function getTranslationValues(elementRef: HTMLElement) {
```
theiliad: This still does something too specific. Take a look at this link:
https://stackoverflow.com/a/17838403

It'd be better if we can just parse the CSS transform values (or the inline transforms), and also return them in a more readable way:
```js
{
    translate: { x: 5, y: 10 },
    scale: { x: 1, y: 1.2 }
}
```

theiliad: Then you'd just take those results and do something like:
```js
result.translate.y
```

packages/core/demo/demo-data/pie-donut.ts
theiliad: Let's remove this line then in favour of the PR you mentioned

packages/core/src/tools.ts
```diff
@@ -81,6 +81,17 @@ export namespace Tools {
...
+		const transformMatrix = window.getComputedStyle(elementRef).getPropertyValue("transform").replace(/\s/g, "");
+		// returns matrix(a, b, c, d, tx, ty) of transformation values (2d transform)
+		const transformValues = transformMatrix.substring(transformMatrix.indexOf("(") + 1, transformMatrix.indexOf(")")).split(",");
+		return {tx: transformValues[4], ty: transformValues[5]};
```
theiliad: Hard to read object here, better to do something like:
```js
return {
   tx: transformValues[4],
   ty: transformValues[5]
};
```

theiliad:
Also looking at the deploy previews, there are a few other issues:
image

1. Pressing `Update Data` breaks the charts
2. Titles keep getting added on re-renders or data updates. In the `v2` refactor I'm doing I use this function that you can use here as well https://github.com/theiliad/carbon-charts/blob/89eb27aca315b7de02061e6c092272a98db5e480/packages/core/src/dom-utils.ts#L113-L124

It will create an element if it doesn't exist, and if it does it'll give you the reference, then you can update regardless...

packages/core/src/tools.ts
natashadecoste: yeah sure, I **replaced** a diff function in tools.ts that only returned the x translate coordinate in favour of **returning both** but I could add scale in case we need that later. i'll change the original parsing also

natashadecoste: after looking at this, it seems that returning a CSSStyleDeclaration object using HTMLElement.style **does not** return accurate (or any) transform that has been applied manually with d3. Only using window.getComputedStyle will take into account the applied transformations after the dom has been setup. They return the same CSSStyleDeclaration but when getting the transform information it is returned as a _matrix(a,b,c,d, tx, ty)_ rather than the nicer _"translate(5,6)rotate(5)"_ format to parse

theiliad:
Another thing I realized just now, these texts aren't perfectly aligned to the squares
image

natashadecoste:
> Another thing I realized just now, these texts aren't perfectly aligned to the squares
> image

I left legends out of this PR, i can fix and recommit but legends need their own layouts/spacing still to be done so would make more sense not in this PR

PR #311, opened 2019-08-06
Backdrop fix

natashadecoste:
### Updates
-  added a resize observer for the gridline changes (due to updated data OR due to resize of chart/charting container or entire window).

- [ ] Demos all features
- [ ] Documented/annotated
- [ ] Matches UI/UX specs
- [ ] Accessible

### DIFF
--- a/packages/core/src/base-axis-chart.ts
+++ b/packages/core/src/base-axis-chart.ts
@@ -1,7 +1,7 @@
 // D3 Imports
 import { select } from "d3-selection";
 import { scaleBand, scaleLinear, ScaleBand, ScaleLinear } from "d3-scale";
-import { axisBottom, axisLeft, axisRight, AxisScale, AxisDomain } from "d3-axis";
+import { axisBottom, axisLeft, axisRight } from "d3-axis";
 import { min, max } from "d3-array";
 
 import { BaseChart } from "./base-chart";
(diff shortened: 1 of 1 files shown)


packages/core/src/base-axis-chart.ts
```diff
@@ -177,6 +181,22 @@ export class BaseAxisChart extends BaseChart {
...
+		const gridHeight = this.holder.querySelector(".x.grid");
+
+		const resizeObserver = new ResizeObserver((entries) => {
+			this.drawBackdrop();
+		});
+
+		// watch resizing of both grids
+		resizeObserver.observe(gridWidth);
+		resizeObserver.observe(gridHeight);
+	}
```
theiliad: we already catch resize events for the entire chart. Seems like in `resizeChart()` we can just call `this.drawBackdrop();`?

natashadecoste: to my understanding (correct me if im wrong), the reason backdrop isnt resizing fluidly using drawBackdrop() within the resizeChart() is because theres is a timeout in updateXandYGrid(). this is since there is some delay with the axis repositioning (of ticks and stuff). having the resize observer on the actual grid will force the backdrop to resize if the grid is to ever resize, rather than if the entire chart receives a resize event (and is not even done resizing). 

the main issue is that the current resizeObserver in base-chart calls resizeChart and all the calculations are done so quick that the x and y grid are not yet in their final locations causing backdrop to render improperly. if the backdrop is tied to the actual x and y grid on resize, then there is a valid position for the backdrop to when it is called after grid has been updated rather than in preparation of an update.

theiliad: is there any way to do this without adding a second resizeObserver?

PR #317, opened 2019-08-15
change backdrop to use axis instead of grid

natashadecoste:
### Updates
- move the timeout to the resize, lets the backdrop update code get put into the queue for the timer handler and should get run after any setTimeout() code blocks that are within resize (of specific chart or of parent charts) 
- moved drawBackdrop into updateXY since it needs to update after the grid does, otherwise updateXY was running a setTImeout (queuing it to the end) and drawBackdrop was running before it
- added a getGridContainer for base chart that uses the main x and y axis (sometimes theres 2 y axis, it uses the main one, not y2) to get a combined bounding box and get the grid container.


packages/core/src/base-axis-chart.ts
```diff
@@ -483,19 +483,35 @@ export class BaseAxisChart extends BaseChart {
...
+
+	/**
+	 * Draws the background for the chart grid. Uses the axis to get the bounds and position of the backdrop.
 	 */
 	drawBackdrop() {
-		// Get height from the grid
-		const xGridHeight = this.innerWrap.select(".x.grid").node().getBBox().height;
-		const yGridBBox = this.innerWrap.select(".y.grid").node().getBBox();
 		const backdrop = Tools.appendOrSelect(this.innerWrap, "rect.chart-grid-backdrop");
+			const gridContainer = this.getGridContainer();
```
theiliad: is there extra indentation here?

packages/core/src/base-axis-chart.ts
```diff
@@ -483,19 +483,35 @@ export class BaseAxisChart extends BaseChart {
...
 
-		backdrop
-			.attr("x", yGridBBox.x)
-			.attr("y", yGridBBox.y)
-			.attr("width", yGridBBox.width)
-			.attr("height", xGridHeight)
+			backdrop
```
theiliad: is there extra indentation here?

packages/core/src/base-axis-chart.ts
```diff
@@ -483,20 +484,19 @@ export class BaseAxisChart extends BaseChart {
...
-			.attr("x", yGridBBox.x)
-			.attr("y", yGridBBox.y)
-			.attr("width", yGridBBox.width)
-			.attr("height", xGridHeight)
-			.lower();
+		.attr("x", xScaleStart)
+		.attr("y", yScaleStart)
+		.attr("width", xScaleEnd)
+		.attr("height", yScaleEnd)
+		.lower();
```
theiliad: These should be indented

PR #333, opened 2019-09-04
Custom width values for bars

### LINKED ISSUE #319: Stack Bar Chart Angular - Unable to provide custom width and height to bars
Hi, 
I am implementing a stack bar chart in Angular.
image

In which there is only one stack bar.
So, when i am trying to provide custom width or height to chart its not taking that and using default minimum defined in chart component itself.

Expectation:
- Way to provide width and height for single stack bar.
- Way to hide background grid, x-axis and y-axis.
- And When I provide custom width using css it cause tooltip width issue when hover on stack bar corners.

image

Please test them specifically in mozilla firefox as well. As it shows quite different behaviour over there on hiding grid and providing custom width using css.

theiliad:
### Updates
- Support for `width` values
- Support for `maxWidth` values

image

Closes #319


packages/core/src/bar-chart.ts
```diff
@@ -31,7 +31,22 @@ const getYMin = configs => {
 // returns the configured max width or the calculated bandwidth
 // whichever is lower
 // defaults to the calculated bandwidth if no maxWidth is defined
-const getMaxBarWidth = (maxWidth, currentBandWidth) => {
+export const getBarWidth = function(chart) {
```
zvonimirfras: This can be an arrow function again :)

packages/core/src/bar-chart.ts
```diff
@@ -43,7 +58,10 @@ const getMaxBarWidth = (maxWidth, currentBandWidth) => {
 
 // returns true if the calculated bandwidth is greater than the maxWidth (if defined)
 // i.e. if we should be constraining ourselves to a specific bar width
-const isWidthConstrained = (maxWidth, currentBandWidth) => {
+export const isWidthConstrained = function(chart) {
```
zvonimirfras: This can be an arrow function again too :)

PR #352, opened 2019-09-30
update color palettes in stories & add scatter to react

theiliad:


packages/react/src/scatter-chart.js
```diff
@@ -0,0 +1,25 @@
...
+		);
+	}
+
+	render() {
+		return (
+			<div
+				ref={chartRef => this.chartRef = chartRef}
+				className="chart-holder">
+			</div>
+		)
```
zvonimirfras:
```suggestion
		);
```

packages/react/stories/line-step.stories.js
```diff
@@ -18,7 +19,7 @@ lineStories.add("Basic", () => (
 ));
 
 lineStories.add("Natural Curve", () => (
-	<LineChart
+	<ScatterChart
```
zvonimirfras:
```suggestion
	<LineChart
```

PR #356, opened 2019-10-02
Update color palettes

theiliad:


packages/core/src/services/colorPalettes.ts
```diff
@@ -1,36 +1,44 @@
...
+	"#d2a106", // Yellow 40
+	colors.teal(40),
+	colors.cyan(20),
+	"#ba4e00", // Orange 60
+	colors.purple(30)
 ];
 
-export const G90 = G10;
-export const G100 = G10;
+export const G10 = DARK;
```
shixiedesign: G10 should be getting light (or white) theme colors.

Issue #133, opened 2019-01-14
Add support for horizontal bar chart


asahlin:
I am submitting a...

- [x ] Feature request
- [ ] Design defect
- [ ] Source code defect
- [ ] Demo/documentation defect
- [ ] Other

charts version:
0.8.1

Issue description
We have a design to show bar charts horizontally. 

![image](https://user-images.githubusercontent.com/13006634/51137367-3cc52680-1804-11e9-8e95-35fb662c610f.png)

theiliad:
This is not in plan currently. What would help putting this in our roadmap would be if you have an immediate need for this in your project(s). Thanks!

theiliad:
Although a great idea, we don't have any plans for this type of chart right now.

What would help putting this in our roadmap would be if you have an immediate need for this in your project(s). Thanks!

asahlin:
We do have a need for this in our current project.  This design is for the 1st quarter deliverable for Gen1 / PUP device details page.

theiliad:
Thanks @asahlin, any deadlines you can provide?

asahlin:
March 31st is when it is suppose to GA.  So needs to before then ;-)

theiliad:
Sounds good, I'll get back to you

asahlin:
@theiliad Any update on this?

theiliad:
Hi @asahlin. I'm currently in the process of planning this. It is likely that we can get this done in time for you to incorporate & debug, I will get back soon with a confirmation. Are there any changes on your end in terms of deadline, requirements etc?

asahlin:
Ok great, no changes in our requirements/deadline.   Would be nice to get https://github.com/carbon-design-system/carbon-charts/issues/136 as well.   Our designers don't like using background colors they want the bar color itself to indicate threshold level.
  
There was discussion of investigating and using a 3rd party graphing package.    Which we don't want to do.

theiliad:
I see, well thank you for going forward with Carbon Charts 🙂.

We will do our best to satisfy our users.

theiliad:
Hi @asahlin, we've started to work on this. You'll see a PR soon

asahlin:
@theiliad That's awesome, thanks for the heads up.

asahlin:
@theiliad Is this still on the radar?   Any update?

lisalukenbill:
Following this as well. I have a need for horizontal bar charts and would like updates as to when the plan is to make these available.  Thank you!

[...]

PR #415, opened 2019-12-02
feat(core): add support for horizontal bar charts

natashadecoste:
### Updates
- change horizontal-zero-line to be vertical or horizontal (depending on bar chart orientation)
- added options for bar chart to be horizontal or vertical orientation
- updated bar-simple, bar-grouped and bar-stacked to support horizontal bars

### Demo screenshots or recording

**Horizontal Simple Bar**
image

**Horizontal Stacked Bar**
image

**Update the multipoint ordering on tooltip for horizontal stacked bar**
image

###  To-do
- [x] make sure domains get padded more so that bars dont overlap the axis lines ever
- [x] group bar chart horizontal
- [x] stacked bar chart horizontal
- [x] reverse order of the tooltip for the stacked bar (values will be top-down in order that bars are left to right)
- [x] add demos for all horizontal bar charts

- [ ] Demos all features
- [ ] Documented/annotated
- [ ] Matches UI/UX specs
- [ ] Accessible


natashadecoste:
Closes https://github.com/carbon-design-system/carbon-charts/issues/133

packages/core/demo/demo-data/bar.ts
```diff
@@ -69,6 +69,23 @@ export const groupedBarOptions = {
...
+	axes: {
+		left: {
+			scaleType: "labels",
+			primary: true,
+		},
+		bottom: {
+			secondary: true
+		}
+	},
+	orientation: "horizontal"
```
cal-smith: What happens if I flip the scales but don't add the `orientation` property?

Maybe this is covered elsewhere, but what's the function of the `primary`/`secondary` properties?

theiliad: The `primary` & `secondary` fields are a bit confusing right now...

We basically need 1 of each, and the flipping of the 2 would do nothing...

I WOULD like to rethink that portion

cal-smith: Okay ... I won't block the PR on that, but maybe something to make an issue for?

theiliad: So currently, you'd have to set `orientation` to `horizontal` then make bottom scale linear and left axis `labels` based.

Shouldn't that scale change be detected to switch the bars around?

packages/core/src/components/graphs/bar-simple.ts
```diff
@@ -47,26 +47,62 @@ export class SimpleBar extends Bar {
...
+						return Math.abs(this.services.axes.getXValue(d.label, i) - xScale.scale.range()[1]);
+					}
+					return Math.abs(this.services.axes.getXValue(d, i) - this.services.axes.getXValue(0));
+				})
+				.transition(this.services.transitions.getTransition("bar-update-enter", animate))
+				.attr("y", (d, i) => this.services.axes.getYValue(d, i) - (this.getBarWidth() / 2))
+				.attr("fill", d => this.model.getFillScale()(d.label))
+				.attr("height", this.getBarWidth.bind(this))
+				.attr("opacity", 1);
+		}
```
cal-smith: It'd be nice if there was a way to reduce the level of code duplication ... Maybe pulling out some of the functions would help?

packages/core/src/components/graphs/bar-stacked.ts
```diff
@@ -123,30 +127,76 @@ export class StackedBar extends Bar {
 
 		// Update styling and position on existing bars
 		// As well as bars that were just added
-		bars.enter()
-			.append("rect")
-			.merge(bars)
+		// horizontal bars need to map across the opposite axis
+		if (this.model.getOptions().orientation === "horizontal") {
```
cal-smith:
```suggestion
		if (this.model.getOptions().orientation === BarOrientationOptions.HORIZONTAL) {
```

packages/core/src/components/graphs/bar-stacked.ts
```diff
@@ -123,30 +127,76 @@ export class StackedBar extends Bar {
...
+							}
+							// create dividers between bars
+							if (datasetLabelIndex > 0 && height >= options.bars.dividerSize) {
+								return height - options.bars.dividerSize;
+							}
+
+							return height;
+						})
+						.attr("opacity", 1);
+			}
```
cal-smith: Same note about code duplication here ... not sure _how_ to fix it, but it'd be nice if we could

packages/core/demo/demo-data/bar.ts
cal-smith: That's what I would expect yeah ... Internally we can track it with an `orientation` enum, but externally it probably makes more sense if it just does the "right thing"

natashadecoste: i wanted the chart to work horizontally regardless of how the user maps the axes. 
for example this **makes sense** :

```

export const simpleHorizontalBarOptions = {
	title: "Simple horizontal bar (discrete)",
	axes: {
		left: {
			primary: true,
			scaleType: "labels"
		},
		bottom: {
			secondary: true,
		}
	},
	orientation: "horizontal"
};
```
 but this (below) does not really make sense, but can provide a chart nonetheless. 

```
export const simpleHorizontalBarOptions = {
	title: "Simple horizontal bar (discrete)",
	axes: {
		left: {
			primary: true,
		},
		bottom: {
			secondary: true,
			scaleType: "labels"
		}
	},
	orientation: "horizontal"
};
```

you would get this chart
image


Although it seems smart to let carbon-charts "decide" whether to go horizontal/vertical it seems to me that it would negatively affect some of the customizability that we introduced in the refactor.
By boasting that our refactor makes custom charts easy to create (even if they are not to spec and nor reccommended) I think we shouldn't decide for the users (or we provide a configuration to force a desired orientation).  I guess im proposing that if we want charts to "decide" that we keep a configuration for orientation which can be provided to force an orientation.

packages/core/src/components/graphs/bar-stacked.ts
natashadecoste: good catch 👍

packages/core/demo/demo-data/bar.ts
cal-smith: I don't think we would ruin the customization by simplifying _how_ we declare the chart orientation ... In your example the chart isn't what I would expect as a user, given that I could very easily miss the `orientaton` property - if we let the data drive the layout of the chart (in this case, the axes config data) we can actually be _more_ customizable, and clearer in how the customization works

packages/core/src/components/graphs/bar.ts
```diff
@@ -1,22 +1,30 @@
...
 export class Bar extends Component {
 	// Gets the correct width for bars based on options & configurations
 	protected getBarWidth() {
-		const mainXAxis = this.services.axes.getMainXAxis();
 		const options = this.model.getOptions();
+		let mainAxis;
 
-		if (!mainXAxis.scale.step) {
+		// if its horizontal we want to use the Y axis to determine instead of X
+		if (options.orientation === BarOrientationOptions.HORIZONTAL) {
```
cal-smith: This seems like the last place `optoins.orientation` is used ... From what I can see we should be able to replace it with a call to `this.services.cartesianScales.getOrientation`

packages/core/demo/demo-data/bar.ts
```diff
@@ -92,6 +108,22 @@ export const simpleBarOptions = {
...
+	axes: {
+		left: {
+			primary: true,
+			scaleType: "labels"
+		},
+		bottom: {
+			secondary: true,
+		}
+	},
+	orientation: "horizontal"
```
cal-smith: From what I can see we should be able to remove this after updating `graphs/bar`

packages/core/src/services/scales-cartesian.ts
```diff
@@ -0,0 +1,335 @@
...
+				const endDate = new Date(domain[1]);
+				if (differenceInYears(endDate, startDate) > 1) {
+					return [subYears(startDate, spaceToAddToEdges), addYears(endDate, spaceToAddToEdges)];
+				}
+				if (differenceInMonths(endDate, startDate) > 1) {
+					return [subMonths(startDate, spaceToAddToEdges), addMonths(endDate, spaceToAddToEdges)];
+				}
+				if (differenceInDays(endDate, startDate) > 1) {
+					return [subDays(startDate, spaceToAddToEdges), addDays(endDate, spaceToAddToEdges)];
+				} else if (differenceInHours(endDate, startDate) > 1) {
```
zvonimirfras: `else if` doesn't seem necessary and in combination with the `if`s above it that don't do else looks a bit peculiar.

packages/core/demo/demo-data/bar.ts
```diff
@@ -92,6 +108,21 @@ export const simpleBarOptions = {
...
+
+export const simpleHorizontalBarOptions = {
+	title: "Simple horizontal bar (discrete)",
+	axes: {
+		left: {
+			primary: true,
+			scaleType: "labels"
+		},
+		bottom: {
+			secondary: true,
```
zvonimirfras:
```suggestion
			secondary: true
```

packages/core/demo/demo-data/bar.ts
```diff
@@ -140,6 +171,22 @@ export const simpleBarTimeSeriesOptions = {
...
+// Horizontal simple time series
+export const simpleHorizontalBarTimeSeriesOptions = {
+	title: "Simple horizontal bar (time series)",
+	axes: {
+		left: {
+			scaleType: "time",
+			primary: true
+		},
+		bottom: {
+			secondary: true,
```
zvonimirfras:
```suggestion
			secondary: true
```

packages/core/demo/demo-data/bar.ts
```diff
@@ -201,6 +248,22 @@ export const stackedBarOptions = {
...
+export const stackedHorizontalBarOptions = {
+	title: "Stacked horizontal bar (discrete)",
+	axes: {
+		left: {
+			scaleType: "labels",
+			primary: true
+		},
+		bottom: {
+			stacked: true,
+			secondary: true,
```
zvonimirfras:
```suggestion
			secondary: true
```

PR #418, opened 2019-12-04
feat(core): add bubble chart + update demos

theiliad:


theiliad:
> It might be useful to have a min/max radius config and automatically scale radius values between them ... it can be tricky to hit the target when some of the bubbles are really small 🤔

Yeah that makes sense. @cameroncalder Would you advise as to whether we'd have min/max readiuses for the circles being rendered? If we do, then we can use that as a reference.

packages/core/src/components/graphs/bubble.ts
```diff
@@ -0,0 +1,65 @@
...
+import { extent } from "d3-array";
+import { scaleLinear } from "d3-scale";
+
+export class Bubble extends Scatter {
+	type = "bubble";
+
+	getRadiusScale(selection: Selection<any, any, any, any>) {
+		const data = selection.data();
+		const allRadii = data.map(d => d.radius)
+			.filter(d => d);
```
cal-smith: Filtering out `null`/`undefined` radii? Could probably be one line with the map

packages/core/src/configuration.ts
```diff
@@ -182,6 +183,21 @@ const scatterChart: ScatterChartOptions = Tools.merge({}, axisChart, {
 	}
 } as ScatterChartOptions);
 
+/**
+ * options specific to bubble charts
+ */
+const bubbleChart: BubbleChartOptions = Tools.merge({}, scatterChart, {
+	bubble: {
+		radiusRange: (chartSize, data) => {
```
cal-smith: This is to constrain the scale of the radii? Definitly going to need some docs around this

packages/angular/src/bubble-chart.component.ts
```diff
@@ -0,0 +1,34 @@
...
+	Component,
+	AfterViewInit
+} from "@angular/core";
+
+import { BaseChart } from "./base-chart.component";
+
+import { BubbleChart } from "@carbon/charts";
+
+/**
+ * Wrapper around `DonutChart` in carbon charts library
```
cal-smith:
```suggestion
 * Wrapper around `BubbleChart` in carbon charts library
```

packages/core/demo/index.ts
```diff
@@ -150,57 +142,25 @@ const createChartContainer = chartType => {
...
-			case "line-step-time-series":
-				classToInitialize = LineChart;
-				break;
-			case "pie":
-				classToInitialize = PieChart;
-				break;
-			case "donut":
-				classToInitialize = DonutChart;
-				break;
+demoGroups.forEach(demoGroup => {
```
cal-smith: Woohoo! way more readable than the crazy switch statement 😁

packages/core/src/components/axes/axis.ts
```diff
@@ -76,6 +77,8 @@ export class Axis extends Component {
 			const timeLocale = Tools.getProperty(options, "locale", "time");
```
zvonimirfras: Would
```suggestion
			const timeLocale = Tools.getProperty(options, "locale", "time") | englishLocale;
```
make it so we don't need the `if-else` block  and just call `timeFormatDefaultLocale(timeLocale)`?

packages/core/src/components/graphs/bubble.ts
```diff
@@ -0,0 +1,65 @@
...
+import { extent } from "d3-array";
+import { scaleLinear } from "d3-scale";
+
+export class Bubble extends Scatter {
+	type = "bubble";
+
+	getRadiusScale(selection: Selection<any, any, any, any>) {
+		const data = selection.data();
+		// Filter out any null/undefined values
+		const allRadii = data.map(d => d.radius).filter(d => d);
```
zvonimirfras: when conventional English is not good enough so it looks like a typo 😂

packages/core/src/components/graphs/bubble.ts
```diff
@@ -0,0 +1,65 @@
...
+			.attr("cx", (d, i) => this.services.cartesianScales.getDomainValue(d, i))
+			.transition(this.services.transitions.getTransition("bubble-update-enter", animate))
+			.attr("cy", (d, i) => this.services.cartesianScales.getRangeValue(d, i))
+			.attr("r", d => radiusScale(d.radius))
+			.attr("fill", d => {
+				if (filled) {
+					return this.model.getFillScale()[d.datasetLabel](d.label) as any;
+				}
+			})
+			.attr("fill-opacity", filled ? 0.2 : 1)
```
zvonimirfras: 0.2 looks like something we would have defined elsewhere.

packages/core/src/components/graphs/bubble.ts
zvonimirfras: Also, would `filter(radius => radius)` be more accurate?

packages/core/src/components/axes/axis.ts
cal-smith: I don't think that change is doing what you think it's doing 😉 `|` is a bitwise operator ... also if/else is way more readable than an inline `||` IMO

packages/core/demo/index.ts
theiliad: That was part of the _original_ codebase 🤣

packages/core/demo/demo-data/bubble.ts
```diff
@@ -0,0 +1,189 @@
+export const bubbleData = {
+	labels: ["Qty", "More", "Sold", "Restocking", "Misc"],
+	datasets: [
+		{
+			label: "Dataset 1",
+			data: [
+				{
+					date: new Date(2019, 0, 1),
```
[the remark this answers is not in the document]
theiliad: I'd create an issue for that. For sure btw

PR #422, opened 2019-12-05
feat(scatter): add the entire datapoint as a parameter to the callbacks to determine custom point colors

scottdickerson:
### Updates
The existing callbacks that allow us to set custom point colors on my graph for certain points(getStrokeColor, getFillColor, getIsFilled) unfortunately didn't solve my usecase.  I actually needed to examine the entire datapoint (not just the value) to determine which color to color each point. 

-in: additional parameter passing the entire datapoint to all 3 callbacks for custom colors and fills
-in: demo of the feature

image

- [x] Demos all features
- [ ] Documented/annotated
- [ ] Matches UI/UX specs
- [ ] Accessible


natashadecoste:
Is it also the intention that the tooltips change color for those points? 
This is what i'm seeing, the datapoint tooltip does not indicate the color of the dataset but rather the custom one
image

scottdickerson:
@natashadecoste this was intentional, but I think maybe a bad idea, do you think I should change it to the original line color?

scottdickerson:
@natashadecoste ok resubmitted with a fix to preserve the original line color in the tooltips

packages/core/demo/demo-data/line.ts
```diff
@@ -121,6 +121,23 @@ export const lineTimeSeriesOptions = {
...
+		return value > 60000 ? '#FF0000' : originalStrokeColor;
+	},
+	getFillColor: (datasetLabel, label, value, datum, originalFillColor)=>{
+		console.log(`originalFillColor: ${originalFillColor}`);
+		return value > 60000 ? '#FF0000' : originalFillColor;
+	},
+	getIsFilled: (datasetLabel, label, value, datum, originalIsFilled)=>{
+		console.log(`originalIsFilled: ${originalIsFilled}`);
+		return value > 60000 ? true : originalIsFilled;
+	}
```
theiliad: Seems like the 3 functions above need to go into fields that address the specific graphs.

e.g. what if you need different functions for line, then the scatter graphs, and the bars if you were using a combo chart?

Seems like this could be an addition to an interface somewhere and then read by each graph.

packages/core/src/components/graphs/scatter.ts
```diff
@@ -54,19 +54,19 @@ export class Scatter extends Component {
 		dotsEnter.merge(dots)
 			.raise()
 			.classed("dot", true)
-			.classed("filled", filled)
-			.classed("unfilled", !filled)
+			.classed("filled", d => this.model.getIsFilled(d.datasetLabel, d.label, d.value, d, filled))
+			.classed("unfilled", d => !this.model.getIsFilled(d.datasetLabel, d.label, d.value, d, filled))
```
theiliad: If we're calling a function here we're better off storing the value and not calling that function twice

packages/core/src/components/graphs/scatter.ts
```diff
@@ -54,19 +54,19 @@ export class Scatter extends Component {
...
 			.attr("cx", (d, i) => this.services.axes.getXValue(d, i))
 			.transition(this.services.transitions.getTransition("scatter-update-enter", animate))
 			.attr("cy", (d, i) => this.services.axes.getYValue(d, i))
 			.attr("r", options.points.radius)
 			.attr("fill", d => {
-				if (filled) {
-					return this.model.getFillScale()[d.datasetLabel](d.label) as any;
+				if (this.model.getIsFilled(d.datasetLabel, d.label, d.value, d, filled)) {
```
theiliad: again we don't need to re-call `this.model.getIsFilled` every time

scottdickerson:
@theiliad if you wanna get together internally to discuss, just lemme know

theiliad:
> @theiliad if you wanna get together internally to discuss, just lemme know

I think overall we just need the addition to every component rather than just scatter, to be able to go into options and look for similar options to the below:
```js
{
   scatter: {
      getStrokeColor: () => "red",
      getFillColor: () => "blue"
   },
   bars: {
      getStrokeColor: () => "red",
      getFillColor: () => "blue"
   },
   line: {
      getStrokeColor: () => "red",
      getFillColor: () => "blue"
   }
}
```

packages/core/src/components/graphs/scatter.ts
scottdickerson: this is true, but I think I need the specific datapoint from the d3 callback

packages/core/src/components/graphs/scatter.ts
scottdickerson: see above

scottdickerson:
@theiliad @natashadecoste this one's ready for a re-review please

packages/core/src/interfaces/charts.ts
```diff
@@ -34,14 +34,18 @@ export interface BaseChartOptions {
...
 	/**
 	 * Optional function to generate the fill color based on datasetLabel, label, and/or value
 	 */
-	getFillColor?: (datasetLabel: any, label?: any, value?: any) => string;
+	getFillColor?: (datasetLabel: any, label?: any, value?: any, data?: any, defaultFillColor?: string) => string;
 	/**
 	 * Optional function to generate the stroke color based on datasetLabel, label, and/or value
 	 */
-	getStrokeColor?: (datasetLabel: any, label?: any, value?: any) => string;
+	getStrokeColor?: (datasetLabel: any, label?: any, value?: any, data?: any, defaultStrokeColor?: string) => string;
```
theiliad: we discussed changing all of these to just receive a `d`, and maybe a `defaultValue`

scottdickerson: good catch, I had to leave the datasetLabel, and label separate from the data parameter because for tooltips they use the same interface (without having the full datapoint), but I have combined value and data into the same parameter

scottdickerson:
@theiliad @cal-smith sorry picked up some lint errors somewhere, just resubmitted with the lint errors fixed,  I think it threw away your reviews and we need to get the approval again

theiliad:
Sorry for dragging this out, but we might not actually want an example with red fills for the dots on our demo site:
image

We can either remove that or get design to review it

Issue #379, opened 2019-10-22
Allow removal of legends/labels in charts and removal of '%' labels at the outer of donut charts


jagdishkumar1:
#### I am submitting a...

- [ ] Feature request

### Issue description

I am trying to achieve something like below:
<img width="288" alt="image" src="https://user-images.githubusercontent.com/2225196/67260687-40d05400-f4ba-11e9-8188-e1aab3e92953.png">

what I am able to achieve with current carbon-charts is below:
![image](https://user-images.githubusercontent.com/2225196/67260610-d8817280-f4b9-11e9-90f7-353d4b257db6.png)

so , what can I do to closely match with the expected behaviour shown in first snapshot ?

1) How can I use to reduce the gap between the donut chart and labels on the right ?
2) How can I use to remove the labels (%) displayed at the outer of donut chart ?
3) What config can I use to remove labels displayed on the right ?
4) How can I remove the grid in bar chart ?
5) Labels/legends should always appear to the right of donut chart, and not like whats shown in second snapshot with `+` sign at the top right of donut chart

theiliad:
This is mostly up for @cameroncalder's review. Would we want to allow for any of those customizations?

scottdickerson:
@theiliad with the new legend options in 0.16.24 version, can we just add a "disabled" legend option to achieve?

theiliad:
> @theiliad with the new legend options in 0.16.24 version, can we just add a "disabled" legend option to achieve?

yeah we could, it's just a matter of design approval for disabling legends

jagdishkumar1:
Similar requirement on line chart as well .. 

Expectation:
<img width="542" alt="image" src="https://user-images.githubusercontent.com/2225196/67929134-a02d1300-fbe2-11e9-926c-1c56ed103745.png">

jagdishkumar1:
Is there any config which can support the positioning on legends in any chart ?

Expectation : 
<img width="244" alt="image" src="https://user-images.githubusercontent.com/2225196/67943817-edb87880-fc00-11e9-98f5-7aff7b1ef617.png">

Here, legends are positioned at bottom of bar chart

cameroncalder:
@theiliad @jagdishkumar1  Many of the points above are quite acceptable and should allow for configuration. Particularly the following:

Removing the legend has appropriate use cases. For example, if there is only a single data category being visualized and the title of the chart provides enough context.

[...]

PR #423, opened 2019-12-05
feat(legend): support hiding legend with new 'enabled' property

scottdickerson:
### Updates
- added new 'enabled' property to hide the legend if set false
- here https://github.com/carbon-design-system/carbon-charts/issues/379

- [ ] Demos all features
- [ ] Documented/annotated
- [ ] Matches UI/UX specs
- [ ] Accessible


packages/core/src/chart.ts
```diff
@@ -171,7 +171,7 @@ export class Chart {
 					this.model,
 					this.services,
 					[
-						legendComponent,
+						...((this.model.getOptions().legend.visible !== false) ? [ legendComponent ] : [ ]),
```
natashadecoste: I'd prefer if you stored` this.model.getOptions().legend.visible` and added a condititional after line 149 so that the checks for layout directions are not performed (since they are unneeded if the chart is the only thing in the fullFrameComponent).

the conditional would then bypass the codeblock below if there isnt going to be a legend

```
const legendPosition = Tools.getProperty(this.model.getOptions(), "legend", "position");
		if (legendPosition === "left") {
			fullFrameComponentDirection = LayoutDirection.ROW;

			if (!this.model.getOptions().legend.orientation) {
				this.model.getOptions().legend.orientation = LegendOrientations.VERTICAL;
			}
		} else if (legendPosition === "right") {
			fullFrameComponentDirection = LayoutDirection.ROW_REVERSE;

			if (!this.model.getOptions().legend.orientation) {
				this.model.getOptions().legend.orientation = LegendOrientations.VERTICAL;
			}
		} else if (legendPosition === "bottom") {
			fullFrameComponentDirection = LayoutDirection.COLUMN_REVERSE;
		}
```

packages/core/src/axis-chart.ts
```diff
@@ -86,7 +86,7 @@ export class AxisChart extends Chart {
 					this.model,
 					this.services,
 					[
-						legendComponent,
+						...((this.model.getOptions().legend.visible !== false) ? [ legendComponent ] : [ ]),
```
natashadecoste: same as with chart.ts

scottdickerson:
@theiliad I was thinking that "enabled" meant "allow clickable entries in the legend", where "visible" is whether to hide the legend or not

theiliad:
> @theiliad I was thinking that "enabled" meant "allow clickable entries in the legend", where "visible" is whether to hide the legend or not

there's already a flag that called `clickable` that handles that. We can discuss the naming. I think `enabled` makes sense in general.

@natashadecoste @zvonimirfras @cal-smith feel free to chime in

cal-smith:
`visible` isn't terrible (maybe not 100% accurate since it seems like we're not just visually hiding the legend) ... If we're actually disabling the legend entirely `enabled` or `disabled` makes more sense I think 🤔

packages/core/src/chart.ts
scottdickerson: ok this should be fixed now

packages/core/src/axis-chart.ts
```diff
@@ -60,23 +60,27 @@ export class AxisChart extends Chart {
...
 		};
 
+		const isEnabled = this.model.getOptions().legend.enabled !== false;
+
 		// Decide the position of the legend in reference to the chart
 		let fullFrameComponentDirection = LayoutDirection.COLUMN;
-		const legendPosition = Tools.getProperty(this.model.getOptions(), "legend", "position");
-		if (legendPosition === LegendPositions.LEFT) {
-			fullFrameComponentDirection = LayoutDirection.ROW;
+		if (isEnabled) {
```
theiliad: So pending some additions I need to make to the `Chart` class (being able to modify the list of components in every render), we'd need to move the `isEnabled` logic into the actual legend component for now.

The reason being that in this case if you do enable the legend after having disabled it, it won't take effect if the logic is inside this file.

So the way it would work for now is the legend component would always get included, and the data-binding should get rid of all the rendered elements inside once we do disable.

theiliad: Also `isEnabled` is a misleading var name considering that this is not the legend component. We should've used something more descriptive such as `isLegendEnabled`

scottdickerson: @theiliad is it possible to integrate this as is for the near term problem then migrate that code to Legend along with the Chart class changes?

theiliad: I wouldn't want to do that. Moving stuff into the legend component shouldn't take long. The other chart class changes don't yet affect us, and I will take care of those changes

scottdickerson:
@theiliad and I met and while this code is limited because switching the enabled option won't rerender, this also applies currently to all the existing layout options.  Those will be enhanced in a future PR and this will be refactored

PR #425, opened 2019-12-05
Meter chart

natashadecoste:
###  TODO
- [x] add storybook demos for Vue, React, Angular
- [x] update value bar and style to match status
- [x] cross browser compatibility 
- [x] remove duplicate getStatus code
- [ ] ~remove hex border color from scss~
- [x] change bar height
- [x] title truncation tooltips
- [ ] data out of status ranges 

- [ ] Demos all features
- [ ] Documented/annotated
- [ ] Matches UI/UX specs
- [ ] Accessible


natashadecoste:
Added support to have a meter chart exported with a regular chart title (using same options as other chart titles) if needed. 

Configurations:
image

Result:
image

packages/core/demo/demo-data/meter.ts
```diff
@@ -0,0 +1,37 @@
...
+				status: [
+					{ range: [0, 1000], status: "success" },
+					// { range: [500, 1000], status: "custom" },
+					{ range: [1000, 1200], status: "warning"},
+					{ range: [1200, 1500], status: "danger"}
+				],
+			}
+		}
+	]
+};
```
theiliad: If we're rendering only 1 meter when using the `MeterChart` class then why do we have the concept of datasets in here?

packages/core/demo/index.ts
```diff
@@ -127,6 +149,10 @@ const createChartContainer = chartType => {
 	// Chart holder
 	const holder = document.createElement("div");
 	holder.className = "demo-chart-holder has-actions";
+	// meter need a more compact chart holder
```
theiliad:
```suggestion
	// meter needs a more compact chart holder
```

packages/core/src/axis-chart.ts
```diff
@@ -63,61 +45,109 @@ export class AxisChart extends Chart {
...
-		if (legendPosition === LegendPositions.LEFT) {
-			fullFrameComponentDirection = LayoutDirection.ROW;
 
-			if (!this.model.getOptions().legend.orientation) {
-				this.model.getOptions().legend.orientation = LegendOrientations.VERTICAL;
-			}
-		} else if (legendPosition === LegendPositions.RIGHT) {
-			fullFrameComponentDirection = LayoutDirection.ROW_REVERSE;
+		// Check if the legend is enabled for the chart
+		if (this.model.getOptions().legend.enabled) {
```
theiliad: What happens if I disable legend in the initial render, then want to enable again?

packages/core/src/styles/graphs/_meter.scss
```diff
@@ -0,0 +1,12 @@
+.#{$prefix}--#{$charts-prefix}--meter {
+	rect.container {
+		fill: $ui-01;
+	}
```
theiliad:
```suggestion
	}

```

packages/core/src/styles/graphs/_meter.scss
```diff
@@ -0,0 +1,12 @@
...
+	rect.value {
+		border-right: 2px;
+	}
```
theiliad:
```suggestion
	}

```

packages/core/src/styles/components/_meter-title.scss
```diff
@@ -0,0 +1,23 @@
...
+		fill: $text-01;
+		font-size: 16px;
+		font-family: carbon--font-family("sans");
+		font-weight: carbon--font-weight("semibold");
+	}
+
+	circle.status-indicator {
+		&.status--danger {
+			fill: $support-01;
+		}
```
theiliad:
```suggestion
		}

```

packages/core/src/styles/components/_meter-title.scss
```diff
@@ -0,0 +1,23 @@
...
+		&.status--warning {
+			fill: $support-03;
+		}
```
theiliad:
```suggestion
		}

```

packages/core/src/styles/components/_meter-title.scss
```diff
@@ -0,0 +1,23 @@
...
+		&.status--success {
+			fill: $support-02;
+
```
theiliad:
```suggestion
```

packages/core/src/model.ts
```diff
@@ -176,7 +176,9 @@ export class ChartModel {
 	setColorScale() {
 		if (this.getDisplayData().datasets[0].fillColors) {
 			this.getDisplayData().datasets.forEach(dataset => {
-				this.colorScale[dataset.label] = scaleOrdinal().range(dataset.fillColors).domain(this.allDataLabels);
+				if (this.allDataLabels) {
+					this.colorScale[dataset.label] = scaleOrdinal().range(dataset.fillColors).domain(this.allDataLabels);
+				}
```
theiliad: Seems like we just need a separate `MeterModel` class that extends `Model` so we don't get the additions here

packages/core/src/interfaces/charts.ts
```diff
@@ -147,3 +147,29 @@ export interface DonutChartOptions extends PieChartOptions {
 		};
 	};
 }
+
+export interface MeterChartOptions extends BaseChartOptions {
+	meter?: {
+		barHeight?: number,
```
theiliad: Wouldn't this just be called `height`? So you'd access it as `chartOptions.meter.height`

packages/core/src/components/graphs/meter.ts
```diff
@@ -0,0 +1,98 @@
...
+	type = "meter";
+
+	render(animate = true) {
+		const self = this;
+		const svg = this.getContainerSVG();
+		const options = this.model.getOptions();
+		const displayData = this.model.getDisplayData();
+
+		// meter only deals with 1 dataset (like pie/donut)
+		const dataset = displayData.datasets[0];
```
theiliad: If we deal with 1 dataset only then we should create a separate model that extends `ChartModel` and remove the concept of `datasets`

packages/core/src/components/graphs/meter.ts
```diff
@@ -0,0 +1,98 @@
...
+			.attr("x", 0 )
+			.attr("y", 0 )
+			.attr("width", options.width)
+			.attr("height", options.meter.barHeight);
+
+		// rect with the value binded
+		const value = svg.selectAll("rect.value")
+			.data([dataset]);
+
+		value.exit().remove();
```
theiliad: Why would we ever need to remove on exit? Don't we always have 1 bar showing?

packages/core/src/components/graphs/meter.ts
```diff
@@ -0,0 +1,98 @@
...
+			.classed("value", true)
+			.merge(value)
+			.attr("x", 0 )
+			.attr("y", 0 )
+			.transition(this.services.transitions.getTransition("meter-bar-update", animate))
+			.attr("width", d =>  xScale(d.data.value))
+			.attr("height", options.meter.barHeight)
+			.attr("fill", d => d.fillColors[0])
+			.attr("fill-opacity", 0.5)
+			.style("border-color", d => d.fillColors[0]);
```
theiliad: If we have 1 dataset then we shouldn't have a `fillColors` array

packages/core/src/components/graphs/meter.ts
```diff
@@ -0,0 +1,98 @@
...
+			.merge(border)
+			.attr("y1", 0)
+			.attr("y2", options.meter.barHeight)
+			.transition(this.services.transitions.getTransition("meter-bar-update", animate))
+			.attr("x1", d => xScale(d.data.value))
+			.attr("x2", d => xScale(d.data.value))
+			.attr("stroke-width", 2)
+			.attr("stroke",  d => {
+				return d.fillColors[0];
+			});
```
theiliad:
```js
.attr("stroke",  d => d.fillColors[0]});
```

packages/core/src/components/graphs/meter.ts
```diff
@@ -0,0 +1,98 @@
...
+
+		// if a peak is supplied within the domain, we want to render it
+		const peak = DOMUtils.appendOrSelect(svg, "line.peak");
+		if (dataset.data.peak && dataset.data.peak >= dataset.data.min &&  dataset.data.peak <= dataset.data.max ) {
+			const peakVal = dataset.data.peak;
```
theiliad: `peakVal` can be initialized before the if-statement so we can use the variable in there as well. Also let's use the full word `peakValue`

packages/core/src/components/graphs/meter.ts
```diff
@@ -0,0 +1,98 @@
...
+			// transitions to its correct location
+			peak
+				.transition(this.services.transitions.getTransition("peak-line-update", animate))
+				.attr("x1", xScale(peakVal))
+				.attr("x2", xScale(peakVal));
+		} else {
+			// remove any stale peak indicators
+			peak.remove();
+		}
+	}
```
theiliad: Can't this entire if-statement be avoided if we use d3 `.data()`, `.enter()` & `.exit()`?

packages/core/src/components/graphs/meter.ts
```diff
@@ -0,0 +1,98 @@
+// Internal Imports
+import { Component } from "../component";
+import { DOMUtils } from "../../services";
+import { Tools } from "../../tools";
+
+// D3 Imports
+import { select } from "d3-selection";
+import { scaleLinear } from "d3-scale";
+
+
```
theiliad:
```suggestion
```

packages/core/src/charts/meter.ts
```diff
@@ -0,0 +1,108 @@
...
+	model = new ChartModel(this.services);
+
+	// TODO - Optimize the use of "extending"
+	constructor(holder: Element, chartConfigs: ChartConfig<MeterChartOptions>, extending = false) {
+		super(holder, chartConfigs);
+
+		// TODO - Optimize the use of "extending"
+		if (extending) {
+			return;
+		}
```
theiliad: We don't need this here

packages/core/src/charts/meter.ts
```diff
@@ -0,0 +1,108 @@
...
+	LayoutComponent,
+	TitleMeter,
+	Spacer
+} from "../components/index";
+
+export class MeterChart extends Chart {
+	model = new ChartModel(this.services);
+
+	// TODO - Optimize the use of "extending"
+	constructor(holder: Element, chartConfigs: ChartConfig<MeterChartOptions>, extending = false) {
```
theiliad:
```suggestion
	constructor(holder: Element, chartConfigs: ChartConfig<MeterChartOptions>) {
```

packages/core/src/components/essentials/title-meter.ts
```diff
@@ -0,0 +1,177 @@
+// Internal Imports
+import { Component } from "../component";
+import { Title } from "./title";
+import { DOMUtils } from "../../services";
+import { TooltipTypes } from "./../../interfaces";
+import { selectAll } from "d3-selection";
+
+export class TitleMeter extends Title {
```
theiliad:
```suggestion
export class MeterTitle extends Title {
```

packages/core/src/components/essentials/title-meter.ts
```diff
@@ -0,0 +1,177 @@
...
+		} else {
+			// if we don't truncate we want to remove any events that might have existed for tooltips on the previous title
+			title
+			.on("mouseenter", null)
+			.on("mouseout", null);
+		}
+
+	}
+
+	render() {
```
theiliad: We should try to keep `render()` relatively at the top of every component class

packages/core/src/components/essentials/title-meter.ts
```diff
@@ -0,0 +1,177 @@
...
+		const containerBounds  = DOMUtils.getSVGElementSize(this.parent, {useAttr: true});
+		// need to check if the width is 0, and try to use the parent attribute
+		// this can happen if the chart is toggled on/off and the height is 0 for the parent, it wont validateDimensions
+		const containerWidth = containerBounds.width ? containerBounds.width : this.parent.node().getAttribute("width");
+
+		// remove any status indicators on the chart to re-render if ranges have been given
+		const status = DOMUtils.appendOrSelect(svg, "circle.status-indicator");
+
+		// if ranges are provided, and the status indicator is enabled, update it
+		if (dataset.data.status && this.model.getOptions().meter.status.enabled) {
```
theiliad: `this.model.getOptions()` should be stored in a variable

theiliad: Also we should use `Tools.getProperty()` for longer reaching keys

packages/core/src/components/essentials/title-meter.ts
```diff
@@ -0,0 +1,177 @@
...
+		// size of the status indicator
+		const circleSize = this.model.getOptions().meter.status.indicatorSize;
+
+		const self = this;
+		status.attr("cx", containerWidth - circleSize)
+			.attr("cy", 20 - circleSize)
+			.attr("r", circleSize)
+			.attr("class", function() {
```
theiliad:
```js
.attr("class", `status-indicator status--${self.getStatus(dataset.data.value, dataset)}`);
```

packages/core/src/components/essentials/title-meter.ts
```diff
@@ -0,0 +1,177 @@
...
+				return `status-indicator status--${self.getStatus(dataset.data.value, dataset)}`;
+			});
+		} else {
+			// remove any stale status indicators
+			status.remove();
+		}
+	}
```
theiliad: Is there any way to avoid the if-else using d3 rendering logic?

packages/core/src/components/essentials/title-meter.ts
```diff
@@ -0,0 +1,177 @@
...
+		// the value of the meter converted to percent
+		const value = Math.round(dataset.data.value / dataset.data.max * 100);
+
+		// use the title's position to append the percentage to the end
+		const svg = this.getContainerSVG();
+		const title = DOMUtils.appendOrSelect(svg, "text.title");
+		const percentage = DOMUtils.appendOrSelect(svg, "text.percent-value");
+
+		// append a percentage if it is enabled, update it
+		if (this.model.getOptions().meter.title.percentageIndicator.enabled) {
```
theiliad: `Tools.getProperty()`

packages/core/src/components/essentials/title-meter.ts
```diff
@@ -0,0 +1,177 @@
...
+		// the horizontal offset of the percentage value from the title
+		const offset = this.model.getOptions().meter.title.paddingRight;
```
theiliad: `Tools.getProperty()`

theiliad: Also there are some incorrect indentations

packages/core/src/components/essentials/title-meter.ts
```diff
@@ -0,0 +1,177 @@
...
+				self.services.events.dispatchEvent("show-tooltip", {
+					hoveredElement: title,
+					type: TooltipTypes.TITLE
+				});
+			})
+			.on("mouseout", function() {
+				self.services.events.dispatchEvent("hide-tooltip", {
+					hoveredElement: title,
+				});
+			});
```
theiliad: We should be re-using truncation logic between title classes

packages/core/demo/demo-data/meter.ts
natashadecoste: too try and keep the data layer as uniform as possible. The same way pie/donut use datasets but only render the first dataset as a visualisation. Also I kept fillColors using an array of colors even though meter will only use the first color.

packages/core/demo/demo-data/meter.ts
```diff
@@ -0,0 +1,22 @@
...
+		max: 1500,
+		peak: 1200,
+		value: 1000,
+		status: [
+			{ range: [0, 1000], status: "success" },
+			// { range: [500, 1000], status: "custom" },
+			{ range: [1000, 1200], status: "warning"},
+			{ range: [1200, 1500], status: "danger"}
+		],
+	}
```
cal-smith: Some of these seem more option-y than data-y to me ... Is there any precedent for option-like data being used as part of the `data` object?

I guess really the question is, how much config should the user be expected to pass in when they call `setData` ... is it likely that the `status`, `min`, or `max` is going to be changing frequently? Since I'd expect a gauge like this to be receiving new data `value`s on a fairly regular basis

theiliad: I think `min`, `max`, `peak` and `status` should be in the chart options, and the chart data would basically just be a `number`.

Options could be under `meter`

theiliad:
*Items to consider here:*
- Icons that signify the `status`
- Percentage font size should be the same as the title

jeanservaas:
Hey @theiliad 

Let's use the icons from the notifications at 16px for the meter


This icon for danger:

image

This icon for warning (exclamation has to be black for accessibility):

image

This for success:

image

theiliad:
@natashadecoste

jeanservaas:
After a discussion on the Data viz guild call and talking with @theiliad -- seems like a good idea to use the alert color in the body of the meter as well as the alert icon. 

If there are instances where a product doesn’t want that… a unique color should be the override, not the default

natashadecoste:
updated to have the new bar height 8px, as well as 8px spacing between meter title and bar

image

designertyler:
@natashadecoste I noticed the icon hangs beyond the bar by 3px. In some of the specs posted it might have been nudged to do that. Can we bring the icon to the left 3px so the right edges of the icon and bar are aligned?

image

jeanservaas:
@natashadecoste 

- looks like the spacing is a little tight from the label

- and Tyler's right that icon shouldn't hang over the edge of the meter, they should right align

image


The only other question I had was about this moment — the other colors have that 50% opacity stroke at the end of the color bar. 

What is that stroke supposed to do? Is it a problem that we don't have it on the yellow one... because of course, it would be weird to add something like that when there is already a stroke... 

@designertyler suggested it might have been added for contrast accessibility on the G10 bar... if that's true then the yellow would be fine without it.

image

natashadecoste:
@jeanservaas @designertyler 
I'm still a bit unclear about the right border, are we keeping it for the green (success) and red (warning) colors? Do we also apply the border if the chart is defaulting back to a user provided color/carbon charts default color? 

Also the implementation of the border is now that it is the same color as the value bar but we darken the color by a certain value (currently 10).  There isnt any opacity, but i can change it back to being opacity so that it appears as a lighter border on light themes and darker on dark ones?

jeanservaas:
Hi @natashadecoste,

I just wanted to add an updated spec here with colors, I realized I never gave you anything in the dark themes. This spec specifies when to use what tokens where and also where there are not tokens. Thanks for being patient! Most of this stuff I think you've already figured out and It looks like it's reflected in code, but I wanted to document the spec and also wanted to see what you thought about the lighter strokes in the peak usage variant in the dark themes.

I know you're using transparencies to make those in the light themes... but I don't think that would work for the dark themes. Again... we may not even need these strokes... I have a feeling that they were just added for contrast accessibility for the gray 20 background. But I checked with Sadek and green and red 60 should be 3:1 accessible on a gray 20 background. So maybe we can just kill them.

<br>

image

[meter_spec_file.zip](https://github.com/carbon-design-system/carbon-charts/files/4506139/meter_spec_file.zip)

theiliad:
can't see the deploy previews

packages/core/src/components/graphs/meter.ts
```diff
@@ -0,0 +1,88 @@
...
+	type = "meter";
+
+	render(animate = true) {
+		const self = this;
+		const svg = this.getContainerSVG();
+		const options = this.model.getOptions();
+		const data = this.model.getDisplayData();
+		const status = this.model.getStatus();
+
+		const width = DOMUtils.getSVGElementSize(this.parent, { useAttrs: true }).width;
```
theiliad, line 20:
```suggestion
		const { width } = DOMUtils.getSVGElementSize(this.parent, { useAttrs: true });
```

packages/core/src/components/essentials/title-meter.ts
```diff
@@ -0,0 +1,184 @@
...
+		const statusWidth = DOMUtils.getSVGElementSize(statusGroup, { useBBox: true }).width + Tools.getProperty(this.model.getOptions(), "meter", "status", "paddingLeft");
+
+		return containerWidth - percentageWidth - offset - statusWidth;
+	}
+
+	/**
+	 * Get the associated status icon for the data
+	 * @param status the active status for the meter chart
+	 */
+	protected getStatusIcon(status) {
```
theiliad, line 174:
```suggestion
	protected getStatusIconPathString(status) {
```

packages/core/demo/data/meter.ts
```diff
@@ -0,0 +1,47 @@
...
+	title: "Meter Chart - with statuses",
+	meter: {
+		peak: 80,
+		status: {
+			ranges: [
+				{ range: [0, 50], status: "success" },
+				{ range: [50, 60], status: "warning" },
+				{ range: [60, 100], status: "danger" }
+			]
+		}
```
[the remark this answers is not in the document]
natashadecoste: yep no status means no icon and the bar color will be the user provided one or carbon dataviz palette color
no peak will not display a peak (just saw it is displaying as 0  which it shouldnt and pushed a fix)

packages/core/src/components/essentials/title-meter.ts
```diff
@@ -0,0 +1,184 @@
...
+	 */
+	appendPercentage() {
+		const dataValue = this.model.getDisplayData().value;
+
+		// use the title's position to append the percentage to the end
+		const svg = this.getContainerSVG();
+		const title = DOMUtils.appendOrSelect(svg, "text.meter-title");
+
+		// check if it is enabled
+		const data = Tools.getProperty(this.model.getOptions(), "meter", "title", "percentageIndicator", "enabled") === true ?
```
Donisius, line 109: Would this work?
```suggestion
		const data = Tools.getProperty(this.model.getOptions(), "meter", "title", "percentageIndicator", "enabled") ?
```
or
```suggestion
		const data = !!Tools.getProperty(this.model.getOptions(), "meter", "title", "percentageIndicator", "enabled") ?
```

packages/core/src/components/essentials/title-meter.ts
```diff
@@ -0,0 +1,184 @@
...
+	truncateTitle(title, maxWidth) {
+		super.truncateTitle(title, maxWidth);
+
+		// update the position on the percentage to be inline with the title
+		const tspan = DOMUtils.appendOrSelect(this.parent, "tspan");
+		const offset = this.model.getOptions().meter.title.paddingRight; // horizontal offset of percent from title
+		const tspanLength = Math.ceil(tspan.node().getComputedTextLength());
+
+		const percentage = DOMUtils.appendOrSelect(this.parent, "text.percent-value");
+		percentage.attr("x", +title.attr("x") + title.node().getComputedTextLength() + tspanLength + offset);
```
Donisius, line 145:
```suggestion
		percentage.attr("x", + title.attr("x") + title.node().getComputedTextLength() + tspanLength + offset);
```

packages/core/src/components/essentials/title-meter.ts
theiliad: It's just better typed this way, since many other values will pass with your proposed way
image

packages/core/src/components/essentials/title-meter.ts
theiliad: usually they do +varName

packages/core/src/components/essentials/title-meter.ts
Donisius: Okay

theiliad:
this looks better, however when removing `height` you still get a fairly tall chart. Thinking about the user putting meter into a dashboard card. Vertically they'd want the chart holder to end where the final meter rectangle ends, correct? How can we make that happen? I know it's difficult in this case...

packages/core/src/components/essentials/title-meter.ts
```diff
@@ -0,0 +1,183 @@
...
+		this.appendPercentage();
+
+		// if status ranges are provided (custom or default), display indicator
+		this.displayStatus();
+
+		// get the max width of a title (with consideration for the status/percentage)
+		const maxWidth = this.getMaxTitleWidth();
+		const titleElement = DOMUtils.appendOrSelect(svg, "text.meter-title");
+
+		if (maxWidth > 0 && titleElement.node().getComputedTextLength() > maxWidth) {
```
Donisius, line 40: maybe
```suggestion
		if (titleElement.node().getComputedTextLength() > maxWidth > 0) {
```

packages/core/src/components/essentials/title.ts
```diff
@@ -6,74 +6,94 @@ import { TooltipTypes, Events } from "./../../interfaces";
...
+			.attr("x", 0)
+			.attr("y", "1em")
+			.html(d => d);
+
+		// check the max space the title has to render
+		const maxWidth = this.getMaxTitleWidth();
+		const title = DOMUtils.appendOrSelect(svg, "text.title");
+
+		// check if title needs truncation (and tooltip support)
+		if (title.node().getComputedTextLength() > maxWidth) {
```
Donisius, line 28: Should we also check if it's `> 0`?
```suggestion
		if (title.node().getComputedTextLength() > maxWidth > 0) {
```

packages/core/src/components/essentials/tooltip.ts
```diff
@@ -143,28 +143,32 @@ export class Tooltip extends Component {
...
 	getTooltipHTML(data: any, type: TooltipTypes) {
-		// check if it is getting styles for a tooltip type
+		// title tooltips
 		if (type === TooltipTypes.TITLE) {
-			const title = this.model.getOptions().title;
-			return `<div class="title-tooltip"><text>${title}</text></div>`;
+			// title is passed in the event details
+			// meter title uses the dataset title
+			const title = data;
+			return `<div class="title-tooltip"><p>${title}</p></div>`;
```
Donisius, line 153: Why not just
```suggestion
			return `<div class="title-tooltip"><p>${data}</p></div>`;
```
?

packages/core/src/model-meter.ts
```diff
@@ -0,0 +1,70 @@
...
+	 * Get the associated status for the data by checking the ranges
+	 */
+	getStatus() {
+		const options = this.getOptions();
+		const dataValue = this.getDisplayData().value;
+
+		// user needs to supply ranges
+		const allRanges = Tools.getProperty(options, "meter", "status", "ranges");
+		if (allRanges) {
+			const result = allRanges.filter(step => (step.range[0] <= dataValue && dataValue <= step.range[1]));
```
Donisius, line 60:
```suggestion
			const result = allRanges.filter(step => step.range[0] <= dataValue <= step.range[1]);
```

packages/core/src/components/essentials/title-meter.ts
Donisius: Actually this won't work.

packages/core/src/services/essentials/dom-utils.ts
```diff
@@ -203,6 +204,25 @@ export class DOMUtils extends Service {
...
+			let childrenHeight = 0;
+			children.nodes().forEach(function (childSVG) {
+				childrenHeight += Number(DOMUtils.getSVGElementSize(select(childSVG), {useBBox : true}).height);
+			});
+
+			// set the chart svg height to the children height
+			// forcing the chart not to take up any more space than it requires
+			childrenHeight <= chartHeight ?  select(this.svg).attr("height", childrenHeight) : select(this.svg).attr("height", "100%") ;
+		}
+	}
```
theiliad, line 224: can we only enable this for meter? as in have MeterChart call the function?

packages/core/src/model.ts
```diff
@@ -389,6 +389,13 @@ export class ChartModel {
 		return this.colorScale;
 	}
 
+	/**
+	 * For charts that might hold an associated status for their dataset
+	 */
+	getStatus() {
+		return null;
+	}
```
theiliad, line 397: do we really need this for the base model class?

packages/core/src/model-meter.ts
```diff
@@ -0,0 +1,70 @@
...
+			return super.getFillColor(group);
+		} else {
+			return null;
+		}
+	}
+
+	/**
+	 * Get the associated status for the data by checking the ranges
+	 */
+	getStatus() {
```
theiliad, line 53: is this something that gauge also needs? can we somehow share?

packages/core/src/components/graphs/meter.ts
```diff
@@ -0,0 +1,90 @@
...
+			// a11y
+			.attr("role", Roles.GRAPHICS_SYMBOL)
+			.attr("aria-roledescription", "value")
+			.attr("aria-label", d => d.value);
+
+		// draw the peak
+		const peakValue = Tools.getProperty(options, "meter", "peak") ?? null;
+
+		// update the peak if it is less than the value, it should be equal to the value
+		const updatedPeak = peakValue !== null && peakValue < dataset.value ? dataset.value : peakValue;
```
theiliad, line 66:
```suggestion
		const updatedPeak = (peakValue !== null && peakValue < dataset.value) ? dataset.value : peakValue;
```

packages/core/src/components/graphs/meter.ts
```diff
@@ -0,0 +1,90 @@
...
+		// dont display peak if there isnt one
+		const peakData = updatedPeak === null ? [] : [updatedPeak];
```
theiliad, line 68:
```suggestion
		const peakData = (updatedPeak === null) ? [] : [updatedPeak];
```

packages/core/src/components/graphs/meter.ts
```diff
@@ -0,0 +1,90 @@
...
+
+		// draw the container to hold the value
+		DOMUtils.appendOrSelect(svg, "rect.container")
+			.attr("x", 0 )
+			.attr("y", 0 )
+			.attr("width", width)
+			.attr("height", Tools.getProperty(options, "meter", "height"));
+
+		// value larger than 100 will display as 100% on meter chart
+		const dataset = data.value <= 100 ? data : data["value"] = 100;
```
theiliad, line 36:
```suggestion
		const dataset = (data.value <= 100) ? data : data["value"] = 100;
```

packages/core/src/components/graphs/meter.ts
```diff
@@ -0,0 +1,90 @@
...
+			.transition(this.services.transitions.getTransition("meter-bar-update", animate))
+			.attr("width", d => xScale(d.value))
+			.attr("fill", d => self.model.getFillColor(d[groupMapsTo]))
+			// a11y
+			.attr("role", Roles.GRAPHICS_SYMBOL)
+			.attr("aria-roledescription", "value")
+			.attr("aria-label", d => d.value);
+
+		// draw the peak
+		const peakValue = Tools.getProperty(options, "meter", "peak") ?? null;
```
theiliad, line 63: what's does this do? `Tools.getProperty(options, "meter", "peak") ?? null`

packages/core/src/model-meter.ts
natashadecoste: i dont think that we could share because the status for meter is based on the ranges provided in options, but gauge has so many use cases that the status is controlled by the user completely (its in the options). for ex. the delta could be positive or negative but the status could be success or warning depending on the case

packages/core/src/components/graphs/meter.ts
natashadecoste: if there is a peak supplied, it is assigned to the `peakValue`, otherwise if it is undefined or null it will be default assigned to `null`(and chart wont render a peak)

packages/core/src/model.ts
natashadecoste: yeah so that i can extend it in model-meter?

packages/core/src/components/graphs/meter.ts
theiliad: `Tools.getProperty` will already return `null` if it fails anywhere in its process

PR #430, opened 2019-12-10
fix(core): fix to allow chart to toggle on and off 

natashadecoste:
fix the stack overflow error that happens when charts are toggled off and on (using html attributes or css properties).

the getSVGElementSIze returns `{width: 0, height:0}` because the height is 0 (chart is still in loading and rendering stage). Added a check for if the attribute dimensions arent validated and 0 or null or false is returned for width. 

to see the original error:
1. you need a chart with a longer title (that will be truncated)
2. toggle chart off in the console: ` document.getElementById("classy-pie-chart-holder").style.display = "none";
`
3. toggle chart back on in the console: 
` document.getElementById("classy-pie-chart-holder").style.display = "block";
`

image

- [ ] Demos all features
- [ ] Documented/annotated
- [ ] Matches UI/UX specs
- [ ] Accessible


packages/core/src/components/essentials/title.ts
```diff
@@ -35,17 +43,17 @@ export class Title extends Component {
 
 			// add events for displaying the tooltip with the title
 			const self = this;
-			title.on("mouseenter", function() {
+			title.on("mouseenter", function () {
```
theiliad: why do we need a space between `function` and the brackets?

natashadecoste: i think my auto formatter did that, but not sure why cause it did it nowhere else ❓

cal-smith:
Toggling using the style inspector on a random chart seems to trigger some exceptions for me:
image
image

From schedule.js (part of d3 I think?) ... No logs in the console though, so I'm not totally sure where or what is breaking

theiliad:
> Toggling using the style inspector on a random chart seems to trigger some exceptions for me:
> image
> image
> 
> From schedule.js (part of d3 I think?) ... No logs in the console though, so I'm not totally sure where or what is breaking

@natashadecoste

natashadecoste:
> Toggling using the style inspector on a random chart seems to trigger some exceptions for me:
> image
> image
> 
> From schedule.js (part of d3 I think?) ... No logs in the console though, so I'm not totally sure where or what is breaking

@cal-smith What browser? I dont get anything in Chrome toggling the grouped-bar chart (with JS or the style inspector), do you remember which chart or were you getting this exception for all?

cal-smith:
@natashadecoste the first one in the demo - firefox dev edition, I also had pause on exceptions turned on in the debugger

theiliad:
Have we had any progress on this?

PR #434, opened 2019-12-11
fix(charts-angular): update build toolchain

cal-smith:
- switches the build to use the angular cli
- updates the tsconfig to a known working state (based on the default angular cli output)
- syncs all the build package versions
- fixes build for ng6


cal-smith:
Just doing some local validation:

- [x] ng6 works
- [ ] ng7 works
- [ ] ng8 works

cal-smith:
All good from a compatibility perspective ... now just sorting out the travis issues

cal-smith:
Travis sorted, seems like netlify was using an outdated node version, so I bumped that - the cache will need to be cleared though before it will build correctly (mainly for node-sass ... we should swap out to sass to avoid that in the future)

packages/angular/tsconfig.json
```diff
@@ -2,19 +2,35 @@
...
+    "declaration": true,
+	"moduleResolution": "node",
+	"module": "es2015",
     "emitDecoratorMetadata": true,
-    "experimentalDecorators": true,
-    "target": "es5",
+	"experimentalDecorators": true,
+	"downlevelIteration": true,
+	"importHelpers": true,
+    "target": "es2015",
```
theiliad: I think there are some mixed indentation types in this file

.travis/before_install.sh
```diff
@@ -2,4 +2,4 @@
 
 set -e
 
-yarn global add lerna@3.13.4
+# yarn global add lerna@3.13.4
```
theiliad: we needed this in the past? Is that not the case anymore?

.travis.yml
```diff
@@ -1,22 +1,22 @@
...
+  - curl -o- -L https://yarnpkg.com/install.sh | bash -s -- --version 1.21.1
   - export PATH="$HOME/.yarn/bin:$PATH"
   - .travis/before_install.sh
   - "google-chrome-stable --headless --disable-gpu --remote-debugging-port=9222 http://localhost &"
 install: yarn --offline
 script:
-  - lerna run lint
-  - lerna run test
+  - yarn run lint
+  - yarn run test
```
theiliad: would yarn workspaces pick that up and run in all packages?

.nvmrc
```diff
@@ -0,0 +1 @@
+11
```
theiliad: does this tell travis what node version to use?

.travis/before_install.sh
cal-smith: I swapped the travis commands to just use the root package.json scripts 👍

.nvmrc
cal-smith: It's for netlify ... as far as I can see it's the only way to specify the node version there ... travis uses the `.travis.yml`

.travis/before_install.sh
theiliad: So let's remove that line completely then?

.travis.yml
cal-smith: That runs the scripts from the root package.json, which in turn run `lerna run x` ... so yeah ... this just makes the environment more consistent

packages/angular/tsconfig.json
cal-smith: 🤦‍♂

.travis/before_install.sh
cal-smith: I can remove the commented out command

cal-smith: Sorry, submitted after you commented ... I'll do that 👍

theiliad:
closing & openning the PR only causes travis to rebuild. could you confirm locally that things work?  I don't think this is an issue

theiliad:
Ran the `demo:build` script locally, works fine

cal-smith:
> closing & openning the PR only causes travis to rebuild. could you confirm locally that things work? I don't think this is an issue

Yeah it's just netlify exploding since the node version changed ... Its trying to run node 8 node-sass on node 11

PR #437, opened 2019-12-12
fix(axis): handle small scales of data 

### LINKED ISSUE #403: [LineChart]: the default date range for one days worth of data is too wide
#### I am submitting a...

- [ ] Feature request
- [ ] Design defect
- [x] Source code defect
- [ ] Other

#### charts version:
0.16.25

### Issue description
If I create a timeseries chart and give it 24 hours of data (one per hour), the x-axis crosses 3 day boundaries so there's lots of wasted space to the left and the right.

### Screenshot or recording
image

scottdickerson:
### Updates
- add(axis.ts): more specific calculation of domain ranges for different scales of timeRange
- fixes #403 

image

- [ ] Demos all features
- [ ] Documented/annotated
- [ ] Matches UI/UX specs
- [ ] Accessible


packages/core/src/components/axes/axis.ts
```diff
@@ -131,16 +148,25 @@ export class Axis extends Component {
 
 		if (axisOptions.scaleType === ScaleTypes.TIME) {
 			if (Tools.getProperty(options, "timeScale", "addSpaceOnEdges")) {
-				// TODO - Need to account for non-day incrementals as well
-				const [startDate, endDate] = domain;
-				startDate.setDate(startDate.getDate() - 1);
-				endDate.setDate(endDate.getDate() + 1);
+				const startDate = new Date(domain[0]);
+				const endDate = new Date(domain[1]);
+				if (differenceInYears(endDate, startDate) > 3) {
```
theiliad: we should move `3` into `configuration.ts` and an interface

scottdickerson: ok added!

packages/core/src/interfaces/axis-scales.ts
```diff
@@ -57,4 +57,5 @@ export interface AxesOptions {
  */
 export interface TimeScaleOptions {
 	addSpaceOnEdges?: boolean;
+	timeRangeToExtend?: number;
```
theiliad: This is confusing now, if `timeRangeToExtend` `"Only extends the axes if the timerange is greater than this amount"`, then what will `addSpaceOnEdges` do if the timerange is not greater than the one specified in `timeRangeToExtend`

scottdickerson: it will not add space on edges.  It will only do that if greater than the time range to extend

PR #453, opened 2019-12-19
chore(vue): add storybook knobs

natashadecoste:
### Updates
- knobs for the vue stories (data, options)

image

- [ ] Demos all features
- [ ] Documented/annotated
- [ ] Matches UI/UX specs
- [ ] Accessible


packages/vue/.storybook/config.js
```diff
@@ -1,10 +1,12 @@
...
 
-setOptions({
+withOptions({
   name: 'Carbon Charts - Vue Wrappers',
-  showDownPanel: false,
-  showAddonPanel: false
+  sortStoriesByKind: true,
+  showDownPanel: true,
+  showAddonPanel: true,
+  panelPosition: 'bottom',
```
theiliad: extra comma

natashadecoste:
> Also changing any of the data or option knobs don't result in anything.
> 
> Is that a storybook problem or vue wrappers problem?

Looking into it now

natashadecoste:
> Also changing any of the data or option knobs don't result in anything.
> 
> Is that a storybook problem or vue wrappers problem?

Seems others are having a problem with knobs when using data() for vue components as opposed to props. It seems to work and update nicely now with knobs, I'm not sure if you had a reason for using data() instead of props, let me know if that is the case.

theiliad:
> > Also changing any of the data or option knobs don't result in anything.
> > Is that a storybook problem or vue wrappers problem?
> 
> Seems others are having a problem with knobs when using data() for vue components as opposed to props. It seems to work and update nicely now with knobs, I'm not sure if you had a reason for using data() instead of props, let me know if that is the case.

I don't have enough vue domain knowledge to decide that. As long as we're able to update the knobs and get the chart re-rendered then it works

Issue #463, opened 2020-01-08
null values are being treated as zeros in LineChart


scottdickerson:
#### I am submitting a...

- [x] Feature request
- [ ] Design defect
- [ ] Source code defect
- [ ] Demo/documentation defect
- [ ] Other

#### charts version:
0.16.25
### Issue description
If I pass null values to a LineChart, instead of getting gaps in the data, the graph graphs the value as a 0 on the y-axis.

#### Steps to produce the issue
```
{
					date: new Date(2019, 0, 1),
					value: 10000
				},
				{
					date: new Date(2019, 0, 1),
					value: null
				},
```

#### Current behavior
Graphs as zeros

#### Expected behavior
I'm expecting something a little like this: 
https://www.carbondesignsystem.com/data-visualization/axes-and-labels/#gaps-in-data

### Screenshot or recording
![image](https://user-images.githubusercontent.com/6663002/71983707-f9ba4b00-31ec-11ea-9977-467b07b3f810.png)

scottdickerson:
@theiliad just an open question on this one, for a multi-line graph, would you only render the 'gap' in data if all of the values of a line were null for a x data point?

If one of the lines values was null for a timestamp and the rest of the lines actually had values, how would it render?

theiliad:
Just from a mathematical perspective I’d say if a point between 2 other
points is undefined, then there’s no real domain between them, hence you’d
get the gap to show.

anne-stevens:
Nulls should not be rendered as zeros.  A zero is an actual value that a user, sensor or system acutally input.  A null the absence of a value.  It is conceptually very different than zero.  In the line chart example, nulls should simply not be plotted.

gvnmagni:
We support what Anne is suggesting and, visually, we would remove any line between datapoints if there are missing data between them.

Technically, this issue https://github.com/carbon-design-system/carbon-charts/pull/464 shows the solution to our problem, it has already been implemented.

gvnmagni:
Russel, this image show the solution that was already developed for a different issue and that solves this one as well.

![72170731-65dcb080-339f-11ea-88a1-e1b91cb33c5e](https://user-images.githubusercontent.com/18484576/76977599-d0791200-6935-11ea-9329-7e242ef49048.png)

theiliad:
Closed by #464

Issue #299, opened 2019-07-09
[Restyle / enhancement] Time series


shixiedesign:
## Time series axis styling, updated (ready for Dev 🤖 )

Specs for axis styling for time series, time interval labeling. Added functionalities requested in https://github.com/carbon-design-system/carbon-charts/issues/280 , https://github.com/carbon-design-system/carbon-charts/issues/277 , https://github.com/carbon-design-system/carbon-charts/issues/258 .

![Timeseries-axis](https://user-images.githubusercontent.com/15144993/62570677-81261b80-b855-11e9-9c77-d4757785b512.png)

theiliad:
@shixiedesign Looks like the image you uploaded is blank

theiliad:
@shixiedesign The image you uploaded is blank

shixiedesign:
Corrected the image!

theiliad:
Looking great 🍺

shixiedesign:
### Updated time series spec: 

Cameron's back so we are going over this right now. Should be ready real sooooon! 

![timeseries](https://user-images.githubusercontent.com/15144993/62498243-2ed8f200-b7a4-11e9-92bc-592b5363f994.png)

scottdickerson:
@theiliad when is this coming? We could really use this in IoT

theiliad:
Hi @scottdickerson. Are we talking about axis breaks here or the entirety of the specs?

caesarsol:
Hi @theiliad, at Accurat we put some thoughts into the interpretation of this design spec, and this is to explain how we'd implement it on the development side.

We are only talking about the time axis problem, nothing has been thought about the axis breaks, on which we'd like to discuss. (Maybe on a different thread?)

From the design spec posted above, we understand:
  
 - The chart shows a certain number of ticks, depending on the user configuration or available space in pixels.
 - The ticks labels are different depending on the timespan of the data series. Not on the number or frequency of datapoints, but only on the distance between first and last temporal datums.
 - Every tick label has a specific formatting, depending on its position:
   - if it's in first place, a long label is shown.
   - if it's on the change of a time range (i.e. day, month, year), a long label is shown. The exact time range can be different for every case, detailed below.
   - in the remaining "normal" ticks, a short label is shown.

This is a table interpreting all the different formats in all cases:

```
15seconds:
  long: MMM DD, hh:mm:ss AM  // hours omitted if 00:00:00 (*)
  short: hh:mm:ss AM
  
[...]

PR #464, opened 2020-01-10
gaps within data

theiliad:
The below is all addressed according to Accurat.

> Related to #299 & #463.
> 
> **The specs above don't consider all the possible use cases**
> 
> In a case similar to this one:
> image
> 
> If you were looking at only 1 dataset you'd be able to draw a pattern over the gap to represent the "disabled" nature of that area within the domain:
> 
> image
> 
> However when dealing with multiple datasets that's not an option.
> 
> What would design suggest we do?


cal-smith:
Would deleting and recreating the core netlify site get the deploy preview working? It's very annoying 😛

SerenaG:
Hi @theiliad,
Accurat designers think that this solution for the #463 is the right one.
Do we need to work on this or it can just be merged?

theiliad:
> Hi @theiliad,
> Accurat designers think that this solution for the #463 is the right one.
> Do we need to work on this or it can just be merged?

We can merge it. Is it complete? Am I missing anything in the PR that we can add?

gvnmagni:
Just wanted to confirm that this approach is fine from the design perspective, please feel free to consider this the design solution for the gaps within data.

packages/core/src/components/graphs/line.ts
```diff
@@ -44,7 +44,14 @@ export class Line extends Component {
 		this.lineGenerator = line()
 			.x((d, i) => this.services.cartesianScales.getDomainValue(d, i))
 			.y((d, i) => this.services.cartesianScales.getRangeValue(d, i))
-			.curve(this.services.curves.getD3Curve());
+			.curve(this.services.curves.getD3Curve())
+      .defined((d: any, i) => {
```
cal-smith: some weird indentation here, is that intentional?
```suggestion
			.defined((d: any, i) => {
```

^ that _may_ fix it

packages/core/src/components/graphs/scatter.ts
```diff
@@ -106,13 +106,22 @@ export class Scatter extends Component {
...
+					return false;
+				}
+
+				return true;
+			})
+			.map((datum, i) => ({
+				date: datum.date,
+				label: labels[i],
+				datasetLabel: d.label,
+        class: datum.class,
```
cal-smith: This indentation is definitely not intentional 
```suggestion
				class: datum.class,
```


^ that _may_ fix it

packages/core/src/components/graphs/line.ts
theiliad: it's the github conflict resolver

packages/core/src/components/graphs/scatter.ts
theiliad: it's the github conflict resolver

packages/core/src/components/graphs/line.ts
```diff
@@ -44,7 +44,14 @@ export class Line extends Component {
...
+				if (!d || d.value === null) {
```
caesarsol: Might be useful to include the check to `undefined` as well.

I'm not sure about the `!d` check:
 - if `d` is always an object, it will always pass
 - if `d` can be a number, then this check is improvable because it will be incorrectly matched by 0
 - I don't believe `d` can be null or undefined

Thanks for any clarification :)

packages/core/src/components/graphs/scatter.ts
```diff
@@ -106,13 +106,22 @@ export class Scatter extends Component {
...
+				value: isNaN(datum) ? datum.value : datum
```
caesarsol: I think `datum` is always an object here, is that right?

packages/core/src/components/graphs/scatter.ts
```diff
@@ -106,13 +106,22 @@ export class Scatter extends Component {
...
-			date: datum.date,
-			label: labels[i],
-			datasetLabel: d.label,
-			class: datum.class,
-			value: isNaN(datum) ? datum.value : datum
-		}));
+		return d.data
+			// Remove datapoints with no value
+			.filter((datum: any) => {
+				if (!datum || datum.value === null) {
```
caesarsol: Might be useful to include the check to `undefined` as well.

packages/core/src/components/graphs/line.ts
cal-smith: Usually github only barfs like this when it's mixed tabs and spaces, but if you've checked and it's not then it's fine

packages/core/src/components/graphs/scatter.ts
theiliad: Not always, if you're using a discrete scale then datum will just be a number.

packages/core/src/components/graphs/scatter.ts
theiliad: 👍

packages/core/src/components/graphs/line.ts
theiliad: 👍

packages/core/src/components/graphs/scatter.ts
caesarsol: That's what I initially thought, but then I realized there is the access to `datum.class` in the line above. Should that be checked then?

PR #465, opened 2020-01-10
add event dispatching on mouseover, mousemove, click & mouseout for all graphs

theiliad:


cal-smith:
Also I don't see events for donut, does that rely on pie?

theiliad:
> I assume line will come later? Or does that rely on the scatter events?

So far nobody's asked for event dispatching on line (I think since lines are really thin surface areas for mouse interaction, not many consider them as a layer for the user to interact with)

> Also I don't see events for donut, does that rely on pie?
Yes that comes in with pie, causing the event name to still be `pie-slice-click`, although I think that's fine

packages/core/src/components/graphs/bar-stacked.ts
```diff
@@ -1,7 +1,7 @@
 // Internal Imports
 import { Tools } from "../../tools";
 import { Bar } from "./bar";
-import { Roles, ScaleTypes, TooltipTypes } from "../../interfaces";
+import { Roles, ScaleTypes, TooltipTypes, Events } from "../../interfaces";
```
zvonimirfras:
```suggestion
import {
	Roles,
	ScaleTypes,
	TooltipTypes,
	Events
} from "../../interfaces";
```

packages/core/src/components/graphs/pie.ts
```diff
@@ -2,7 +2,7 @@
 import { Component } from "../component";
 import { DOMUtils } from "../../services";
 import { Tools } from "../../tools";
-import { CalloutDirections, Roles, TooltipTypes } from "../../interfaces";
+import { CalloutDirections, Roles, TooltipTypes, Events } from "../../interfaces";
```
zvonimirfras:
```suggestion
import {
	CalloutDirections,
	Roles,
	TooltipTypes,
	Events
} from "../../interfaces";
```

cal-smith:
>So far nobody's asked for event dispatching on line (I think since lines are really thin surface areas for mouse interaction, not many consider them as a layer for the user to interact with)

The line points would likely still be useful ... perhaps `line-point-x` events, to distinguish them from scatter events

> Yes that comes in with pie, causing the event name to still be pie-slice-click, although I think that's fine

Slightly confusing, but we can always add specific variations later I suppose

theiliad:
> The line points would likely still be useful ... perhaps `line-point-x` events, to distinguish them from scatter events

Line points are part of the scatter graph, `scatter-mouseover`

packages/core/src/components/graphs/pie.ts
```diff
@@ -306,35 +311,54 @@ export class Pie extends Component {
...
+			})
+			.on("mouseout", function(datum) {
 				const hoveredElement = select(this);
 				hoveredElement.classed("hovered", false)
 					.transition(self.services.transitions.getTransition("pie_slice_mouseover"))
 					.attr("d", self.arc);
 
 				// Dispatch mouse event
-				self.services.events.dispatchEvent("pie-slice-mouseout", hoveredElement);
+				self.services.events.dispatchEvent(Events.Pie.SLICK_MOUSEOUT, {
```
cal-smith:
```suggestion
				self.services.events.dispatchEvent(Events.Pie.SLICE_MOUSEOUT, {
```

Typo?

packages/core/src/interfaces/events.ts
```diff
@@ -0,0 +1,29 @@
+/**
+ * enum of all pie graph events
+ */
+export enum Pie {
+	SLICE_MOUSEOVER = "pie-slice-mouseover",
+	SLICE_MOUSEMOVE = "pie-slice-mousemove",
+	SLICE_CLICK = "pie-slice-click",
+	SLICK_MOUSEOUT = "pie-slice-mouseout"
```
cal-smith: Same typo here ... totally missed this

cal-smith:
> Line points are part of the scatter graph, scatter-mouseover

Potentially confusing ... maybe adding an enum for line that just ends up pointing to the scatter enum values would be clearer? So `Line.POINT_CLICK` would have the same value as `Scatter.POINT_CLICK`

PR #477, opened 2020-01-24
Tiled maps

theiliad:


packages/geo/src/components/map.ts
```diff
@@ -0,0 +1,751 @@
+// Internal Imports
+import { Component } from "@carbon/charts/src/components/component";
```
cal-smith: I'm not sure how this will work as a deployed package ... We should probably declare a `devDependency`  and `peerDependency` on `@carbon/charts` and let lerna do it's linking thing ... It's still an "internal" import, but from a packager standpoint it still has to go through the standard exports

cal-smith: This will effectively compile down to `const Component = require("@carbon/charts/src/components/component");` which would then have the consumer transpiling charts core ... _or_ we would have to bundle the entirety of charts core as part of geo, which isn't ideal

theiliad: Yeah for sure

PR #479, opened 2020-01-30
build(core): remove node-sass, build styles using dartSass

theiliad:
- Removes node-sass
- Switches all Sass builds over to the native DartSass implementation


packages/react/.storybook/webpack.config.js
```diff
@@ -5,8 +5,9 @@ module.exports = ({ config }) => {
 			{
 				loader: "ts-loader?configFile=../../../core/tsconfig-demo.json"
 			}
-		]
+		],
```
zvonimirfras:
```suggestion
		]
```

PR #480, opened 2020-01-31
fix(styles): don't import and use carbon themes in styles.scss

### LINKED ISSUE #461: Carbon 10 transition via cloud PAL and the v9 theme is being overridden
#### I am submitting a...

- [ ] Feature request
- [x] Design defect
- [x] Source code defect
- [ ] Other

#### charts version:
v 0.11.15 shows the carbon 9 theme
v 0.19.x-0.24.x shows the carbon 10 theme

### Issue description
On transitioning our microservice to use Carbon 10 with the v9 theme, I noticed after upgrading to the latest version of carbon-charts, the microservice wasn't loading pages with css at all. 
Look at src/client/scss/app.scss in my branch and the package.json to verify that I am not doing something really off.

While debugging that issue I discovered that a reference to $text-05 could not be found when building the scss. After I swapped out our references to $text-05 the page loaded, but with the Carbon 10 theme instead of the v9 theme.

I started the process to migrate our microservice to Carbon 10. It is no small feat to reproduce.
This is the current progress: https://github.ibm.com/Bluemix/Gen1.VSI/pull/3667
It is not even 20% done.

Here's the migration guide I am following:
https://pages.github.ibm.com/ibmcloud/pal/cloudpal-adoption/phase-1-tutorial

Here's the v9 theme: https://github.com/carbon-design-system/carbon/blob/master/packages/themes/src/v9.js

#### Current behavior
On upgrading the carbon-charts-react version, the v9 theme stops working with PAL.

#### Expected behavior
We would be able to use the latest version of carbon-charts-react and still see the v9 theme for Carbon.

cal-smith:
closes #461 

Carbon automatically loads the default (white) theme, so the theme import was redundant in `styles.scss`, however, it did mean that it would overwrite any earlier imports and usages of any other themes (for example, the v9 theme) ... This introduces a `base-theme.scss` that forces the default theme and serves as a base for the g10/90/100 theme files. Users of custom themes should use `styles.scss` since it will use the colors and styles specified by that theme. Alternatively they could use one of the prebuilt `styles(-g10/90/100).css` if they want v10 chart styling with their custom theme.


theiliad:
There seems to be a bit of glitch now when changing the theme on the demo site. Can we fix that?

cal-smith:
> There seems to be a bit of glitch now when changing the theme on the demo site. Can we fix that?

For sure, I'll look into it 🤔

cal-smith:
> > There seems to be a bit of glitch now when changing the theme on the demo site. Can we fix that?
> 
> For sure, I'll look into it 🤔

@theiliad maybe I'm missing it, but I don't see anything broken on the demo site? The behaviour seems the same between the live site and the deploy preview

theiliad:
> > > There seems to be a bit of glitch now when changing the theme on the demo site. Can we fix that?
> > 
> > 
> > For sure, I'll look into it 🤔
> 
> @theiliad maybe I'm missing it, but I don't see anything broken on the demo site? The behaviour seems the same between the live site and the deploy preview

You're right it's the same as the demo site

theiliad:
@cal-smith could you please explain how users would now:
- Use all 4 themes through SCSS
- Customize each theme
?

cal-smith:
> Use all 4 themes through SCSS

Same as before - import `styles.scss` after applying the relevant carbon theme. They really shouldn't be using the `gXXX` files, since those are more for generating the `.css` files

> Customize each theme

AFAIK this isn't really documented for charts as is. As far as carbon themes are concerned, see above - they'd just follow the standard carbon theming guidelines, though they may also have to supply some custom variables specific to charts (unsure what those are, if any)

cal-smith:
Also check my initial comment in the PR for more detail

cal-smith:
big update! from the commit:

- adds a warning when `$carbon--theme` doesn't exist
- adds `styles-white.scss` as a base file for the `styles-xxx.scss` files
- improves standalone SCSS (vendored carbon, explicit variable imports)
- removes unneded imports from `styles.scss` (since that will be handled at the app level, or by one of the `styles-xxx.scss` files

cal-smith:
vendoring carbon gives us the ability to have truly dependency free SCSS files in the `styles-xxx.scss` files ... `styles.scss` now strongly depends on having the correct carbon environment set up, which should eliminate any (custom) theme errors

cal-smith:
@theiliad 

> Did you test the sass files with carbon & non-carbon environments?

Yep! Manually checked the build, and checked the demos - everything works as it should

> Also did you test the produced css files?

Manually verified, all the values look like they change in the right places thought

cal-smith:
@theiliad sorted the offline mirror nonsense ... should be 100% good to go now 👍

cal-smith:
> Although I'm not sure about adding a massive package to the vendors directory (carbon-components) the code LGTM.

I have a few reservations too, but it shouldn't be a _huge_ issue since none of it is bundled and shipped ... I could probably look into stripping it down a bit though, if we're really concerned about package sizes :+1:

zvonimirfras:
> Although I'm not sure about adding a massive package to the `vendors` directory (`carbon-components`) the code LGTM.
> 
> @zvonimirfras what do you think?

Same.

Issue #431, opened 2019-12-11
Chart animation-in direction


cameroncalder:
- [x] Scatter & bubble: animate the nodes starting from the x-axis and transitioning upward, instead of top-down.
- [ ] Line: animate the node in at the same placement as their related line instead of top-down to meet the line.
- [ ] Bar: animate the bars starting from the x-axis and transitioning upward, instead of top-down.

theiliad:
> * [ ]  Scatter & bubble: animate the nodes starting from the x-axis and transitioning upward, instead of top-down.

Fei is working on that.

> * [ ]  Line: animate the node in at the same placement as their related line instead of top-down to meet the line.
> * [ ]  Bar: animate the bars starting from the x-axis and transitioning upward, instead of top-down.

How necessary would you say these 2 tasks are? In order to get bars to render both vertically and horizontally we switched them from svg rectangles to paths, which means we'd need additional code in there to animate the height.

And in the case of line, line uses the scatter graph nodes which would animate in in their own way. Just trying to keep the code clean here and avoid adding a lot of extra logic. Is there a compromise we can make in terms of animating a lines in a different way?

PR #482, opened 2020-02-04
Scatter transition upward

sophiiae:
Solve #431: 
- [x] Scatter & bubble: animate the nodes starting from the x-axis and transitioning upward, instead of top-down.

### DIFF
--- a/packages/core/src/components/graphs/scatter.ts
+++ b/packages/core/src/components/graphs/scatter.ts
@@ -58,6 +58,7 @@ export class Scatter extends Component {
 			.classed("filled", filled)
 			.classed("unfilled", !filled)
 			.attr("cx", (d, i) => this.services.cartesianScales.getDomainValue(d, i))
+			.attr("cy", 311)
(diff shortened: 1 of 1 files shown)


packages/core/src/components/graphs/scatter.ts
```diff
@@ -58,6 +58,7 @@ export class Scatter extends Component {
 			.classed("filled", filled)
 			.classed("unfilled", !filled)
 			.attr("cx", (d, i) => this.services.cartesianScales.getDomainValue(d, i))
+			.attr("cy", 311)
```
natashadecoste, line 61:
```suggestion
			.attr("cy", this.services.cartesianScales.getRangeScale().range()[0])
```
If we use the scale funtions we dont have to hardcode the 311 (magic number?) .. 
we can use the bound on the range scale to get where the bottom x axis is (regardless of whether we are using a top/bottom axes). Just depends if we want bottom up transtiton even if the user has a top labelled axis chart

PR #485, opened 2020-02-06
feat(core): sanitize data

### LINKED ISSUE #334: [React] Donut chart doesn't honor the colors passed in the configuration
#### I am submitting a...

- [ ] Feature request
- [ ] Design defect
- [x] Source code defect
- [ ] Other

#### charts version:
0.15.8

### Issue description
The colors of the Donut chart are not as per the specifications provided by the input.

1. Define custom color array for a Donut chart
2. Pass the color array to the `DonutChart` component

#### Current behavior
The sections in the chart doesn't render in corresponding colors.

#### Expected behavior
The sections should render in the same order as passed by the configuration color

### Screenshot or recording
https://codesandbox.io/s/xlkpxqw59p

theiliad:
- parse non JS date objects within the model
- sort pie values and re-sort pie labels based on the order of values

Closes #334


packages/core/src/model.ts
```diff
@@ -38,6 +38,19 @@ export class ChartModel {
 		this.services = services;
 	}
 
+	santizie(data) {
```
cal-smith: Typo 🙂 
```suggestion
	sanitize(data) {
```

packages/core/src/components/graphs/pie.ts
```diff
@@ -78,7 +78,7 @@ export class Pie extends Component {
 		// Setup the pie layout
 		const pieLayout = pie()
 			.value((d: any) => d.value)
-			.sort((a: any, b: any) => b.value - a.value)
+			.sort(null)
```
cal-smith: If we don't need the sort, maybe we can just remove this line?

PR #489, opened 2020-02-10
Draft: update build and core packaging

cal-smith:
Lots of updates and general refinement

- core now builds a flatter and cleaner dist
- we manually symlink `dist` into `node_modules` since lerna doesn't work how we need it to
- better sharing of tsconfig
- core demo uses symlinked `@carbon/charts` instead of pulling from local `src` (simplifies build and makes it easier to use outside of core)
- react is UMD again! This shouldn't actually affect anything, but it gives consumers more options 
  - TODO investigate state of the art for react packaging and see if there's an obvious standard
  - TOOD (but not in this PR) switch react to typescript (not having the types available is a real pain)
- updated rollup global variable definitions - they used to all be `Charts`, now they're `Charts`/`ChartsReact`/`ChartsGeo` etc
- updated rollup externals and global var mapping


cal-smith:
This was supposed to be a draft, totally my bad 🤦‍♂ I even had the button expanded and everything

theiliad:
probably a good idea to add me to your fork so I can add the storybook bits for core

theiliad:
@cal-smith still issues with yarn. Try upgrading, there's a new 1.x version available

cal-smith:
@theiliad we should be good to go now 👍

theiliad:
@cal-smith 
image

cal-smith:
Weird! I'll take a look

cal-smith:
Should be good to go now!

Issue #402, opened 2019-11-06
[LineChart]: support y-axis basepoint starting above zero for large values


scottdickerson:
#### I am submitting a...

- [x] Feature request
- [ ] Design defect
- [ ] Source code defect
- [ ] Demo/documentation defect
- [ ] Other

#### charts version:
0.16.25

### Issue description
For graphs that have large y-values that hover around the same range, it would be nice to have a broken y-axis (or a y-axis that doesn't start from 0) so that we can see more variance in the data

Here's an example:
![image](https://user-images.githubusercontent.com/6663002/68342615-9c9cfe80-00b0-11ea-9d3e-a3256b93ca25.png)

theiliad:
Currently this is what happens:
![image](https://user-images.githubusercontent.com/14989804/69979147-b8eb5a00-14fb-11ea-9150-478a8f45b1ce.png)
(0 gets added to the scale domain)

we might need a `min` & `max` value in place in the options, defaulted to `0` and `null` respectively, that you can customize.

anne-stevens:
I support this request.  It is good practice for line charts to NOT always start with zero on the y-axis.  As Scott said, if the y-axis data range better suited the range in data, then it would make the chart more legible and understandable.

caesarsol:
In addition to #491 we'd like to add an option to include/exclude the `0`.
Our designers at Accurat (cc @gvnmagni) prefer to always show the 0 with a truncated axis, so in case of non-included 0 they would show the 0 with a truncation.

However, I would tackle the "Axes Truncation" problem in a separated issue, what do you think @theiliad?

theiliad:
> In addition to #491 we'd like to add an option to include/exclude the `0`.
> Our designers at Accurat (cc @gvnmagni) prefer to always show the 0 with a truncated axis, so in case of non-included 0 they would show the 0 with a truncation.
> 
> However, I would tackle the "Axes Truncation" problem in a separated issue, what do you think @theiliad?

Agree with all of that Caesar

ryshah:
+1 for allowing configuration option that allows axis to respect min/max ranges specified...

gvnmagni:
Hi! Generally speaking, we support the idea of allowing the y-axis to start a different values according to data.

We'd like to get some feedback on how to visually deal with it, we have few ideas:

1. Usual axis, instead of starting at zero it would start at the desired valued (top-left option). The most clean and elegant option of all

[...]

PR #491, opened 2020-02-12
optional domain setting

lucafalasco:
### Updates
Partially solves #402 by allowing to specify a fixed domain through each axis configuration.
The keyword `domain` has been added to `AxisOptions` interface. If you prefer, we can change it to something more inline with your naming conventions (maybe something like `extents` would be better?)

- [ ] Demos all features
- [ ] Documented/annotated
- [ ] Matches UI/UX specs
- [ ] Accessible

### DIFF
--- a/packages/core/src/interfaces/axis-scales.ts
+++ b/packages/core/src/interfaces/axis-scales.ts
@@ -1,4 +1,5 @@
 import { ScaleTypes } from "./enums";
+import { AxisDomain } from "d3";
 
 /**
  * options to configure a scale. not all options are used by all scales
@@ -22,6 +23,7 @@ export interface AxisOptions {
 	 * Range usually follows a linear scale
 	 */
 	useAsRange?: boolean;
+	domain?: AxisDomain[];
 	primary?: boolean;
 	secondary?: boolean;
 	/**
--- a/packages/core/src/interfaces/axis-scales.ts
+++ b/packages/core/src/interfaces/axis-scales.ts
@@ -23,6 +23,13 @@ export interface AxisOptions {
 	 * Range usually follows a linear scale
 	 */
 	useAsGraphRange?: boolean;
+	/**
+	 * Whether the Axis should use the specified domain
+	 * instead of it being dynamically generated based on data extents.
+	 * The type of values should depend on the scale type.
+	 * Example for continuous axis scale: [-100, 100]
+	 * Example for discrete axis scale: ['Qty', 'More', 'Sold']
+	 */
 	domain?: AxisDomain[];
 	primary?: boolean;
 	secondary?: boolean;
--- a/packages/core/src/interfaces/axis-scales.ts
+++ b/packages/core/src/interfaces/axis-scales.ts
@@ -15,14 +15,14 @@ export interface AxisOptions {
 	 * you would expect to only have 1 axis (dimension) being used as domain
 	 * Domain usually represents labels, ordinal values, time intervals etc.
 	 */
-	useAsDomain?: boolean;
+	useAsGraphDomain?: boolean;
 	/**
 	 * Whether the Axis should be used as the range
 	 * axis of the chart. In the case of Cartesian Scales
 	 * you would expect to only have 1 axis (dimension) being used as range
 	 * Range usually follows a linear scale
 	 */
-	useAsRange?: boolean;
+	useAsGraphRange?: boolean;
 	domain?: AxisDomain[];
 	primary?: boolean;
 	secondary?: boolean;
--- a/packages/core/demo/demo-data/bar.ts
+++ b/packages/core/demo/demo-data/bar.ts
@@ -106,6 +106,20 @@ export const simpleBarOptions = {
 	}
 };
 
+export const simpleBarFixedDomainOptions = {
+	title: "Simple bar (fixed domain)",
+	axes: {
+		left: {
+			primary: true,
+			domain: [-100000, 100000]
+		},
+		bottom: {
+			scaleType: "labels",
+			secondary: true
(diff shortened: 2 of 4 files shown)


packages/core/src/interfaces/axis-scales.ts
```diff
@@ -22,6 +23,7 @@ export interface AxisOptions {
 	 * Range usually follows a linear scale
 	 */
 	useAsRange?: boolean;
+	domain?: AxisDomain[];
```
theiliad: what's the definition of the `AxisDomain` type? Domains in d3 are usually an array of 2 values, is that reflected within that type?

caesarsol: This is the definition:
https://github.com/DefinitelyTyped/DefinitelyTyped/blob/84a9f0a15d5e765e20fef3a04b43824c9d3370e1/types/d3-axis/index.d.ts#L21

I think you are right, we could make it a tuple of 2 elements, it would be more specific.

This makes me think of a problem: there is a naming conflict, `domain` is something else in your code, here we mean a "Scale Domain". How would you call it?

lucafalasco: hey @theiliad,  here's the definition: https://github.com/DefinitelyTyped/DefinitelyTyped/blob/84a9f0a15d5e765e20fef3a04b43824c9d3370e1/types/d3-axis/index.d.ts#L21

Actually d3 domains are arrays of 2 or more values, as you can see here with the piecewise scale for example: https://github.com/d3/d3-scale#continuous_domain
`AxisDomain` type is just a helper type to define all the possibile values of the domain.

Open to discuss if you think it would be better to restrict the definition of the domain setting in this case though, as cesare was suggesting. But in that case I think it would be better to name it differently, to differentiate it from the "standard" d3 scale domain which may contain more than 2 values

theiliad: Right good catch 👍

theiliad:
@lucafalasco I'm good to merge once we get the rest of the reviews. Have you tested the functionality in different scenarios? Is everything ready to go?

lucafalasco:
> @lucafalasco I'm good to merge once we get the rest of the reviews. Have you tested the functionality in different scenarios? Is everything ready to go?

Well, at the moment it is not taken into account if `axisOptions.ScaleType === ScaleTypes.LABELS`, because you basically define the ordinal scale domain with the labels array in the `data` object supplied to the chart, so I'd say it does not work on discrete domains right now.

https://github.com/carbon-design-system/carbon-charts/pull/491/files#diff-44eb5956ed532e4793e3268bc07670a7R224-R231

theiliad:
Let's rename `useAsDomain` and `useAsRange` to `useAsGraphDomain` & `useAsGraphRange`

packages/core/src/interfaces/axis-scales.ts
```diff
@@ -14,14 +15,15 @@ export interface AxisOptions {
...
+	useAsGraphDomain?: boolean;
 	/**
 	 * Whether the Axis should be used as the range
 	 * axis of the chart. In the case of Cartesian Scales
 	 * you would expect to only have 1 axis (dimension) being used as range
 	 * Range usually follows a linear scale
 	 */
-	useAsRange?: boolean;
+	useAsGraphRange?: boolean;
+	domain?: AxisDomain[];
```
zvonimirfras: This is a useful addition, let's just add some documentation to show users how to use it.

Documentation should say what it's used for, what the possible values are and what the defaults are, if the value is not provided, along with some examples. It doesn't have to be longer than any other documentation strings.

cal-smith:
> Let's rename `useAsDomain` and `useAsRange` to `useAsGraphDomain` & `useAsGraphRange`

Do we really need the extra specificity here? Since it's already in the axis options wouldn't it be fairly clear that it's referring to the graph? 

Or are there other things the axis could be used as the domain or range of? In that case would something like `useAs: Graph.Domain` make sense?

caesarsol:
Hi @cal-smith, our concerns were in the ambiguity of the terminologies used:
 - in *carbon-charts* the "domain" is the X axis and the "range" is the Y axis – from now on "graph domain" and "graphRange"
 - in *d3-scale* a "domain" is a scale domain (so the span of output of a scale function) where "range" is the scale input – from now on "scale domain" and "scale range"

Let us know if you agree on this distinction.

However (cc @theiliad), we are actually wondering whether the two options `useAsDomain/Range` are actually used throughout the codebase: from a [text-based search](https://github.com/carbon-design-system/carbon-charts/search?q=useAsDomain&unscoped_q=useAsDomain) we found no other matches.

theiliad:
> Hi @cal-smith, our concerns were in the ambiguity of the terminologies used:
> 
> * in _carbon-charts_ the "domain" is the X axis and the "range" is the Y axis – from now on "graph domain" and "graphRange"
> * in _d3-scale_ a "domain" is a scale domain (so the span of output of a scale function) where "range" is the scale input – from now on "scale domain" and "scale range"
> 
> Let us know if you agree on this distinction.
> 
> However (cc @theiliad), we are actually wondering whether the two options `useAsDomain/Range` are actually used throughout the codebase: from a [text-based search](https://github.com/carbon-design-system/carbon-charts/search?q=useAsDomain&unscoped_q=useAsDomain) we found no other matches.

They were supposed to be used in the horizontal bar implementation, I missed using them and they seem to have slept through the PR. We can actually remove those for now until I drop in the use case for those

theiliad:
@lucafalasco Could we please add a couple of storybook demos for this one? I think once we do that we can merge it in, thanks!

PR #492, opened 2020-02-12
feat(core): Show a different number of vertical axes ticks depending …

### LINKED ISSUE #483: Option to define the ticks range, max number of ticks and distribution on Y -axis on carbon SimpleBarChart
#### I am submitting a...
 Feature request:

Do we have any option to define the ticks range, max number of ticks and distribution on Y -axis on carbon SimpleBarChart?

#### charts version:

    "@carbon/charts": "^0.24.3",
    "@carbon/charts-react": "^0.24.3"

### Issue description

Unable to set ticks distribution which allows the charts to auto decide and hence it looks squashed.
image

#### Expected behavior

Option to define the distribution on Y -axis on carbon SimpleBarChart as we have in d3.

SerenaG:
### Updates
Closes #483.
Enhanced vertical axis to calculate how many ticks can be shown depending on the chart height. 
If `Tools.getProperty(axisOptions, "ticks", "number")` is defined, it will be used instead.

Strategy for calculating height:
- Add text element in axis (take the correct font size)
- Get height (tickHeight)
- Remove it 
- Calculate number of ticks as: chartHeight / (tickHeight * 2.5) 
2.5 was chosen with Accurat designers as a font size coefficient to make labels look not-crowded.

### Recording
ticksnum

- [ ] Demos all features
- [ ] Documented/annotated
- [ ] Matches UI/UX specs
- [ ] Accessible

### DIFF
--- a/packages/core/src/components/axes/axis.ts
+++ b/packages/core/src/components/axes/axis.ts
@@ -105,13 +105,10 @@ export class Axis extends Component {
 			if (isNumberOfTicksProvided) {
 				numberOfTicks = numberOfTicksProvided;
 			} else {
-				// If number of ticks is not provided:
+				numberOfTicks = Configuration.axis.ticks.number;
 				if (isVerticalAxis) {
-					// Get how many ticks based on height
+					// Set how many ticks based on height
 					numberOfTicks = this.getNumberOfFittingTicks(height, tickHeight, Configuration.tickSpaceRatioVertical);
-				} else {
-					// Set default ticks number
-					numberOfTicks = Configuration.axis.ticks.number;
 				}
 			}
 
--- a/packages/core/src/components/axes/axis.ts
+++ b/packages/core/src/components/axes/axis.ts
@@ -31,6 +31,9 @@ export class Axis extends Component {
 		const { position: axisPosition } = this.configs;
 		const options = this.model.getOptions();
 		const axisOptions = Tools.getProperty(options, "axes", axisPosition);
+		const numberOfTicksProvided = Tools.getProperty(axisOptions, "ticks", "number");
+		const isNumberOfTicksProvided = numberOfTicksProvided !== null;
+		const isVerticalAxis = axisPosition === AxisPositions.LEFT || axisPosition === AxisPositions.RIGHT;
 
 		const svg = this.getContainerSVG();
 		const { width, height } = DOMUtils.getSVGElementSize(this.parent, { useAttrs: true });
@@ -70,13 +73,48 @@ export class Axis extends Component {
 				break;
 		}
 
+		// Add axis into the parent
+		const container = DOMUtils.appendOrSelect(svg, `g.axis.${axisPosition}`);
+		const axisRefExists = !container.select(`g.ticks`).empty();
+		let axisRef = DOMUtils.appendOrSelect(container, `g.ticks`);
+		if (!axisRefExists) {
+			axisRef.attr("role", `${Roles.GRAPHICS_OBJECT} ${Roles.GROUP}`);
+		}
+
+		// We draw the invisible axis because of the async nature of d3 transitions
+		// To be able to tell the final width & height of the axis when initiaing the transition
+		// The invisible axis is updated instantly and without a transition
+		const invisibleAxisRef = DOMUtils.appendOrSelect(container, `g.ticks.invisible`)
+			.style("opacity", "0")
+			.attr("aria-hidden", true);
+
+		// Append to DOM a fake tick to get the right computed font height
+		const fakeTick = DOMUtils.appendOrSelect(invisibleAxisRef, `g.tick`);
+		const fakeTickText = DOMUtils.appendOrSelect(fakeTick, `text`).text("0");
+		const tickHeight = DOMUtils.getSVGElementSize(fakeTickText.node(), { useBBox: true }).height;
+		fakeTick.remove();
+
 		// Initialize axis object
 		const axis = axisFunction(scale)
 			.tickSizeOuter(0)
 			.tickFormat(Tools.getProperty(axisOptions, "ticks", "formatter"));
 
 		if (scale.ticks) {
-			const numberOfTicks = Tools.getProperty(axisOptions, "ticks", "number") || Configuration.axis.ticks.number;
+			let numberOfTicks;
+
+			if (isNumberOfTicksProvided) {
+				numberOfTicks = numberOfTicksProvided;
+			} else {
+				// If number of ticks is not provided:
+				if (isVerticalAxis) {
+					// Get how many ticks based on height
+					numberOfTicks = this.getNumberOfFittingTicks(height, tickHeight, Configuration.tickSpaceRatioVertical);
+				} else {
+					// Set default ticks number
+					numberOfTicks = Configuration.axis.ticks.number;
+				}
+			}
+
 			axis.ticks(numberOfTicks);
 
 			if (this.scaleType === ScaleTypes.TIME) {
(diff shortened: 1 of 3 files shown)


packages/core/src/components/axes/axis.ts
```diff
@@ -70,13 +70,44 @@ export class Axis extends Component {
...
+		// To be able to tell the final width & height of the axis when initiaing the transition
+		// The invisible axis is updated instantly and without a transition
+		const invisibleAxisRef = DOMUtils.appendOrSelect(container, `g.ticks.invisible`)
+			.style("opacity", "0")
+			.attr("aria-hidden", true);
+
+		// Get font height
+		const fakeTick = DOMUtils.appendOrSelect(invisibleAxisRef, `g.tick`);
+		const fakeTickText = DOMUtils.appendOrSelect(fakeTick, `text`).text("0");
+		const tickHeight = fakeTickText.node().getBBox().height;
```
theiliad: `getBBox()` doesn't always work in some browser versions. Use `DOMUtils.getSVGElementSize()` please, search for usage examples in the code

packages/core/src/components/axes/axis.ts
```diff
@@ -70,13 +70,44 @@ export class Axis extends Component {
...
+		}
+
+		// We draw the invisible axis because of the async nature of d3 transitions
+		// To be able to tell the final width & height of the axis when initiaing the transition
+		// The invisible axis is updated instantly and without a transition
+		const invisibleAxisRef = DOMUtils.appendOrSelect(container, `g.ticks.invisible`)
+			.style("opacity", "0")
+			.attr("aria-hidden", true);
+
+		// Get font height
```
theiliad: Maybe a bit more comments here to explain why we're rendering . the fakeTick

packages/core/src/components/axes/axis.ts
```diff
@@ -70,13 +70,44 @@ export class Axis extends Component {
...
+		fakeTick.remove();
+
 		// Initialize axis object
 		const axis = axisFunction(scale)
 			.tickSizeOuter(0)
 			.tickFormat(Tools.getProperty(axisOptions, "ticks", "formatter"));
 
 		if (scale.ticks) {
-			const numberOfTicks = Tools.getProperty(axisOptions, "ticks", "number") || Configuration.axis.ticks.number;
+			let numberOfTicks = Tools.getProperty(axisOptions, "ticks", "number");
```
theiliad: You'd still need `|| Configuration.axis.ticks.number`. User-provided options normally are merged with our default configurations, except for `axes` options (because of the limitations that'd apply there) so we'd still need to fallback on a couple of things

packages/core/src/components/axes/axis.ts
```diff
@@ -70,13 +70,44 @@ export class Axis extends Component {
...
+
+			if (axisPosition === AxisPositions.LEFT || axisPosition === AxisPositions.RIGHT) {
+				if (!numberOfTicks) {
+					const numberOfTicksThoretical = Math.floor(height / (tickHeight * 2.5));
+					numberOfTicks = Tools.clamp(numberOfTicksThoretical, 2, Configuration.axis.ticks.number);
+				}
+			} else {
+				numberOfTicks = numberOfTicks || Configuration.axis.ticks.number;
+			}
```
theiliad: What if I move the y-axis on the left to the bottom? (e.g. horizontal bar)

We should really have a way of identifying the nature of the scale domain and based on that decide whether we should remove some ticks

packages/core/src/components/axes/axis.ts
caesarsol: This was implemented in this way to follow this logic:
 - if the user does specify an `options.[axis].ticks.number`, that tickNumber is used (ignoring the chart height)
 - if the user does NOT specify the option, the `Configuration.axis.ticks.number` is used as a default but it's adapted on the chart height.

The fallback to `|| Configuration.axis.ticks.number` is done in line 108.

Are you suggesting a different logic? 
We could make it more obvious, maybe a boolean `adaptTicksNumberBasedOnHeight`, but we feel it would be less easy to use.

packages/core/src/components/axes/axis.ts
caesarsol: 👍

packages/core/src/components/axes/axis.ts
caesarsol: 👍

packages/core/src/components/axes/axis.ts
SerenaG: What we are trying to fix with this PR is to adjust the number of ticks on vertical axises based on the chart height.

We are not interested on whether it's the Y or X axis, we only care if it is a vertical one.
Do you think it's the wrong logic?

We can extend this solution to horizontal axes in the future too.

packages/core/src/components/axes/axis.ts
caesarsol: @theiliad any feedback on this point? Just to understand if I misunderstood

packages/core/src/tools.ts
```diff
@@ -170,7 +172,7 @@ export namespace Tools {
 	 * @returns The function to be used by D3 to push element to the top of the canvas
 	 */
 	export function moveToFront(element) {
-		return element.each(function() {
+		return element.each(function () {
```
theiliad: extra space

packages/core/src/components/axes/axis.ts
```diff
@@ -188,6 +206,13 @@ export class Axis extends Component {
 			}
 
 			if (rotateTicks) {
+				if (numberOfTicksProvided === null) {
+					axis.ticks(this.getNumberOfFittingTicks(width, tickHeight, 3.5));
```
theiliad: What's `3.5`?

theiliad:
Demos are awesome 💯

packages/core/src/components/axes/axis.ts
SerenaG: It is the space ratio we want to keep between ticks, used to calculate how many ticks will fit the axis `size / (tickSize * spaceRatio)`.

theiliad: Seems like we should move it to `configuration.ts`

theiliad: Seems like we should move it to `configuration.ts`

packages/core/src/components/axes/axis.ts
```diff
@@ -70,13 +73,48 @@ export class Axis extends Component {
...
 			.tickSizeOuter(0)
 			.tickFormat(Tools.getProperty(axisOptions, "ticks", "formatter"));
 
 		if (scale.ticks) {
-			const numberOfTicks = Tools.getProperty(axisOptions, "ticks", "number") || Configuration.axis.ticks.number;
+			let numberOfTicks;
+
+			if (isNumberOfTicksProvided) {
+				numberOfTicks = numberOfTicksProvided;
+			} else {
```
zvonimirfras: A bit of nitpicking:

It looks like this can be either that number of ticks is provided, or that it's a vertical axis, or that it's the default number of ticks. That would make this block look like

```javascript
if (isNumberOfTicksProvided) {
	// ...
} else if (isVerticalAxis) {
	// ...
} else {
	// ...
}
```

or that we set the default (since it's just a simple assignment) and then conditionally overwrite like

```javascript
numberOfTicks = Configuration.axis.ticks.number;
if (isNumberOfTicksProvided) {
	// ...
} else if (isVerticalAxis) {
	// ...
}
```

or to merge the ideas

```javascript
if (isNumberOfTicksProvided) {
	// ...
} else {
	numberOfTicks = Configuration.axis.ticks.number;
	if (isVerticalAxis) {
		// ...
	}
}
```

SerenaG: Hi @zvonimirfras, 
I've refactored the code as the 3rd idea! The if statement should be more readable now. 
Thank you.

PR #499, opened 2020-02-17
Ruler interaction replacing Grid mouse listeners for comparative tooltips 

### LINKED ISSUE #397: [LineChart]: in the new scaleType: time, I can't get multiline tooltips to show up for my data, it appears from the demo data that they only show up if the x exactly matches one of the gridlines
#### I am submitting a...

- [x] Feature request
- [ ] Design defect
- [ ] Source code defect
- [ ] Other

#### charts version:
0.16.25

### Issue description
You can see this limitation in the demo data page, by hovering over a non-gridlined datapoint
image

Versus a gridlined datapoint:
image

### LINKED ISSUE #478: Axis ticks placement / tooltip issues
This issue is highly visible in the **area** chart demos:
https://deploy-preview-412--carbon-charts-core.netlify.com

However it also exists in **any** chart that supports continuous scales.

**Description:**
image

**Possible solutions:**
- Refactor of axis ticks to line up more closely with datapoints presented in the graphs
- Sticky tooltips (could introduce its own set of disadvantages as well, still may be a possibility for users to enable if they believe that there's a continuous relationship between their datapoints...?)
- Clustering of datapoints within a tooltip (e.g. divide grid into regions that can activate tooltips on all the points within the region + possible extra styling to highlight those points on hover within the region...?)

Looking for options here

lucafalasco:
- [x] add ruler functionality
- [x] refactor and add different but composable tooltip modes
- [x] color datapoints with tooltip activated
- [x] get dynamic data point accessor 
- [x] basic snap 
- [x] snap to nearest data point 
- [x] work with discrete domains
- [x] test
- [x] change ruler color
- [x] disable the ruler for every chart that does not have a continuous domain (but keep snap)
- [x] adapt ruler behaviour when continuous axis is positioned on top
- [x] axis tooltip should not go out of the svg bounding box
- [x] only enable for timescale linechart
- [x] show ruler only when it's matching some points, hide it otherwise
- [x] remove axis tooltip
- [x] disable interaction with grid
- [x] enable ruler for all charts with data points (barcharts and piecharts excluded)
 
### Updates
close #478  
close #397 

ruler

### Deploy
https://deploy-preview-499--carbon-charts-core.netlify.app/?path=/story/line--line-dense-time-series


packages/core/src/components/axes/ruler.ts
```diff
@@ -30,8 +30,10 @@ export class Ruler extends Component {
 		const ruler = DOMUtils.appendOrSelect(svg, "g.ruler").attr("opacity", 1);
 		const line = DOMUtils.appendOrSelect(ruler, "line.ruler-line");
 		const dataPoints = svg.selectAll("[role=graphics-symbol]");
+		const height = Number(this.backdrop.attr("height"));
```
theiliad: We should use `DOMUtils.getSVGElementSize()` if this is something that might change since some browsers will mess up that calculation

theiliad:
The `snap to grid` threshold seems a bit small to me, is that finalized?

lucafalasco:
> The `snap to grid` threshold seems a bit small to me, is that finalized?

Forwarded question to designers 😉 
I personally think it's better to keep it small (in case of really close data points), but we can maybe tweak it a little bit

gvnmagni:
> > The `snap to grid` threshold seems a bit small to me, is that finalized?
> 
> Forwarded question to designers 😉
> I personally think it's better to keep it small (in case of really close data points), but we can maybe tweak it a little bit

I'd support Luca's idea, since this kind of tooltip will be helpful especially when we'll have unusual data distributions, it would be better to keep this as small as possible. 

Given this, we are totally open to make it few pixels wider if it works better, since we can change it at any time I'd say to try and to see what people think about it and, if necessary, adapt it eventually

packages/core/package.json
```diff
@@ -51,7 +51,8 @@
 		"babel-polyfill": "6.26.0",
 		"date-fns": "2.8.1",
 		"lodash-es": "4.17.15",
-		"resize-observer-polyfill": "1.5.0"
+		"resize-observer-polyfill": "1.5.0",
+		"text-width": "^1.2.0"
```
theiliad: We should get rid of the `^`

packages/core/src/components/axes/ruler.ts
```diff
@@ -0,0 +1,206 @@
+// Internal Imports
+import { Component } from "../component";
+import { DOMUtils } from "../../services";
+import textWidth from "text-width";
+import { scaleLinear } from "d3-scale";
+import { Selection, BaseType } from "d3-selection";
+import { isEqual } from "lodash-es";
```
theiliad: 1. The 3 imports above aren't internal so they should be part of the other block
2. `d3-selection` was already imported below, we should add those modules to the import statement below

lucafalasco:
> Haven't reviewed all of the code yet. I see lots of readability issues so far, but the demos seem to be functioning well.

@theiliad what do you mean with "readability issues"? are you talking about specific code readability (in that case what exactly did you find unreadable?) or are you referring to the ruler/tooltips functionality which you consider not clear enough?

cal-smith:
image

The box around the x-axis values doesn't look finished ... it seems like it might need some more padding ... is this the final design or still WIP?

Also, I haven't looked into the code yet, but (how) does this handle double x-axis, and only top x-axis?

lucafalasco:
> image
> 
> The box around the x-axis values doesn't look finished ... it seems like it might need some more padding ... is this the final design or still WIP?
> 
> Also, I haven't looked into the code yet, but (how) does this handle double x-axis, and only top x-axis?

thank you @cal-smith, apparently that was just an issue with firefox not handling a specific SVG attribute for baseline, should be ok now, and yes, the design is final.

Regarding different axis behaviours, this is the only one supported from the design at the moment.

lucafalasco:
> A few quick possible improvements. I'll have to do a deeper read later.

@zvonimirfras thanks, pushed a commit with some refactor before, maybe you had an older version of the code, sorry about that

cal-smith:
>thank you @cal-smith, apparently that was just an issue with firefox not handling a specific SVG attribute for baseline, should be ok now, and yes, the design is final.

Cool, just checked it out and that definitely looks better :+1:

theiliad:
@gvnmagni Couple of questions here:

1. Shouldn't I be seeing either a ruler here or see the tooltip if I get closer to the stacks?

image

In the case of a discrete line chart we show the ruler experience even though there isn't a continuous domain, shouldn't we offer the same experience with bars?

image

We do the same for discrete scatter

image


2. What if the time-series axis is on the left, top or right?

image

3. How do we deal with tooltips sticking out of the chart container?

image

4. Shouldn't we see a horizontal ruler here?

image

lucafalasco:
Here is some feedback from the design team:
1. it would be better to disable the ruler for every chart that does not have a continuous domain. 
At the moment it's like that because the construction of the linechart and scatterplot components is the same both for continuous and discrete domains. We'll find the best way to disable it given the current implementation.
2. ok, we'll work on adapting the ruler behaviour when the continuous axis is positioned on top of the chart. 
3. ok, we'll make sure it does not go out of the svg bounding box, design-wise it's ok if sticks out at the end of axis. 
4. we decided not to show the ruler for barcharts.

lucafalasco:
also @theiliad we’re still waiting for feedback regarding readability issues you mentioned, curious to know which parts you found difficult to understand, thank you!

lucafalasco:
> * Bubbles aren't highlighting when the ruler snaps to them

yes that's because of the way we're selecting the graphic elements to highlight here:

https://github.com/carbon-design-system/carbon-charts/blob/939534e42a3061acdcba4978465d3a02dd2a98b9/packages/core/src/components/axes/ruler.ts#L52

It seems that the bubbles are the only elements without that attribute at the moment, was that on purpose?

theiliad:
> It seems that the bubbles are the only elements without that attribute at the moment, was that on purpose?

No! Seems like we missed that one. Let's add it in

theiliad:
Also, are the colors here readable enough? @gvnmagni 
image

gvnmagni:
Russel, this is the image you needed

75669871-0f437280-5c7c-11ea-8b04-8bcb0e830b7a

caesarsol:
In light of the new discussions on the single and comparative tooltips (#550) we decided on making this PR smaller.

We will then:
- remove the "black box" of the Axis Tooltip
- show the Ruler only when it's matching some points, hide it otherwise
- enable the Ruler logics only for Linechrts with a Timescale
- keep the snapping logics
- keep the tooltips as they are, will be changed in a different PR

With this implementation, the original issues (#478, #397) can be closed and we will focus on a new issue on the tooltip content itself.

packages/core/src/components/axes/ruler.ts
```diff
@@ -0,0 +1,187 @@
...
+			"[role=graphics-symbol]"
+		);
+		const { datasets } = this.model.getDisplayData();
+		const mainXScale = this.services.cartesianScales.getMainXScale();
+		const mainYScale = this.services.cartesianScales.getMainYScale();
+		const [yScaleEnd, yScaleStart] = mainYScale.range();
+
+		const scaledData: number[] = flatten(datasets.map(({ data }) =>
+			data.map((d, i) => Number(this.services.cartesianScales.getDomainValue(d, i)))
+		));
```
theiliad, line 55: again this won't be needed since we'll have to make the switch to tabular data

packages/core/src/components/axes/ruler.ts
```diff
@@ -0,0 +1,187 @@
...
+
+		/**
+		 * Find matches, reduce is used instead of filter
+		 * to only get elements which belong to the same axis coordinate
+		 */
+		const scaledValuesMatches: number[] = scaledData.reduce((acc, cur) => {
```
theiliad, line 61:
```suggestion
		const scaledValuesMatches: number[] = scaledData.reduce((accum, currentValue) => {
```

packages/core/src/components/axes/ruler.ts
```diff
@@ -0,0 +1,187 @@
...
+			const sampleAccValue = acc[0];
```
theiliad, line 62: it's a bit unclear to me what the difference between `acc` & `sampleAccValue` is, and why we need to use it. Maybe a more readable variable name + some comments?

packages/core/src/components/axes/ruler.ts
```diff
@@ -0,0 +1,187 @@
...
+
+			// if current value is bigger than already existing values forget it
```
theiliad, line 64:
```suggestion
			// if current value is bigger than already existing values, don't render the ruler
```

packages/core/src/components/axes/ruler.ts
```diff
@@ -0,0 +1,187 @@
...
+					// cur is equal to already stored values, there's another match on the same coordinate
+					acc.push(cur);
+				}
+			}
+
+			return acc;
+		}, []);
+
+		// some data point match
+		if (scaledValuesMatches.length > 0) {
```
theiliad, line 84: what does `scaledValuesMatches` do exactly? is there a better var name we can use?

packages/core/src/components/axes/ruler.ts
```diff
@@ -0,0 +1,187 @@
...
+			/** if we pass from a trigger area to another one
+			 * mouseout on previous elements won't get dispatched
+			 * so we need to do it manually
+			 */
+			if (
+				this.hoveredElements &&
+				this.hoveredElements.size() > 0 &&
+				!isEqual(this.hoveredElements, hoveredElements)
+			) {
+				this.hoveredElements.dispatch("mouseout");
```
theiliad, line 106: can we call `this.hideRuler()` here?

packages/core/src/components/axes/ruler.ts
```diff
@@ -0,0 +1,187 @@
...
+		const domainAxisPosition = this.services.cartesianScales.getDomainAxisPosition();
+		const domainScaleType = this.services.cartesianScales.getScaleTypeByPosition(
+			domainAxisPosition
+		);
+		const isTimeSeries = domainScaleType === ScaleTypes.TIME;
+
+		// if scale type is not timeSeries do not show rule
+		if (!isTimeSeries) {
+			return;
+		}
```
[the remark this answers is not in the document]
lucafalasco: We decided to only enable the ruler for time scales at the moment

packages/core/src/components/axes/ruler.ts
```diff
@@ -0,0 +1,187 @@
...
+
+				self.showRuler(pos);
+			})
+			.on("mouseout", function() {
+				self.hideRuler();
+				self.services.events.dispatchEvent("hide-tooltip", {});
+			});
+	}
+
+	drawBackdrop() {
```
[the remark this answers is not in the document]
lucafalasco: It is needed to attach mouse listeners

packages/core/src/components/axes/ruler.ts
lucafalasco: It's an array containing all the data points that match the current mouse x position, which are basically the ones that need to be triggered. I called it like that to be "extensible" (when we started working on this we planned to have the same ruler interaction on other charts with different graphic elements too) but if you have a better name we can change it, no prob 👍

packages/core/src/components/axes/ruler.ts
lucafalasco: `acc` here is the accumulator array, basically the "result" of the reducer, `sampleAccValue` is instead a "sample" value taken from the accumulator array, it is used to do the various comparisons with the current value being processed. 
I've added some more comments, hope it's clearer now

lucafalasco: The reason behind the need of `sampleAccValue` is that we need to do an additional filter to only store data points which belong to the exact same axis coordinate, not just those that make it though the `pointIsWithinThreshold()` condition

theiliad:
should we still show the ruler here?
image

gvnmagni:
> should we still show the ruler here?
> image

I'd say yes, I like the idea of showing the ruler since we are on a tick (consistent behaviour) but showing no tooltips since we have no datapoints. What do you think Eliad?

lucafalasco:
Yep, that's actually the interaction with the grid component, so not related to ruler component.
It's the same one you currently get on other charts, event if there's no data to show, here for example:
Screenshot 2020-04-17 at 10 46 29

theiliad:
> Yep, that's actually the interaction with the grid component, so not related to ruler component.
> It's the same one you currently get on other charts, event if there's no data to show, here for example:
> Screenshot 2020-04-17 at 10 46 29

yes we show that to enable tooltips on the grid, we should just stop showing if there are no values lying within that line

packages/core/src/components/axes/ruler.ts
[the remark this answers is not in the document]
theiliad: what's stopping us from enabling it for all linear scales?

packages/core/src/components/axes/ruler.ts
```diff
@@ -0,0 +1,187 @@
...
+			const sampleAccumValue = accum[0];
+
+			// if accumulator is not empty and current value is bigger than already existing value in the accumulator, skip current iteration
+			if (sampleAccumValue && currentValue > sampleAccumValue) {
+				return accum;
+			}
+
+			// there's a match and currentValue is either less then or equal to already stored values
+			if (pointIsWithinThreshold(currentValue, x)) {
+				if (sampleAccumValue && currentValue < sampleAccumValue) {
```
theiliad, line 71: wouldn't this break if `sampleAccumValue` is 0?

packages/core/src/components/axes/ruler.ts
theiliad: I think the name `dataPointsMatchingRulerLine` should be good, what do you think?

packages/core/src/components/axes/ruler.ts
```diff
@@ -0,0 +1,187 @@
...
+
+			const highlightItems = this.services.cartesianScales.getDataFromDomain(
+				domainScale.invert(sampleMatch)
+			).filter(d => d.value);
+
+			const hoveredElements = dataPoints.filter((d, i) =>
+				scaledValuesMatches.includes(
+					Number(this.services.cartesianScales.getDomainValue(d))
+				)
+			);
```
theiliad, line 95: can we please add some comments here describing what's happening?

packages/core/src/components/axes/ruler.ts
```diff
@@ -0,0 +1,187 @@
...
+
+		this.drawBackdrop();
+		this.addBackdropEventListeners();
+	}
+
+	showRuler([x, y]: [number, number]) {
+		const svg = this.parent;
+		const ruler = DOMUtils.appendOrSelect(svg, "g.ruler");
+		const line = DOMUtils.appendOrSelect(ruler, "line.ruler-line");
+		const dataPoints: GenericSvgSelection = svg.selectAll(
```
theiliad, line 44:
```suggestion
		const dataPointElements: GenericSvgSelection = svg.selectAll(
```

packages/core/src/components/axes/ruler.ts
```diff
@@ -0,0 +1,187 @@
...
+		this.backdrop
+			.on("mousemove mouseover", function() {
+				const chartContainer = self.services.domUtils.getMainSVG();
+				const pos = mouse(chartContainer);
+
+				self.showRuler(pos);
+			})
+			.on("mouseout", function() {
+				self.hideRuler();
+				self.services.events.dispatchEvent("hide-tooltip", {});
```
theiliad, line 157:
```suggestion
				self.services.events.dispatchEvent("hide-tooltip");
```

packages/core/src/components/axes/ruler.ts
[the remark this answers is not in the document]
theiliad: that means we can get rid of the `grid.ts` event listeners for tooltip...?

packages/core/src/configuration.ts
```diff
@@ -79,7 +79,7 @@ export const baseTooltip: TooltipOptions = {
 export const axisChartTooltip: AxisTooltipOptions = Tools.merge({}, baseTooltip, {
 	gridline: {
 		enabled: true,
-		threshold: 0.25
+		threshold: 0.02
```
theiliad, line 82: the current threshold I see on any datapoints seems to be too small. can we increase that please?

packages/core/src/styles/components/_ruler.scss
```diff
@@ -0,0 +1,20 @@
+.#{$prefix}--#{$charts-prefix}--ruler {
+	line.ruler-line {
+		@if $ui-background == map-get($carbon--theme--g90, 'ui-background') {
+			stroke: #fff;
+		} @else if $ui-background == map-get($carbon--theme--g100, 'ui-background') {
+			stroke: #fff;
+		} @else {
+			stroke: #000;
+		}
```
theiliad, line 9:
```suggestion
		}

```

packages/core/src/tools.ts
```diff
@@ -11,6 +11,8 @@ import {
 	cloneDeep as lodashCloneDeep,
 	uniq as lodashUnique,
 	clamp as lodashClamp,
+	isEqual as lodashIsEqual,
+	flatten as lodashFlaten,
```
theiliad, line 15: is `flatten` being exported?

theiliad: seems like we don't use it anymore?

packages/core/src/components/axes/ruler.ts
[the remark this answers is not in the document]
lucafalasco: The idea we suggested when we talked about that was to only expect time scales to be used on LineCharts as it's the only one that makes sense with that type of chart

packages/core/src/configuration.ts
lucafalasco: Do you mean the ruler threshold or the gridline threshold? The reason we made the ruler threshold value so small is to prevent overlapping trigger events on datapoints which are really close to each other.

packages/core/src/components/axes/ruler.ts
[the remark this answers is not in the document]
lucafalasco: we discussed this and while we don't think it would be good to remove listeners from `grid.ts` as it would affect other charts too (where ruler does not exist), what we can do is add the possibility to pass a boolean option to the `Grid` class constructor to disable the event listeners for specific cases (right now only when ruler is active, which means only on the linechart)

designertyler:
I think I agree with @theiliad that the ruler shouldn't appear unless there's a value
> yes we show that to enable tooltips on the grid, we should just stop showing if there are no values lying within that line.
> 

One thing I noticed on hover is there's a double ruler when there are multiple values on a tick.

on white and g10 the second one is blue and the dark themes they're both white.

image

packages/core/demo/data/line.ts
```diff
@@ -77,6 +77,69 @@ export const lineTimeSeriesOptions = {
...
+	{ group: "Dataset 2", date: new Date(2019, 0, 2), value: 35000 },
+	{ group: "Dataset 2", date: new Date(2019, 0, 2, 7), value: 36000 },
+	{ group: "Dataset 2", date: new Date(2019, 0, 3), value: 20000 },
+	{ group: "Dataset 2", date: new Date(2019, 0, 3, 6), value: 20000 },
+	{ group: "Dataset 2", date: new Date(2019, 0, 3, 18), value: 22000 },
+	{ group: "Dataset 2", date: new Date(2019, 0, 5), value: 62000 },
+	{ group: "Dataset 2", date: new Date(2019, 0, 6), value: 52000 },
+	{ group: "Dataset 2", date: new Date(2019, 0, 7), value: 52000 },
+	{ group: "Dataset 2", date: new Date(2019, 0, 7, 15), value: 52000 },
+	{ group: "Dataset 2", date: new Date(2019, 0, 7), value: 52000 },
```
natashadecoste, line 113:
```suggestion

```
accidental duplication? Do we need this? I think the demo does already show how overlapping datapoints are handled between the Datasets... would it even make sense for a dataset to have duplicate data key/value pairs? We can't control users' data and duplicates don't impact the graphic at all but the the tooltip will show all the duplicate points on hover 🤔

packages/core/src/components/axes/ruler.ts
```diff
@@ -0,0 +1,188 @@
...
+			) {
+				this.hideRuler();
+			}
+
+			hoveredElements.dispatch("mouseover");
+
+			// set current hovered elements
+			this.hoveredElements = hoveredElements;
+
+			this.services.events.dispatchEvent("show-tooltip", {
```
natashadecoste, line 115: nice 👍

packages/core/src/components/axes/ruler.ts
```diff
@@ -0,0 +1,188 @@
...
+		} else {
+			ruler.attr("opacity", 0);
+			dataPointElements.dispatch("mouseout");
+		}
+	}
+
+	hideRuler() {
+		const svg = this.parent;
+		const ruler = DOMUtils.appendOrSelect(svg, "g.ruler");
+		const dataPointElements = svg.selectAll("[role=graphics-symbol]");
```
natashadecoste, line 137: why use role to select instead of something like  `svg.selectAll("circle.dot");` ? 
if it also works, it seems more conventional in this codebase

packages/core/src/styles/components/_ruler.scss
```diff
@@ -0,0 +1,21 @@
...
+
+		stroke-width: 1px;
+		stroke-dasharray: 2;
+		pointer-events: none;
+	}
+
+	text.axis-tooltip-text {
+		fill: white;
```
natashadecoste, line 17:
```suggestion
		fill: $carbon--white-0;
```

packages/core/src/styles/components/_ruler.scss
```diff
@@ -0,0 +1,21 @@
+.#{$prefix}--#{$charts-prefix}--ruler {
+	line.ruler-line {
+		@if $ui-background == map-get($carbon--theme--g90, 'ui-background') {
+			stroke: #fff;
```
natashadecoste, line 4:
```suggestion
			stroke: $carbon--white-0;
```

packages/core/src/styles/components/_ruler.scss
```diff
@@ -0,0 +1,21 @@
...
+		} @else if $ui-background == map-get($carbon--theme--g100, 'ui-background') {
+			stroke: #fff;
```
natashadecoste, line 6:
```suggestion
			stroke: $carbon--white-0;
```

packages/core/src/styles/components/_ruler.scss
```diff
@@ -0,0 +1,21 @@
...
+		} @else {
+			stroke: #000;
```
natashadecoste, line 8:
```suggestion
			stroke: $carbon--black-100;
```

packages/core/src/components/axes/ruler.ts
lucafalasco: The use of that selector seemed to me more appropriate since we initially thought to enable the ruler on very different charts, which might have different graphic elements that need to be highlighted (for example, if I remember correctly there was an idea to have other shapes for the scatterplot other than circles - squares, triangles etc) that way we could have a single selector which includes all different svg elements. 
But you're right, at the moment the `circle.dot` selector should work across all current scenarios 👍

packages/core/src/components/axes/ruler.ts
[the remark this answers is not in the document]
lucafalasco: just to recap after latest discussion, I'm now going to disable mouse listeners on grid and enable ruler interaction for every other chart and scale type 😅

lucafalasco:
thanks @theiliad, that was an issue with linear scales, should be fixed now 👍

PR #504, opened 2020-02-19
fix(build): fix bundle output for core

cal-smith:
fixes:

- package.json points to the correct module and main
- rollup no longer rewrites `this` to `undefined` (potentially causing issues)
- should fix codesandbox, since it uses bundles by default, and it's module transpiler doesn't seem to compile our enums correctly
- bundle now includes the en-US.json

changes:
- simplifies core `index.ts` (we can just `export * from "...";` in most places)


cal-smith:
I did a ton of digging as to why it was working locally but not on codesandbox, and through that I found that (with no other changes) it would load the bundle.js for 0.29.1 ... this lead me down a debugging rabbit hole where I found that codesandbox applies a _very_ custom traspliation process to all external modules (when it can't find a bundle) - my current theory is that something in that process broke the references between the `var MyEnum;` declarations, the `exports.MyEnum = MyEnum` and the IIFE that populates the variables ... in the debugger you could clearly see that all the local vars were populated, but the `exports.MyEnum` variables would always remain `undefined` ... of course, this all works locally since we're using standard toolchains that aren't inserting extra calls in between the variable exports, and aren't recompiling the entirety of `node_modules`.

cal-smith:
This is a draft for now, just so we can catch any glaring issues before we're confident this covers everything.

packages/core/src/index.ts
```diff
@@ -1,33 +1,11 @@
...
-import * as interfaces from "./interfaces";
+export * from "./configuration";
+export * from "./interfaces";
 
-import * as colorPalettes from "./services/colorPalettes";
+export * from "./services/colorPalettes";
+import { DEFAULT } from "./services/colorPalettes";
 // TODO 1.0 - Remove deprecated API
-const defaultColors = colorPalettes.DEFAULT;
+export const defaultColors = DEFAULT;
```
theiliad: could contain some major breaking changes

cal-smith: done!

PR #509, opened 2020-02-21
Fix locale json import 

lucafalasco:
### Updates
- add `esModuleInterop: true` flag in `tsconfig.json` to fix json import


theiliad:
> @lucafalasco not sure exactly what this is fixing... The JSON import is working fine, is it not working for you?

I see, weird... It works fine on Codesandbox

PR #515, opened 2020-02-25
feat(core): Add configuration option to not include 0 on axes

### LINKED ISSUE #402: [LineChart]: support y-axis basepoint starting above zero for large values

SerenaG:
### Updates
Closes #402.
Add configuration option `includeZero?: boolean` (default true) to let users exclude 0 from the axis. 

### Screenshots
`includeZero: true` (default behaviour)
include

`includeZero: false`
not include

- [ ] Demos all features
- [ ] Documented/annotated
- [ ] Matches UI/UX specs
- [ ] Accessible

### DIFF
--- a/packages/core/src/components/axes/two-dimensional-axes.ts
+++ b/packages/core/src/components/axes/two-dimensional-axes.ts
@@ -1,6 +1,6 @@
 // Internal Imports
 import { Component } from "../component";
-import { AxisPositions, ScaleTypes } from "../../interfaces";
+import { AxisPositions, ScaleTypes, AxesOptions } from "../../interfaces";
 import { Axis } from "./axis";
 import { Tools } from "../../tools";
 import { DOMUtils } from "../../services";
@@ -19,40 +19,15 @@ export class TwoDimensionalAxes extends Component {
 
 	render(animate = false) {
 		const axes = {};
-
 		const axisPositions = Object.keys(AxisPositions);
 		const axesOptions = Tools.getProperty(this.model.getOptions(), "axes");
 
-		if (axesOptions) {
-			let primaryAxisOptions, secondaryAxisOptions;
-			axisPositions.forEach(axisPosition => {
-				const axisOptions = axesOptions[AxisPositions[axisPosition]];
-				if (axisOptions) {
-					axes[AxisPositions[axisPosition]] = true;
-
-					if (axisOptions.primary === true) {
-						primaryAxisOptions = axisOptions;
-					} else if (axisOptions.secondary === true) {
-						secondaryAxisOptions = axisOptions;
-					}
-				}
-			});
-		} else {
-			this.model.getOptions().axes = {
-				left: {
-					primary: true,
-					includeZero: true
-				},
-				bottom: {
-					secondary: true,
-					type: this.model.getDisplayData().labels ? ScaleTypes.LABELS : undefined,
-					includeZero: true
-				}
-			};
-
-			axes[AxisPositions.LEFT] = true;
-			axes[AxisPositions.BOTTOM] = true;
-		}
+		axisPositions.forEach(axisPosition => {
+			const axisOptions = axesOptions[AxisPositions[axisPosition]];
+			if (axisOptions) {
+				axes[AxisPositions[axisPosition]] = true;
+			}
+		});
 
 		this.configs.axes = axes;
 
--- a/packages/core/src/components/axes/two-dimensional-axes.ts
+++ b/packages/core/src/components/axes/two-dimensional-axes.ts
@@ -40,11 +40,13 @@ export class TwoDimensionalAxes extends Component {
 		} else {
 			this.model.getOptions().axes = {
 				left: {
-					primary: true
+					primary: true,
+					includeZero: true
 				},
 				bottom: {
 					secondary: true,
-					type: this.model.getDisplayData().labels ? ScaleTypes.LABELS : undefined
+					type: this.model.getDisplayData().labels ? ScaleTypes.LABELS : undefined,
+					includeZero: true
 				}
 			};
 
--- a/packages/core/src/configuration.ts
+++ b/packages/core/src/configuration.ts
@@ -82,7 +82,7 @@ export const axisChartTooltip: AxisTooltipOptions = Tools.merge({}, baseTooltip,
 	}
 } as AxisTooltipOptions);
 
-export const barChartTooltip: BarTooltipOptions = Tools.merge({}, axisChartTooltip , {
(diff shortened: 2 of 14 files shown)


packages/core/src/services/scales-cartesian.ts
```diff
@@ -218,7 +231,7 @@ export class CartesianScales extends Service {
 	protected getScaleDomain(axisPosition: AxisPositions) {
 		const options = this.model.getOptions();
 		const axisOptions = Tools.getProperty(options, "axes", axisPosition);
-
+		const includeZero = axisOptions.includeZero === undefined || axisOptions.includeZero;
```
theiliad: why wouldn't we just do `axisOptions.includeZero` here? If it's undefined or null it'll be falsy, and if not it'll resolve

packages/core/src/services/scales-cartesian.ts
```diff
@@ -304,14 +317,7 @@ export class CartesianScales extends Service {
...
 
-		// TODO - Work with design to improve logic
-		domain[1] = domain[1] * 1.1;
-
-		// if the lower bound of the domain is less than 0, we want to add padding
-		if (domain[0] < 0) {
-			domain[0] = domain[0] * 1.1;
-		}
-		return domain;
+		return addPaddingInDomain(domain, 10);
```
theiliad: What's `10`? Shouldn't that be configurable and stored somewhere in `Configuration.ts`?

SerenaG: Added it as `paddingRatio` into axis configuration.

packages/core/src/services/scales-cartesian.ts
SerenaG: The reason we do this check is that if the user does not set a value, `axisOptions.includeZero` will be undefined and it has to be treated as true by default.

We've added the logic to make possible to add default values into axes too. 
So now as default we set `axisOptions.includeZero` as true and we can just do `axisOptions.includeZero`.

theiliad: Can we please add a comment then saying what the `||` conditional is doing?

packages/core/src/tools.ts
```diff
@@ -170,7 +197,7 @@ export namespace Tools {
 	 * @returns The function to be used by D3 to push element to the top of the canvas
 	 */
 	export function moveToFront(element) {
-		return element.each(function() {
+		return element.each(function () {
```
theiliad: extra space here

packages/core/src/components/axes/two-dimensional-axes.ts
```diff
@@ -40,11 +40,13 @@ export class TwoDimensionalAxes extends Component {
...
+					primary: true,
+					includeZero: true
 				},
 				bottom: {
 					secondary: true,
-					type: this.model.getDisplayData().labels ? ScaleTypes.LABELS : undefined
+					type: this.model.getDisplayData().labels ? ScaleTypes.LABELS : undefined,
+					includeZero: true
 				}
 			};
```
theiliad: Are we able to eliminate having to do this by performing the logic inside of `mergeDefaultChartOptions`?

packages/core/src/services/scales-cartesian.ts
```diff
@@ -218,7 +232,7 @@ export class CartesianScales extends Service {
 	protected getScaleDomain(axisPosition: AxisPositions) {
 		const options = this.model.getOptions();
 		const axisOptions = Tools.getProperty(options, "axes", axisPosition);
-
+		const includeZero = axisOptions.includeZero;
```
theiliad:
```suggestion
		const { includeZero } = axisOptions;
```

packages/core/src/tools.spec.ts
```diff
@@ -37,3 +37,61 @@ describe("Tools.getProperty", () => {
...
+				left: {
+					includeZero: true
+				},
+				right: {
+					includeZero: true
+				}
+			},
+			timeScale: {
+				addSpaceOnEdges: 1
+			},
```
theiliad:
```suggestion
			}
```

packages/core/src/tools.spec.ts
```diff
@@ -37,3 +37,61 @@ describe("Tools.getProperty", () => {
...
+					secondary: true
+				},
+				left: {
+					includeZero: true,
+					primary: true
+				}
+			},
+			timeScale: {
+				addSpaceOnEdges: 1
+			},
```
theiliad:
```suggestion
			}
```

packages/core/src/configuration.ts
```diff
@@ -287,7 +301,8 @@ export const axis = {
 	ticks: {
 		number: 7,
 		rotateIfSmallerThan: 30
-	}
+	},
+	paddingRatio: 0.1
```
theiliad: What does this do now? Does it clash with `addSpaceOnEdges`?

packages/core/src/services/scales-cartesian.ts
theiliad: 👍

packages/core/src/components/axes/two-dimensional-axes.ts
SerenaG: We can do it easily by setting these as default axes configuration and then if axes are not provided take the left and bottom default ones. 
The problem is that for set `type` (that should be `scaleType`) we don't have the model.

Also, we can remove these lines
https://github.com/carbon-design-system/carbon-charts/blob/535378fb4c02a7fd9bddb5dcca5d134703428d17/packages/core/src/components/axes/two-dimensional-axes.ts#L32-L37

packages/core/src/configuration.ts
SerenaG: Screenshot 2020-02-26 at 20 04 08

We just improved the old logic to add padding, where we use the aspect ratio to calculate the padding we should add to the domain, based on domain length.

packages/core/src/components/axes/two-dimensional-axes.ts
theiliad: Oh right. We should rename `type` to `scaleType` (we should also probably wrap the object to the correct TS type so it would shout if we mess it up down the road)

Anything we can remove out of this file in terms of setting the correct defaults and move into the `CartesianScales` service would be great!

PR #524, opened 2020-03-09
Fix function that calculates size of SVG DOM elements

theiliad:
added a bit of validation to ensure the best possible dimensions returned for element

solves the issue of incorrect sizing when hiding and showing the chart back-to-back

### DIFF
--- a/packages/core/src/services/essentials/dom-utils.ts
+++ b/packages/core/src/services/essentials/dom-utils.ts
@@ -27,7 +27,9 @@ export class DOMUtils extends Service {
 				if (dimensions[dimensionKey]) {
 					const dimension = dimensions[dimensionKey];
 					const dimensionNumber = parseFloat(dimension);
-					if (dimension && dimensionNumber > finalDimensions[dimensionKey] &&  ("" + dimension).indexOf("%") === -1) {
+					if (dimension &&
+						dimensionNumber > finalDimensions[dimensionKey] &&
+						("" + dimension).indexOf("%") === -1) {
 						finalDimensions[dimensionKey] = dimensionNumber;
 					}
 				}
--- a/packages/core/src/services/essentials/dom-utils.ts
+++ b/packages/core/src/services/essentials/dom-utils.ts
@@ -17,6 +17,23 @@ export class DOMUtils extends Service {
 			svgSelector = select(svgSelector as any);
 		}
 
+		const finalDimensions = {
+			width: 0,
+			height: 0
+		};
+
+		const validateAndSetDimensions = dimensions => {
+			Object.keys(dimensions).forEach(dimensionKey => {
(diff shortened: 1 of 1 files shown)


packages/core/src/services/essentials/dom-utils.ts
```diff
@@ -17,6 +17,23 @@ export class DOMUtils extends Service {
...
+			width: 0,
+			height: 0
+		};
+
+		const validateAndSetDimensions = dimensions => {
+			Object.keys(dimensions).forEach(dimensionKey => {
+				if (dimensions[dimensionKey]) {
+					const dimension = dimensions[dimensionKey];
+					const dimensionNumber = parseFloat(dimension);
+					if (dimension && dimensionNumber > finalDimensions[dimensionKey] &&  ("" + dimension).indexOf("%") === -1) {
```
Donisius:
```suggestion
					if (
						dimension &&
						dimensionNumber > finalDimensions[dimensionKey] && 
						("" + dimension).indexOf("%") === -1) {
```

Donisius: Not sure if this would look cleaner

zvonimirfras:
```suggestion
					if (dimension &&
						dimensionNumber > finalDimensions[dimensionKey] &&
						("" + dimension).indexOf("%") === -1) {
```

packages/core/src/services/essentials/dom-utils.ts
```diff
@@ -46,34 +63,39 @@ export class DOMUtils extends Service {
...
-
 		// If both attribute values are numbers
 		// And not percentages or NaN
 		if (options) {
-			if (options.useAttrs && validateDimensions(attrDimensions)) {
-				return attrDimensions;
+			if (options.useAttrs) {
+				validateAndSetDimensions(attrDimensions);
+
+				if (finalDimensions.width > 0 && finalDimensions.height > 0) {
```
zvonimirfras: Can these checks be part of `validateAndSetDimensions` so you can just `return validateAndSetDimensions(...)`?
It doesn't seem like it would affect the cases that don't use a return value.

packages/core/src/services/essentials/dom-utils.ts
Donisius: Z beat me to it 😓

packages/core/src/services/essentials/dom-utils.ts
theiliad: This is really about priority of alternative calculations within the function.

If I always `return validateAndSet...()` then none of the sections below it would execute

theiliad:
@zvonimirfras @Donisius good to go?

PR #520, opened 2020-03-04
Storybook in core

ilariaventurini:
This PR adds Storybook in the core package.

Each demo object has this structure:

    {
      options: ...,
      data: ...,
      chartType: ...,
      isDemoExample: boolean,
    }

`isDemoExample` allows you to choose whether the example should be visible only in the storybook or also in the demo page.

So, if:
- `isDemoExample = true` -> the example is inserted both in the demo page and in the Storybook
- `isDemoExample = false` -> the example is inserted only in the Storybook.

### Review checklist (for reviewers only)
- [ ] Demos all features
- [ ] Documented/annotated
- [ ] Matches UI/UX specs
- [ ] Meets the [code style guide](https://github.com/IBM/carbon-charts/wiki/Code-style-guide)
- [ ] Accessible
- [ ] Mobile first (responsive)
- [ ] RTL support (bidirectional text)
- [ ] Performant (limited bloat)

PR #528, opened 2020-03-10
 Update carbon colors and themes

### LINKED ISSUE #518: Gray 100 background has wrong hex value
I see #171717 as fill in `.bx--cc--grid rect.chart-grid-backdrop`. It should be $161616 per https://www.carbondesignsystem.com/guidelines/color/usage.

Updating to a newer `@carbon/themes` should fix this?

https://github.com/carbon-design-system/carbon/blob/master/packages/colors/src/colors.js#L213

sophiiae:
closes #518


theiliad:
@sophiiae Thanks this is great! The Travis build seems to be broken

theiliad:
Addressed in #520

Issue #516, opened 2020-02-25
Proposal: Dataset in Tabular Format


caesarsol:
Hi, this is a proposal for @theiliad, about a change in the shape of the `data` object given as chart input.

This is based on a pattern we often use at Accurat, which is to always start from non-nested datasets.

```js
// Dataset with columns `x`, `y`, `category`

const tabularDataset = [
  { x: new Date(2020, 1, 1), y: 32100, category: 'A' },
  { x: new Date(2020, 1, 2), y: 23500, category: 'A' },
  { x: new Date(2020, 1, 3), y: 53100, category: 'A' },
  { x: new Date(2020, 1, 4), y: 42300, category: 'A' },
  { x: new Date(2020, 1, 5), y: 12300, category: 'A' },
]
```

We find that tabular datasets are, for the fact that they avoid nesting:
 - easy to generate for the developer using the chart, very often being that a CSV import
 - easy to manipulate from the chart internals
 - easy to debug, being each datapoint/datasymbol associated to a single dataset row/object
 - pretty simple to transform into nested format (`lodash.groupBy` or `d3.nest`)
 - usable for different charts, without changing the shape. They only depend on the semantic meaning of the chart.

Follows two cases of conversions from old to new proposed format.

### AS IS:
https://github.com/carbon-design-system/carbon-charts/blob/master/packages/core/demo/demo-data/line.ts#L81-L185
```js
const lineData_AS_IS = {
  labels: ["Qty", "More", "Sold", "Restocking", "Misc"],
  datasets: [
    {
      label: "Dataset 1",
      data: [32100, 23500, 53100, 42300, 12300]
    },
    {
      label: "Dataset 2",
      data: [34200, 53200, 42300, 21400, 0]
    },
    {
      label: "Dataset 3 long name",
      data: [41200, 23400, 34210, 1400, 42100]
    },
    {
      label: "Dataset 4 long name",
      data: [22000, 1200, 9000, 24000, 3000]
    },
    {
      label: "Dataset 5 long name",
      data: [2412, 30000, 10000, 5000, 31000]
    },
    {
      label: "Dataset 6 long name",
      data: [0, 20000, 40000, 60000, 80000]
    }
  ]
};

const lineTimeSeriesData_AS_IS = {
  datasets: [
    {
      label: "Dataset 1",
      data: [
        { date: new Date(2019, 0, 1), value: 10000 },
        { date: new Date(2019, 0, 5), value: 65000 },
[...]

PR #529, opened 2020-03-11
Tabular data format support

### LINKED ISSUE #206: Bar chart with single bar not centered on label
I am submitting a...

- [ ] Feature request
- [ ] Design defect
- [x] Source code defect
- [ ] Other

charts version:
0.10.2

Issue description:
I have bar chart with grouped data, when their is only one value (one bar) the bar is not centered on the x-axis

Steps to produce the issue:
Supply multiply datasets to bar chart, with only have one value
code example: 
        if (tempValue > 90) {
          dataSafeLevel.push(null);
          dataWarningLevel.push(null);
          dataNeedsActionLevel.push(tempValue);
        } else if (tempValue > 50) {
          dataSafeLevel.push(null);
          dataWarningLevel.push(tempValue);
          dataNeedsActionLevel.push(null);
        } else {
          dataSafeLevel.push(tempValue);
          dataWarningLevel.push(null);
          dataNeedsActionLevel.push(null);
        }
      });

    return {
      labels: labels,
      datasets: [
        {
          backgroundColors: ['#4178be'], // $color__blue-50
          data: dataSafeLevel,
          label: 'Safe',
        },
        {
          backgroundColors: ['#734098'], //$color__purple-60
          data: dataWarningLevel,
          label: 'Warning',
        },
        {
          backgroundColors: ['#ad1625'], // '$color__red-60
          data: dataNeedsActionLevel,
          label: 'Needs Action',
        },
      ],
    };
Current behavior:
Single bar does not line up

Expected behavior:
Bar is centered

Screenshot or recording
image

showing array with values & nulls in each index.
image

### LINKED ISSUE #521: Stacked bar chart timeseries missing labels cause the bars to go off the grid
#### I am submitting a...

- [ ] Feature request
- [ ] Design defect
- [x] Source code defect
- [x] Demo/documentation defect
- [ ] Other

#### charts version:
0.29.8

### Issue description
When setting the `options` for a stacked bar time series, it's not clear what `labels` does. If you leave a quantity of labels that is less than the number of any datasets data points, it renders bars outside the grid.

It might be related to a possible bug I noticed here: https://github.com/carbon-design-system/carbon-charts/blob/master/packages/core/src/components/graphs/bar-stacked.ts#L147

`yScale.scaleType` is always `undefined` because is not a valid property of `d3.scale()`  (while it's used correctly [here](https://github.com/carbon-design-system/carbon-charts/blob/master/packages/core/src/components/graphs/bar-stacked.ts#L49))

- change [this line](https://github.com/carbon-design-system/carbon-charts/blob/master/packages/core/demo/demo-data/bar.ts#L281) on the demo data to `labels: ["Qty"],`
- press 'Update data' as many times as needed to reproduce the problem

### Screenshot or recording
image

theiliad:
This is a WIP and potentially a place for us to discuss some of the specifics of the ideal charting data format.

**Updates:**
- Adds support for tabular data
- As a side-effect we'll automatically add support for multiple linear axes
- No longer need `primary` & `secondary` in axis options
- Ability to provide an `identifier` for each axis that the axis will use to look through the data for corresponding values
- Ability to customize `groupIdentifier`

**TODO:**
- [x] Support for all chart types
- [x] Support the existing data format
- [x] Tooltips
  - [x] Fix getDataFromDomain()
- [x] Add write-up about new data format
- [x] Fix stacked bar transition issues
- [x] Remove `primary` & `secondary`
- [x] Remove `useAsGraphDomain` & `range`
- [x] Fix time-series padding issue
- [ ] Remove graph listeners in `destroy()`
- [x] `fillColors`?
- [x] Fix groupedBar positioning issues

I've got a basic demo in the `core` deploy preview.

**Example:**
```js
export const lineData = [
	{ group: "Dataset 1", key: "Qty", value: 32100 },
	{ group: "Dataset 1", key: "More", value: 23500 },
	{ group: "Dataset 1", key: "Sold", value: 53100 },
	{ group: "Dataset 1", key: "Restocking", value: 42300 },
	{ group: "Dataset 1", key: "Misc", value: 12300 },
	{ group: "Dataset 2", key: "Qty", value: 34200 },
	{ group: "Dataset 2", key: "More", value: 53200 },
	{ group: "Dataset 2", key: "Sold", value: 42300 },
	{ group: "Dataset 2", key: "Restocking", value: 21400 },
	{ group: "Dataset 2", key: "Misc", value: 0 },
];

export const lineOptions = {
	title: "Line (discrete)",
	axes: {
		bottom: {
			title: "2018 Annual Sales Figures",
			identifier: "key",
			scaleType: "labels"
		},
		left: {
			identifier: "value",
			scaleType: "linear"
		}
	}
};
```

**Things to consider:**
We're going to need multiple utils possibly within `Model` that'll give us:
- List of all `group`s
- All corresponding data relating to a `group`
- We should assign `left` axis to the `value` identifier & the `bottom` axis to `key` (this will require some of the "detault axis options" logic to be merged in first - #515 

**Questions:**
- Can we assume there’ll always be a `group` field within every datum? (If not, what if some datums have it and some don’t?)
- Would we still need `scaleType`?

@cal-smith @zvonimirfras @natashadecoste @caesarsol 

Closes #206
Closes #521


caesarsol:
From the discussion in https://github.com/carbon-design-system/carbon-charts/issues/516 and other thoughts with @theiliad and my team:

> Would we still need `scaleType`?

The `scaleType` is definitely useful to understand how to interpret the column:
 - there could be stringified dates, and `scaleType: 'time'` enables us to parse them into `new Date()`s. As I said, I suggest supporting a small set of formats (such as YYYY-MM-DD and other ISO international formats), the user-developer will use other libraries to convert if needed.
 - there could be numbers to be used as categories, so `scaleType: 'labels'` would enable us convert `1` into `"1"`
 - the same with `scaleType: 'linear'` and stringified numbers `"2.345"` -> `2.345` (very often a product of CSV imports)

> Can we assume there’ll always be a `group` field within every datum? (If not, what if some datums have it and some don’t?)

I think the `missing data` problem (or `null` or `undefined`) could be approached this way:
 - `scaleType: 'labels'`: if a label is missing, we convert it to a `"N/A"` label and treat it as a normal label. If all labels are missing, we just treat it as a single unnamed label and hide the legend.
 - `scaleType: 'time'`: we can probably ignore it altogether, but we should probably trigger an application error
 - `scaleType: 'linear'`: it's really a "missing data" problem, it probably already has an approach. If the chart has a visual treatment for those – such as Linechart after #464 – we have to use it, otherwise we can either treat as zero, ignore it, error out. Probably depending on the chart type.


A final note: would you rely on an implicitly declared `group` key in all the data objects, or do you think it could be another customizable option as `identifier`? Maybe a new axis called `group`, or `color`, or a `legend` option.

theiliad:
> From the discussion in #516 and other thoughts with @theiliad and my team:
> 
> > Would we still need `scaleType`?
> 
> The `scaleType` is definitely useful to understand how to interpret the column:
> 
> * there could be stringified dates, and `scaleType: 'time'` enables us to parse them into `new Date()`s. As I said, I suggest supporting a small set of formats (such as YYYY-MM-DD and other ISO international formats), the user-developer will use other libraries to convert if needed.
> * there could be numbers to be used as categories, so `scaleType: 'labels'` would enable us convert `1` into `"1"`
> * the same with `scaleType: 'linear'` and stringified numbers `"2.345"` -> `2.345` (very often a product of CSV imports)
> 
> > Can we assume there’ll always be a `group` field within every datum? (If not, what if some datums have it and some don’t?)
> 
> I think the `missing data` problem (or `null` or `undefined`) could be approached this way:
> 
> * `scaleType: 'labels'`: if a label is missing, we convert it to a `"N/A"` label and treat it as a normal label. If all labels are missing, we just treat it as a single unnamed label and hide the legend.
> * `scaleType: 'time'`: we can probably ignore it altogether, but we should probably trigger an application error
> * `scaleType: 'linear'`: it's really a "missing data" problem, it probably already has an approach. If the chart has a visual treatment for those – such as Linechart after #464 – we have to use it, otherwise we can either treat as zero, ignore it, error out. Probably depending on the chart type.
> 
> A final note: would you rely on an implicitly declared `group` key in all the data objects, or do you think it could be another customizable option as `identifier`? Maybe a new axis called `group`, or `color`, or a `legend` option.

I think if we're sure that every type of vis we're going to support requires the concept of groups then we can assume that fieldname to show up rather than adding a customizable identifier for it

cal-smith:
`identifier` is slightly awkward ... I'd prefer `key` or `id` (same thing I know, but more conventional)

I wouldn't have `group` as a special field within the row ... `group` as a config for the scale should be able to accomplish the same thing, and let users use whatever key they want to enable grouping

packages/core/demo/data/bar.ts
```diff
@@ -306,16 +307,53 @@ export const stackedBarTimeSeriesData = {
...
+// 	{ group: "Dataset 4", date: new Date(2019, 0, 15), value: 40323 },
+// 	{ group: "Dataset 4", date: new Date(2019, 0, 19), value: 31300 },
+// ];
+
 export const stackedBarTimeSeriesOptions = {
 	title: "Stacked bar (time series)",
 	axes: {
 		left: {
-			primary: true,
+			identifier: "value",
```
cal-smith: `identifier` is awkward, and kinda unexpected ... As a user I would expect to use a `key` or `id` property instead of `identifier` ... any reason we'd want to stick with `identifier` over something else?

packages/core/demo/data/bar.ts
```diff
@@ -306,16 +307,53 @@ export const stackedBarTimeSeriesData = {
...
+// 	{ group: "Dataset 3", date: new Date(2019, 0, 5), value: 15000 },
+// 	{ group: "Dataset 3", date: new Date(2019, 0, 8), value: 20000 },
+// 	{ group: "Dataset 3", date: new Date(2019, 0, 13), value: 39213 },
+// 	{ group: "Dataset 3", date: new Date(2019, 0, 17), value: 61213 },
+// 	{ group: "Dataset 4", date: new Date(2019, 0, 2), value: 10 },
+// 	{ group: "Dataset 4", date: new Date(2019, 0, 6), value: 37312 },
+// 	{ group: "Dataset 4", date: new Date(2019, 0, 8), value: 51432 },
+// 	{ group: "Dataset 4", date: new Date(2019, 0, 15), value: 40323 },
+// 	{ group: "Dataset 4", date: new Date(2019, 0, 19), value: 31300 },
+// ];
```
cal-smith: Are these samples or can we remove them?

theiliad: good catch. will remove

packages/core/src/model.ts
```diff
@@ -80,23 +110,122 @@ export class ChartModel {
...
+		}));
+	}
+
+	getDataGroups() {
+		return this.get("dataGroups");
+	}
+
+
+	getDataGroupNames() {
+		return this.get("dataGroups").map(dataGroup => dataGroup.name);
```
zvonimirfras:
```suggestion
		return this.getDataGroups().map(dataGroup => dataGroup.name);
```
might make sense, conceptually

packages/core/src/model.ts
```diff
@@ -80,23 +110,122 @@ export class ChartModel {
...
 		});
 
-		return dataLabels;
+		return Object.keys(groupedData)
+			.map(groupName => {
+				return {
+					name: groupName,
+					data: groupedData[groupName]
+				};
+			});
```
zvonimirfras:
```suggestion
			.map(groupName => ({
				name: groupName,
				data: groupedData[groupName]
			}));
```

packages/core/src/model.ts
```diff
@@ -80,23 +110,122 @@ export class ChartModel {
...
+
+		const stackKeys = map(displayData, datum => datum[domainIdentifier]).keys();
+		const dataGroupNames = this.getDataGroupNames();
+
+		return stackKeys.map(key => {
+			const correspondingValues = { sharedStackKey: key };
+			dataGroupNames.forEach(dataGroupName => {
+				const correspondingDatum = displayData.find(datum => {
+					return datum[groupIdentifier] === dataGroupName &&
+						`${datum[domainIdentifier]}` === key;
```
zvonimirfras: Looks like
```suggestion
						datum[domainIdentifier] === key;
```
should work too?

Or perhaps
```suggestion
						datum[domainIdentifier] as string === key;
```

packages/core/src/tools.ts
```diff
@@ -40,8 +43,29 @@ export namespace Tools {
...
 
+		// Update deprecated options to work with the tabular data format
+		// Similar to the functionality in model.transformToTabularData()
 		for (const axisName in defaultOptions.axes) {
-			if (!providedAxesNames.includes(axisName)) {
+			if (providedAxesNames.includes(axisName)) {
+				const providedAxisOptions = providedOptions.axes[axisName];
+
+				if (providedAxisOptions["primary"] || providedAxisOptions["secondary"]) {
+					console.warn("`primary` & `secondary` are no longer needed for axis configurations. Read more (TODO-TDF) here")
```
zvonimirfras: TODO later?

packages/core/demo/data/bubble.ts
```diff
@@ -1,189 +1,94 @@
...
-				}
-			]
+		left: {
+			title: "Profit",
+			identifier: "profit",
+			includeZero: false
+		}
+	},
+	bubble: {
+		radiusIdentifier: "surplus"
```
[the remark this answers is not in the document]
theiliad: no

packages/core/src/components/graphs/bar-stacked.ts
```diff
@@ -203,16 +124,19 @@ export class StackedBar extends Bar {
...
 	}
 
 	addEventListeners() {
+		const options = this.model.getOptions();
+		const { groupIdentifier } = options.data;
+
 		const self = this;
 		this.parent.selectAll("path.bar")
-			.on("mouseover", function(datum) {
+			.on("mouseover", function (datum) {
```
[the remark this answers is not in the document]
natashadecoste: i noticed that too, i believe there should **not** be one because the linter has it set to never for all function cases 

_tslint.json_
`    "space-before-function-paren": [
      true,
      "never"
    ],`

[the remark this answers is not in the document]
theiliad: that was recently added to the library, so it could be that I hadn't yet pulled down that `tslint` update before pushing this

packages/core/src/components/graphs/bar.ts
```diff
@@ -4,7 +4,7 @@ import { CartesianOrientations } from "../../interfaces";
 
 export class Bar extends Component {
 	// Gets the correct width for bars based on options & configurations
-	protected getBarWidth() {
+	protected getBarWidth(allDataLabels?: string[]) {
```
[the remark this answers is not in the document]
theiliad: the extension of that method in `bar-grouped.ts` uses `allDataLabels`

packages/core/src/model.ts
```diff
@@ -148,6 +154,48 @@ export class ChartModel {
...
+
+	getStackedData() {
+		const options = this.getOptions();
+		const { groupIdentifier } = options.data;
+
+		const dataGroupNames = this.getDataGroupNames();
+		const dataValuesGroupedByKeys = this.getDataValuesGroupedByKeys();
+		return stack().keys(dataGroupNames)(dataValuesGroupedByKeys)
+			.map((series, i) => {
+				// Add data group names to each serie
```
natashadecoste:
```suggestion
				// Add data group names to each series
```

packages/core/src/model.ts
```diff
@@ -24,50 +26,78 @@ export class ChartModel {
...
-		data.datasets.forEach(dataset => {
-			dataset.data = dataset.data.map(d => {
-				if (d.date && !d.date.getTime) {
-					d.date = new Date(d.date);
+	/**
+	 * Converts data provided in the older format to tabular
+	 * 
+	 */
+	protected transformToTabularData(data) {
+		console.warn("We've updated the charting data format to be tabular by default. The current format you're using is deprecated and will be removed in v1.0, read more here")
```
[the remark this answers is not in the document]
theiliad: Yes, I'm writing up a story

packages/core/src/model.ts
```diff
@@ -80,23 +110,122 @@ export class ChartModel {
...
+	 * Data groups
+	*/
+	protected updateAllDataGroups() {
+		// If allDataGroups hasn't been initialized yet
+		// Set it to the current set of data groups
+		if (!this.allDataGroups) {
+			this.allDataGroups = this.getDataGroups().map(group => group.name);
+		} else {
+			// Loop through current data groups
+			this.getDataGroups().forEach(dataGroup => {
```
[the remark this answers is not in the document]
theiliad: `allDataGroups` is used to generate a color scale that applies to all the groups.

Now when the data updates, you might remove a group, and then bring it back in a newer date update, therefore the order of the groups in `allDataGroups` matters so that you'd never have an incorrect color assigned to it.

Also, a new `group` going into `allDataGroups` shouldn't have existed in there before, hence why the `forEach` and the `if` inside of that.

packages/core/src/model.ts
theiliad: that's done to make sure that everything is stringified. e.g. if you had a `number` or a `date` type

theiliad: @zvonimirfras the logic would actually break if we get rid of that for time-series charts

packages/core/src/tools.ts
theiliad: no I'll add that soon in this PR

packages/core/demo/data/bar.ts
theiliad: I added a comment here yesterday which I don't see anymore. I'll post again.

I'd like to involve @natashadecoste @zvonimirfras @caesarsol in the discussion here.

We could really choose any name we'd like instead of `identifier`. The reason I went with `identifier` is because `key` & `id` would show up in different places in the library codebase and could cause confusion in the code.

Also I felt that `key` & `id` are more likely to show up in user data as well, where `identifier` seems a bit unlikely to show up there.

packages/core/src/components/essentials/legend.ts
```diff
@@ -37,14 +38,14 @@ export class Legend extends Component {
 			.attr("rx", 1)
 			.attr("ry", 1)
 			.style("fill", d => {
-				return d.value === options.legend.items.status.ACTIVE ? this.model.getStrokeColor(d.key) : null;
+				return d.status === options.legend.items.status.ACTIVE ? this.model.getStrokeColor(d.name) : null;
 			}).classed("active", function (d, i) {
```
natashadecoste:
```suggestion
			}).classed("active", function (d) {
```

packages/core/demo/data/bar.ts
cal-smith: I'd still argue `key: "key"` or `id: "id"` is more readable vs. `identifier: "id"` ... Internally we can use whatever, `getIdentifier` or `getDimensionKey` or whatever is absolutely appropriate there ... The options on the user side are pretty discoverable right now, so I'd hate to break that with an awkward value selector

cal-smith: I would still propose something along these lines: https://github.com/carbon-design-system/carbon-charts/issues/516#issuecomment-591047052

`map` still isn't totally ideal, but perhaps it's getting closer? Basically something that makes it clear that the relationship is a mapping between the overall dataset and the specific axis.

packages/core/demo/data/bubble.ts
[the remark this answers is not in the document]
zvonimirfras: so let's name it just `identifier`?

packages/core/demo/data/bar.ts
theiliad: The `map` function call is a bit much IMO. handling calling that function every where we need a value is going to be too complicated.

I think if you were intending the values of an axis to be `a + b` mapped through every datapoint, you should just provide `c = a + b` in every datapoint manually

zvonimirfras: I was initially ok with `identifier`. I'd find `id` a bit misleading.
`key` could make sense.
`propName` possibly
`mapsTo` possibly too

Actually `mapsTo` is becoming my favorite 😸

cal-smith: I'm down for `mapsTo` 👍 Bonus points because extending to be a function later would be trivial ... I will note that:

> handling calling that function every where we need a value is going to be too complicated.

Shouldn't be the case ... in the event that we support a function for `mapsTo`, we should have one (and only one) place that we retrieve data internally - that location can handle calling `mapsTo` (and any other functions we may need) ... generally pre-processing the data to add the appropriate columns should be preferred, but I don't think it's unreasonable to design with the eventuality of `string | (any) => any` in mind

cal-smith: I would even argue that _today_ we should have centralized data sourcing, but that's a different conversation 👍

packages/core/demo/data/bubble.ts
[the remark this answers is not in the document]
caesarsol: hi, I think this `radius` could be thought as a theoretical new Axis, because it's like a "third dimension". We could keep the same shape of the other `top, left, right, bottom` axes, so:
```
bubble: {
  radius: {
    identifier: "xxx",
    scaleType: "linear" // does not make much sense to use different scaleTypes here, but we could support it anyway and the choice can be left to the user
    // ...
  }
}
```

I think the main advantage here would be the future extensibility. For example, that would be the place to change the limits of the scale, or some legend configurations.

packages/core/src/model.ts
caesarsol: I'd maybe suggest a more explicit `.toString()` to make it clear we need it to be a string.

packages/core/demo/data/bar.ts
caesarsol: - In data science these are often called "features", but I doubt `feature` would be a clear name for users of any background.
 - In JS this is of course an object `key`, but I think this is more of an implementation detail than a logic step;
   - If we look at similar concepts used by D3, pretty much a standard in this world, they often use the term `accessor`: https://github.com/d3/d3-shape#line_x
   - Another standard worth mentioning: Lodash has the `property` function which accesses an object/array property by string: https://lodash.com/docs/#property
 - If we look at it from the point of view of the data, being it tabular would make this a `column`. This would also make clear that the data is thought as a table, and it would be decoupled from the actual array-of-objects representation.

I guess we need to define if we want to only support strings, or add functions as well as @cal-smith proposes. @theiliad is actually making a good point saying that a simple `.map(d => ({ ...d, c: d.a + d.b }))` on the dataset to add a new "column" is pretty easy to do for the user.

If we choose only string properties, I'd go with `column`.
Otherwise I think `accessor` is pretty explicit to support both strings and functions.

packages/core/demo/data/bubble.ts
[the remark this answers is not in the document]
zvonimirfras: That makes a lot of sense actually.

packages/core/src/model.ts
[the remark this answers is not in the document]
zvonimirfras: That makes sense. Maybe add a comment that explains that?

packages/core/demo/data/bar.ts
cal-smith: Great points @caesarsol 👍 Definitly agree that `feature` or `dimension` would be unintuitive for most users ... `column`, `accessor`, and `mapsTo` are definitely my favorite options right now ... My preference leans toward `column` or `mapsTo`, just because `accessor` isn't the usual term outside of the D3 space - it's a reasonable option, but not my first choice.

natashadecoste: I think my preference is towards `mapsTo` or `dimension`

packages/core/demo/data/bubble.ts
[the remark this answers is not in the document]
cal-smith: probably should be `radiusMapsTo` now 👍

packages/core/src/configuration.ts
```diff
@@ -138,6 +138,9 @@ const chart: BaseChartOptions = {
 	legend,
 	style: {
 		prefix: "cc"
+	},
+	data: {
+		groupIdentifier: "group"
```
cal-smith, line 143: Maybe `groupMapsTo`? `groupIdentifier` isn't bad, but if everything else is `mapsTo` and `xMapsTo` maybe we want to change it?

packages/core/demo/data/bubble.ts
```diff
@@ -1,189 +1,94 @@
...
-				}
-			]
+		left: {
+			title: "Annual sales",
+			mapsTo: "profit",
+			includeZero: false
+		}
+	},
+	bubble: {
+		radiusMapsTo: "surplus"
```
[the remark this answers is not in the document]
caesarsol: I support this idea! In the future, every related option could go in that object - sooner or later you'll need a domain customization or max radius.

The same for group as a "color axis", IMO.

PR #533, opened 2020-03-12
fix(core): add shared label to customHTML function

natashadecoste:
### Updates
the function now gets passed the multidata object containing the activeData (points with datum) and the label (shared x-axis label) as well as the defaultHTML to use 

- [ ] Demos all features
- [ ] Documented/annotated
- [ ] Matches UI/UX specs
- [ ] Accessible


natashadecoste:
> Looks good 💯
> Is there a cleaner way you can think of where we might be able to achieve the same result?
> 
> Also are there any implications with other tooltip types that this change will bring? Any considerations?

Not sure what you mean, unless you think the naming is weird, because there wasn't anything extra added in realms of computation, just making sure to pass in the shared label for associated tooltip items. It does mean that data is within an object called `multidata` which then contains the `activeData` (points and data) as well as their shared `label`. This is different before since multidata just had all the activeData in it, but i think its better to keep that separate from the other event logic (tooltip type, actual hovered element, etc). 

As for implications, it only affects multidata tooltips since the items sent to the tooltip function are grouped together by some label - the single data points have all the required data already attached (unless there is a new requirement). I did look into the regular tooltip (single tooltips) and they seems to have all mapping data passed as well as values and i think the user has enough to go with to create custom tooltip implementations

PR #544, opened 2020-03-27
Legend filtering event

### LINKED ISSUE #541: Dispatching an event on Legend filter
#### I am submitting a...

- [ x] Feature request
- [ ] Design defect
- [ ] Source code defect
- [ ] Other

#### Expected behavior
When a legend's checkbox is selected/deselected an event should trigger with the data of which of the Legends are selected

natashadecoste:
### Updates
- dispatch event when legend items are changed (active status to disabled) by filtering
- event has all the data labels and their associated status (active: 1/true, disabled: 0/false)

closes #541 

- [ ] Demos all features
- [ ] Documented/annotated
- [ ] Matches UI/UX specs
- [ ] Accessible

### DIFF
--- a/packages/core/src/model.ts
+++ b/packages/core/src/model.ts
@@ -180,6 +180,11 @@ export class ChartModel {
 			});
 		}
 
+		// dispatch legend filtering event with the status of all the dataLabels
+		this.services.events.dispatchEvent("legend-items-update", {
+			dataLabels
+		});
+
 		// Update model
 		this.set({
 			dataLabels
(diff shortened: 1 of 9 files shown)


packages/core/src/model.ts
```diff
@@ -180,6 +180,11 @@ export class ChartModel {
 			});
 		}
 
+		// dispatch legend filtering event with the status of all the dataLabels
+		this.services.events.dispatchEvent("legend-items-update", {
```
theiliad: We should use an ENUM for the event name like so https://github.com/carbon-design-system/carbon-charts/blob/79c1bc714aa9c6bd6971c49cf92c4122e7faf84e/packages/core/src/components/graphs/bar-simple.ts#L142

PR #547, opened 2020-03-30
fix(core): title string contains html ascii characters 

natashadecoste:
Title strings passed into the options are HTML number encoded (apostrophes, etc -> &#39;)
Setting to `.text()` renders the string as is, setting to `.html()` renders the string after it is decoded. The DOM html is unaffected by the change. 

This bug appears in the legend (see below) and axis titles but not within the toolltip (toolltip passes the strings through `.html()`)

image

- [ ] Demos all features
- [ ] Documented/annotated
- [ ] Matches UI/UX specs
- [ ] Accessible


packages/core/src/components/essentials/title.ts
```diff
@@ -35,22 +35,22 @@ export class Title extends Component {
...
-			title.text(titleString.substring(0, substringIndex - 1))
+			title.html(titleString.substring(0, substringIndex - 1))
 				.append("tspan")
 				.text("...");
 
 			// add events for displaying the tooltip with the title
 			const self = this;
 			title
-				.on("mouseenter", function() {
+				.on("mouseenter", function () {
```
theiliad:
```suggestion
				.on("mouseenter", function() {
```

packages/core/src/components/essentials/title.ts
```diff
@@ -35,22 +35,22 @@ export class Title extends Component {
...
 					self.services.events.dispatchEvent("show-tooltip", {
 						hoveredElement: title,
 						type: TooltipTypes.TITLE
 					});
 				})
-				.on("mouseout", function() {
+				.on("mouseout", function () {
```
theiliad:
```suggestion
				.on("mouseout", function() {
```

packages/core/demo/data/line.ts
```diff
@@ -22,7 +22,7 @@ export const lineData = [
 ];
 
 export const lineOptions = {
-	title: "Line (discrete)",
+	title: "Line (discrete) - long long long long long long long long long long long long long long  time format ('MMM d, hh a and 'hh a')",
```
theiliad, line 25: we should just keep this short like before. as long as it works we're good

PR #549, opened 2020-03-31
Stacked Area chart

lucafalasco:
Screenshot 2020-03-31 at 16 58 42

### Deploy
https://deploy-preview-549--carbon-charts-core.netlify.com/?path=/story/area--area-stacked-time-series


theiliad:
Glad to see this shaping up!

The colors look a bit off, shouldn't they be matching this? https://deploy-preview-412--carbon-charts-core.netlify.com

Also why would we need `multiply`?

lucafalasco:
> Glad to see this shaping up!
> 
> The colors look a bit off, shouldn't they be matching this? https://deploy-preview-412--carbon-charts-core.netlify.com
> 
> Also why would we need `multiply`?

hey @theiliad, it was a design decision to keep the colors solid and use `mix-blend-mode` to show the grid underneath. However there seems to be an issue with some css conditions/variables, specifically those if statements seem to have no effect:  
https://github.com/carbon-design-system/carbon-charts/blob/8729d54ff9622177d066ccd60e76b71d8dfefd23/packages/core/src/styles/graphs/_area.scss#L3-L9

you can see that with the dark theme when selecting the area chart, the colors look pretty dark due to that css property taking the `multiply` value instead of `screen`
do you have any idea why?

caesarsol:
@theiliad this was the CSS problem I mentioned yesterday. It seems the SCSS is not picking up the property?

theiliad:
> @theiliad this was the CSS problem I mentioned yesterday. It seems the SCSS is not picking up the property?

Hmm.. try doing a `yarn build` in `core` then `yarn start`. We need to refactor some of the storybook build logic for sure, it doesn't seem to be working well

lucafalasco:
> Hmm.. try doing a `yarn build` in `core` then `yarn start`. We need to refactor some of the storybook build logic for sure, it doesn't seem to be working well

Yes I tried that too.. apparently still not working though 🤔

designertyler:
I'm not sure if these are related to the CSS issue

These guys are hanging over the Y axis

image

There's a weird transition happening to the area borders when hovering on a label. It looks like we're overlapping colors to create The border but the transition isn't a smooth as the other colors fading in/out. Is this a result of the changing blending modes and it possible to make it less noticeable? 

Apr-09-2020 14-50-56

lucafalasco:
FYI I converted this back to draft because rebasing onto `master` to include 36ddddd09e6e0a8053967226f12de3d3554dfa27 broke some of the data aggregation logic, will first fix those issues and then move on to the design tweaks

designertyler:
Those changes look good 👍 

The tick marks seem to pop instead of smoothly transitioning back when hovering off the legend. You can see it in the purple area when hovering off Dataset 3. They go away and then pop back.

Apr-22-2020 09-05-23

caesarsol:
Note: the chart does not have the [new Ruler interaction](https://github.com/carbon-design-system/carbon-charts/pull/499) because it's in a separate PR.

@theiliad does it make sense to implement the "old" grid behaviour, or should we wait for the Ruler PR merge?

theiliad:
> Note: the chart does not have the [new Ruler interaction](https://github.com/carbon-design-system/carbon-charts/pull/499) because it's in a separate PR.
> 
> @theiliad does it make sense to implement the "old" grid behaviour, or should we wait for the Ruler PR merge?

we should just pull the ruler updates since it's merged now. shouldn't require too many changes in this PR right?

theiliad:
looks great. we were supposed to show the highlighting points once the tooltip is activated, is that coming soon?

@designertyler do you see any weirdness in the lack of points in the area? do the gridlines make you think that you would see a tooltip on those lines onHover?

lucafalasco:
> looks great. we were supposed to show the highlighting points once the tooltip is activated, is that coming soon?

I hope we haven't missed something, but we understood it was not expected from the design, also because those points could disappear permanently from other charts too

caesarsol:
@theiliad we understood the highlighted points were not included in the design, because they were going to be removed from all the designs, is that uncorrect? Probably Mike said this in a design call, but maybe he meant in the next future. 
In that case, we probably won't have the time to finish this change today. We are looking into that anyway.

packages/core/src/charts/area-stacked.ts
```diff
@@ -0,0 +1,54 @@
...
+	Line,
+	Ruler,
+	Tooltip
+} from "../components/index";
+
+export class StackedAreaChart extends AxisChart {
+	constructor(
+		holder: Element,
+		chartConfigs: ChartConfig<StackedAreaChartOptions>
+	) {
```
natashadecoste, line 21:
```suggestion
	constructor(holder: Element, chartConfigs: ChartConfig<StackedAreaChartOptions>) {
```

packages/core/src/components/graphs/area-stacked.ts
```diff
@@ -0,0 +1,186 @@
...
+	init() {
+		// Highlight correct area on legend item hovers
+		this.services.events.addEventListener("legend-item-onhover", e => {
+			const { hoveredElement } = e.detail;
+
+			this.parent
+				.selectAll("g.areas")
+				.transition(
+					this.services.transitions.getTransition("legend-hover-area")
+				)
```
natashadecoste, line 25:
```suggestion
				.transition(this.services.transitions.getTransition("legend-hover-area"))
```

packages/core/src/components/graphs/area-stacked.ts
```diff
@@ -0,0 +1,186 @@
...
+
+		// Un-highlight areas on legend item mouseouts
+		this.services.events.addEventListener("legend-item-onmouseout", e => {
+			this.parent
+				.selectAll("g.areas")
+				.transition(
+					this.services.transitions.getTransition(
+						"legend-mouseout-area"
+					)
+				)
```
natashadecoste, line 43:
```suggestion
				.transition(this.services.transitions.getTransition("legend-mouseout-area"))
```

packages/core/src/components/graphs/area-stacked.ts
```diff
@@ -0,0 +1,186 @@
...
+				// Dispatch mouse event
+				self.services.events.dispatchEvent(Events.Area.AREA_MOUSEOVER, {
+					element: hoveredElement,
+					datum
+				});
+
+				// self.services.events.dispatchEvent("show-tooltip", {
+				// 	hoveredElement,
+				// 	type: TooltipTypes.DATAPOINT
+				// });
```
natashadecoste, line 77: keeping for later or forgot to remove?

packages/core/src/components/graphs/area-stacked.ts
```diff
@@ -0,0 +1,186 @@
...
+				// Dispatch mouse event
+				self.services.events.dispatchEvent(Events.Area.AREA_MOUSEOUT, {
+					element: hoveredElement,
+					datum
+				});
+
+				// Hide tooltip
+				// 	self.services.events.dispatchEvent("hide-tooltip", {
+				// 		hoveredElement
+				// 	});
```
natashadecoste, line 100: Keeping? or deleting?

packages/core/src/components/graphs/area-stacked.ts
```diff
@@ -0,0 +1,186 @@
...
+			// .datum(function(d) {
+			// 	console.log(d);
+			// 	this._datasetLabel = d[0].group;
+			// 	return d[0].data;
+			// })
+			.attr("role", Roles.GRAPHICS_SYMBOL)
+			.attr("aria-roledescription", "area")
+			// .attr("aria-label", d =>
+			// 	d.map(datum => datum.value || datum).join(",")
+			// )
```
natashadecoste, line 172: same as above

packages/core/src/components/graphs/area-stacked.ts
```diff
@@ -0,0 +1,186 @@
...
+						self.services.transitions.getTransition(
+							"graph_element_mouseover_fill_update"
+						)
+					)
+					.attr(
+						"fill",
+						color(hoveredElement.attr("fill"))
+							.darker(0.2)
+							.toString()
+					);
```
natashadecoste, line 66: could we collapse some of these chained commands into one liners (as above comments). Some might be auto formatted by your code editor...

for example transition is mostly a one liner in the rest of the codebase
`					.transition(self.services.transitions.getTransition("graph_element_mouseover_fill_update"))`

packages/core/src/components/graphs/area-stacked.ts
```diff
@@ -0,0 +1,186 @@
...
+			.attr("aria-roledescription", "area")
+			// .attr("aria-label", d =>
+			// 	d.map(datum => datum.value || datum).join(",")
+			// )
+			.transition(
+				this.services.transitions.getTransition(
+					"area-update-enter",
+					animate
+				)
+			)
```
natashadecoste, line 178: could be one line

packages/core/src/components/axes/two-dimensional-axes.ts
```diff
@@ -71,13 +70,18 @@ export class TwoDimensionalAxes extends Component {
 			// To be able to tell the final width & height of the axis when initiaing the transition
 			// The invisible axis is updated instantly and without a transition
 			const invisibleAxisRef = child.getInvisibleAxisRef();
-			const { width, height } = DOMUtils.getSVGElementSize(invisibleAxisRef, { useBBox: true });
+			const {
+				width,
+				height
+			} = DOMUtils.getSVGElementSize(invisibleAxisRef, { useBBox: true });
```
natashadecoste, line 76:
```suggestion
			const { width, height } = DOMUtils.getSVGElementSize(invisibleAxisRef, { useBBox: true });
```
would rather this on one line

packages/core/src/components/axes/two-dimensional-axes.ts
```diff
@@ -71,13 +70,18 @@ export class TwoDimensionalAxes extends Component {
...
 
 			let offset;
 			if (child.getTitleRef().empty()) {
 				offset = 0;
 			} else {
-				offset = DOMUtils.getSVGElementSize(child.getTitleRef(), { useBBox: true }).height;
+				offset = DOMUtils.getSVGElementSize(child.getTitleRef(), {
+					useBBox: true
+				}).height;
```
natashadecoste, line 84: same as above

theiliad:
There are conflicts again because of the other PRs we've merged

packages/core/src/components/graphs/area-stacked.ts
theiliad: code formatting related comments are somewhat irrelevant now since prettier will format them back to be that way again

sophiiae:
Screen Shot 2020-06-03 at 10 22 30 AM
Screen Shot 2020-06-03 at 10 26 50 AM

It's a bit hard to find data points on area charts. Are we going to have dot or circle on the edge similar to line chart?

theiliad:
> Screen Shot 2020-06-03 at 10 22 30 AM
> 
> Screen Shot 2020-06-03 at 10 26 50 AM
> 
> It's a bit hard to find some data points on area charts. Are we going to have dot or circle on the edge similar to line chart?

I don't yet have a design decision on that, I'll bring it up today

PR #563, opened 2020-04-09
add gauge

### LINKED ISSUE #294: Gauge Component
Requested by ADI, HQS & SCI.

**ADI:**
image

**HQS:**
image

**SCI:**
image

@shixiedesign could you give me an estimation of when we'll be exploring/implementing these?

### LINKED ISSUE #345: ADDI Immediate Needs
- Bubble chart
image

- Gauge
image

- Horizontal bar
image

lucamattiazzi:
### Updates
closes #294 
closes #345

adds a gauge as new chart, with the possibility to set the arc size

### Datasource structure

The datasource needs 3 key/value pair relative to the same group:

- current
- old
- total

The ratio between current and total is the value in the center of the gauge, the ratio between the difference old - current and current is the delta below.

### Options

The arc angle can be set in the options, using the key `arcRatio` in `gauge`. The value can have values between 0 and 1, that map to angle values of 0 and 2 PI.

Semicircular
image

Circular
image

- [ ] Demos all features
- [ ] Documented/annotated
- [ ] Matches UI/UX specs
- [ ] Accessible


caesarsol:
@lucamattiazzi Hi, we just found a bug: 
- [ ] on Firefox the numbers are not shown (`font-size` remains set to `0px`)
image

theiliad:
great progress so far 💯

theiliad:
Hi,

looks great, although a few things are breaking.

1. smaller gauges should have smaller elements.

image

2. overall it can break fairly easily with height or arcRation adjustments

image

3. is there anyway to specify which way the arrow would point?
4. is there a way to set the color of the arrow?
5. is there a way to swap the 2 numbers shown?
6. is there a way to provide a custom value for either number?
7. is there a way to hide the arrow or any of the numbers?
8. is there a way to customize the color of the gauge?
9. the arrow icon seems a bit small and transparent

lucamattiazzi:
Hi,
1&2. now the size of the text inside the gauge adapts with the radius of the gauge, also it should be a little bit bigger with higher arcRatios. 
3. the arrow points upward if the old value was lower than the current, else it points downward
4 to 8. no, not at the moment, but it can be added in the config
9. ok!

designertyler:
If the arrows point up or down depending on a positive or negative change we should use the arrows up and down with the tail to show that 

image

The caret icon is more often used as an interactive element rather than a status indicator
image

image


Or is this icon meant to represent "delta" like this example from excel? Where it's a unique icon that always "points up" 
image

packages/core/src/components/graphs/gauge.ts
```diff
@@ -86,9 +91,12 @@ export class Gauge extends Component {
 
 	getArcSize(): number {
 		const options = this.model.getOptions();
-		const { arcRatio = 1 } = options.gauge;
-		const clampedArcRatio = clamp(arcRatio, 0, 1);
-		return clampedArcRatio * Math.PI * 2;
+		const { type = "semi" } = options.gauge;
```
theiliad, line 94: this should be defaulted in `configuration.ts` to `semi`

packages/core/src/interfaces/charts.ts
```diff
@@ -174,6 +174,22 @@ export interface PieChartOptions extends BaseChartOptions {
...
+	gauge?: {
+		center?: {
+			valueFontSize?: Function;
+			deltaFontSize?: Function;
+			titleFontSize?: Function;
+			deltaYPosition?: Function;
+			valueYPosition?: Function;
+			numberFormatter?: Function;
+		};
+		type?: "semi" | "full"
```
theiliad, line 190: this should be an ENUM

caesarsol: You were anticipated by @ilariaventurini telling me the same :D

caesarsol:
@theiliad we performed the following changes:

- intriduced gauge types `semi` and `full`
- all the sizings and positionings should now feel correct
- the data format was simplified to `[ { key: "value", value: 123 }, { key: "delta", value: 45 } ]`
- the colors of `arcBackground`, `arcForeground`, `arrow` are now options the user can customize

caesarsol:
About the problem on the spacing below the title, it seems to be caused by a bug in the `legend.display: false` logic.
You can see a spacer added above the chart that should not be there. 

image

We verified the Pie chart has the same problem if the legend is hidden in the same way.

theiliad:
> About the problem on the spacing below the title, it seems to be caused by a bug in the `legend.display: false` logic.
> You can see a spacer added above the chart that should not be there.
> 
> image
> 
> We verified the Pie chart has the same problem if the legend is hidden in the same way.

ok no problem

theiliad:
@lucamattiazzi is this good to go?

caesarsol:
@theiliad yes

theiliad:
there's a number of conflicts now, we anticipated that with all the PR merges going on

PR #565, opened 2020-04-09
fix backwards compability issues with older data format

theiliad:


packages/core/src/model.ts
```diff
@@ -50,10 +50,21 @@ export class ChartModel {
...
+				if (Tools.getProperty(dataset, "label") === null) {
+					const correspondingLabel = Tools.getProperty(labels, i);
+					if (correspondingLabel) {
+						group = correspondingLabel;
+					} else {
+						group = "Ungrouped";
+					}
+				} else {
+					group = Tools.getProperty(dataset, "label");
+				}
```
zvonimirfras, line 66: Just taking the `getProperty` call out?

```suggestion
				const datasetLabel = Tools.getProperty(dataset, "label");
				if (datasetLabel === null) {
					const correspondingLabel = Tools.getProperty(labels, i);
					if (correspondingLabel) {
						group = correspondingLabel;
					} else {
						group = "Ungrouped";
					}
				} else {
					group = datasetLabel;
				}
```

theiliad:
> when using the old data format all slices seem to get the same color and it seems the legend holds just the shared dataset rather than the labels for the different slices. Tooltip also says the dataset and value so theres no way to know what the different slice labels are.
> 
> image
> 
> #### data for pie/donut
> ```
> export const pieData = {
> 	labels: ["2V2N 9KYPM version 1", "L22I P66EP L22I P66EP L22I P66EP", "JQAI 2M4L1", "J9DZ F37AP",
> 		"YEL48 Q6XK YEL48", "P66EP L22I L22I"],
> 	datasets: [
> 		{
> 			label: "Dataset 1",
> 			data: [20000, 65000, 75000, 1200, 10000, 25000]
> 		}
> 	]
> };
> ```
> 
> (using the demo data from before tabular data format was merged)

Yes unfortunately we can't do much about that, since you've got a dataset name in there. If you remove that it'll work fine

PR #567, opened 2020-04-09
Update vue version

theiliad:


theiliad:
> Looks good to me. Carbon components Vue has been on 2.6 for over one year.

@lee-chase something must be breaking with this change because the storybook site didn't build properly https://deploy-preview-567--carbon-charts-vue.netlify.com

PR #574, opened 2020-04-13
optimize axis tick formatting

theiliad:
Currently axis tick formatters are only being called for linear scales.

This PR adds support for all scale types as long as a formatter is provided by the user.

### DIFF
--- a/packages/core/src/components/axes/axis.ts
+++ b/packages/core/src/components/axes/axis.ts
@@ -152,7 +152,7 @@ export class Axis extends Component {
 			} else {
 				formatter = (t: number, i: number) => {
 					const defaultFormattedValue = formatTick(t, i, timeInterval, timeScaleOptions);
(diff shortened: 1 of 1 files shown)


packages/core/src/components/axes/axis.ts
```diff
@@ -139,20 +140,34 @@ export class Axis extends Component {
...
-			// create the right ticks formatter
-			let formatter;
-			if (isTimeScaleType) {
-				const timeInterval = computeTimeIntervalName(axis.tickValues());
+		// create the right ticks formatter
+		let formatter;
+		const userProvidedFormatter = Tools.getProperty(axisOptions, "ticks", "formatter");
+		if (isTimeScaleType) {
+			const timeInterval = computeTimeIntervalName(axis.tickValues());
+			if(userProvidedFormatter === null) {
```
natashadecoste, line 150:
```suggestion
			if (userProvidedFormatter === null) {
```

PR #576, opened 2020-04-14
fix(core): fix backwards compatibility for pie/donut

natashadecoste:
### Updates
- if the data is return in tabular format, the pie model reassigns the `group` value to `key` value 
- if the dataset name is not provided, the data is the same (`group` and `key` values are both slice labels)

##### data 

```
export const pieData = {
	labels: ["2V2N 9KYPM version 1", "L22I P66EP L22I P66EP L22I P66EP", "JQAI 2M4L1", "J9DZ F37AP",
		"YEL48 Q6XK YEL48", "P66EP L22I L22I"],
	datasets: [
		{
			label: "Dataset 1",
			data: [20000, 65000, 75000, 1200, 10000, 25000]
		}
	]
};
```

##### before
image

##### after
image

- [ ] Demos all features
- [ ] Documented/annotated
- [ ] Matches UI/UX specs
- [ ] Accessible


packages/core/src/model-pie.ts
```diff
@@ -11,6 +11,17 @@ export class PieChartModel extends ChartModel {
...
+		// update the group to the key so the slices render with the correct tooltips and colors
+		if (data !== tabularData) {
+			// If a label was set for the overall dataset, reassign it to key value
+			tabularData.forEach(d => {
+				if (d.key && d.key !== d.group) {
+					d.group = d.key;
+				}
+			});
+		}
+
```
theiliad, line 24: all of this should go into `this.getTabularData()` in `model-pie.ts`, first calling `super. getTabularData()` then applying the rest of the additions

natashadecoste: sounds good!

PR #581, opened 2020-04-17
Radar

ilariaventurini:
Core deploy: https://deploy-preview-581--carbon-charts-core.netlify.app/?path=/story/radar--radar

Design:
image

- [ ] Demos all features
- [ ] Documented/annotated
- [ ] Matches UI/UX specs
- [ ] Accessible


caesarsol:
Preview on Core: 
https://deploy-preview-581--carbon-charts-core.netlify.app/?path=/story/radar--radar

theiliad:
Great progress so far, thank you 💯

packages/core/src/services/angle-utils.ts
```diff
@@ -0,0 +1,63 @@
...
+
+function isInRange(x: number, [min, max]: [number, number]) {
+	return x >= min && x <= max;
+}
+
+export function radToDeg(rad: Angle): Angle {
+	return rad * (180 / Math.PI);
+}
+
+export function degToRad(deg: Angle): Angle {
```
theiliad, line 47: good math review 💯

packages/core/src/services/angle-utils.ts
```diff
@@ -0,0 +1,63 @@
...
+	x: number;
+	y: number;
+}
+
+export type Angle = number;
+
+interface LabelAlignment {
+	textAnchor: "start" | "middle" | "end"; // *___   __*__   ___*
+	dominantBaseline: "baseline" | "middle" | "hanging"; // __*   --*--   --.
+}
```
theiliad, line 11: 💯

packages/core/src/interfaces/events.ts
```diff
@@ -63,6 +63,16 @@ export enum Line {
 	POINT_MOUSEOUT = "scatter-mouseout"
 }
 
+/**
+ * enum of all radar graph events
+ */
+export enum Radar {
+	X_AXIS_MOUSEOVER = "radar-x-axis-mouseover",
+	X_AXIS_MOUSEMOVE = "radar-x-axis-mousemove",
+	X_AXIS_CLICK = "radar-x-axisb-click",
```
theiliad, line 72:
```suggestion
	X_AXIS_CLICK = "radar-x-axis-click",
```

packages/core/src/configuration.ts
```diff
@@ -274,6 +275,24 @@ const donutChart: DonutChartOptions = Tools.merge({}, pieChart, {
...
+		opacity: {
+			unselected: 0.1,
+			selected: 0.3
+		}
+	},
+	tooltip: {
+		gridline: {
+			enabled: true
+		},
+		valueFormatter: value => value ? value : "NA"
```
theiliad, line 292:
```suggestion
		valueFormatter: value => value ? value : "N/A"
```

packages/core/src/components/graphs/radar.ts
```diff
@@ -0,0 +1,473 @@
+// Internal Imports
+import { Component } from "../component";
+import { DOMUtils } from "../../services";
+import * as Configuration from "../../configuration";
+import { Events, TooltipTypes, Roles } from "../../interfaces";
+import { Tools } from "../../tools";
+import { flatMapDeep, kebabCase } from "lodash-es";
```
theiliad, line 7: these should go into tools

packages/core/src/components/graphs/radar.ts
```diff
@@ -0,0 +1,473 @@
...
+		}
+
+		const data: Array<Datum> = this.model.getData();
+		const displayData: Array<Datum> = this.model.getDisplayData();
+		const groupedData = this.model.getGroupedData();
+		const options = this.model.getOptions();
+		const configuration = Configuration.options.radarChart.radar;
+
+		this.groupMapsTo = options.data.groupMapsTo;
+		this.uniqKeys = Array.from(new Set(data.map(d => d.key)));
```
theiliad, line 69:
```suggestion
		this.uniqueKeys = Array.from(new Set(data.map(d => d.key)));
```

packages/core/src/components/graphs/radar.ts
```diff
@@ -0,0 +1,473 @@
...
+		this.uniqGroups = Array.from(new Set(displayData.map(d => d[this.groupMapsTo])));
```
theiliad, line 70:
```suggestion
		this.uniqueGroups = Array.from(new Set(displayData.map(d => d[this.groupMapsTo])));
```

packages/core/src/components/graphs/radar.ts
```diff
@@ -0,0 +1,473 @@
...
+		const xLabelPadding = 10;
+		const yLabelPadding = 8;
+		const labelHeight = this.labelDimensions(this.uniqKeys[0]).height;
+		const margin = 2 * (labelHeight + yLabelPadding);
+		const size = Math.min(width, height);
+		const diameter = size - margin;
+		const radius = diameter / 2;
+		const yTicksNumber = 4;
+		const minRange = 10;
+		const xAxisRectHeight = 50;
```
theiliad, line 83: most of these need to go into `configuration.ts` and the radar interface

packages/core/src/components/graphs/radar.ts
```diff
@@ -0,0 +1,473 @@
...
+		const oldRadialLineGenerator = lineRadial<Datum>()
+			.angle(radialLineGenerator.angle())
+			.radius(d => oldYScale ? oldYScale(d.value) : minRange)
+			.curve(radialLineGenerator.curve());
+
+		// compute the space that each x label needs
+		const horizSpaceNeededByEachXLabel = this.uniqKeys.map(key => {
+			const tickWidth = this.labelDimensions(key).width;
+			// compute the distance between the point that the label rapresents and the vertical diameter
+			const distPointDiam = distanceBetweenPointOnCircAndVerticalDiameter(xScale(key), radius);
```
theiliad, line 118:
```suggestion
			const distanceFromDiameter = distanceBetweenPointOnCircAndVerticalDiameter(xScale(key), radius);
```

packages/core/src/components/graphs/radar.ts
```diff
@@ -0,0 +1,473 @@
...
+		const leftPadding = max(horizSpaceNeededByEachXLabel);
+
+		// center coordinates
+		const c: Point = {
+			x: leftPadding + xLabelPadding,
+			y: height / 2
+		};
+
+		/////////////////////////////
+		// Drawing section
```
theiliad, line 131:
```suggestion
		// Drawing the radar
```

packages/core/src/components/graphs/radar.ts
```diff
@@ -0,0 +1,473 @@
...
+		/////////////////////////////
+
+		// y axes
+		const yAxes = DOMUtils.appendOrSelect(this.svg, "g.y-axes").attr("role", Roles.GROUP);
+		const yAxisUpdate = yAxes.selectAll("path").data(yTicks, tick => tick);
+		// for each tick, create array of data corrisponding to the points composing the shape
```
theiliad, line 137:
```suggestion
		// for each tick, create array of data corresponding to the points composing the shape
```

packages/core/src/components/graphs/radar.ts
```diff
@@ -0,0 +1,473 @@
...
+	// creates corrisponding data with value = 0
+	normalizeFlatData = (dataset: Array<Datum>) => {
+		const completeBlankData = flatMapDeep(this.uniqKeys.map(key => {
+			return this.uniqGroups.map(group => ({ key, [this.groupMapsTo]: group, value: null }));
+		}));
+		return Tools.merge(completeBlankData, dataset);
+	}
+
+	// Given a a grouped array of objects, if there are missing data on key,
+	// creates corrisponding data with value = 0
```
theiliad, line 361:
```suggestion
	// creates corresponding data with value = 0
```

packages/core/src/components/graphs/radar.ts
```diff
@@ -0,0 +1,473 @@
...
+		const labelHeight = this.labelDimensions(this.uniqKeys[0]).height;
+		const margin = 2 * (labelHeight + yLabelPadding);
+		const size = Math.min(width, height);
+		const diameter = size - margin;
+		const radius = diameter / 2;
+		const yTicksNumber = 4;
+		const minRange = 10;
+		const xAxisRectHeight = 50;
+
+		// given a key, return the corrisponding angle in radiants
```
theiliad, line 85:
```suggestion
		// given a key, return the corresponding angle in radiants
```

packages/core/src/components/graphs/radar.ts
```diff
@@ -0,0 +1,473 @@
...
+	labelDimensions = (label: string) => {
+		const tmpTick = DOMUtils.appendOrSelect(this.svg, `g.tmp-tick`);
+		const tmpTickText = DOMUtils.appendOrSelect(tmpTick, `text`).text(label);
+		const { width, height } = DOMUtils.getSVGElementSize(tmpTickText.node(), { useBBox: true });
+		tmpTick.remove();
+		return { width, height };
+	}
+
+	// Given a flat array of objects, if there are missing data on key,
+	// creates corrisponding data with value = 0
```
theiliad, line 352:
```suggestion
	// creates corresponding data with value = 0
```

packages/core/src/components/graphs/radar.ts
```diff
@@ -0,0 +1,473 @@
...
+		.attr("transform", key => `rotate(${radToDeg(xScale(key))}, ${c.x}, ${c.y})`);
+
+		// Add event listeners
+		this.addEventListeners();
+
+		oldYScale = yScale; // save the current scale as the old one
+	}
+
+	// append temporarily the label to get the exact space that it occupies
+	labelDimensions = (label: string) => {
```
theiliad, line 343:
```suggestion
	getLabelDimensions = (label: string) => {
```

packages/core/src/components/graphs/radar.ts
```diff
@@ -0,0 +1,473 @@
...
+		const tmpTick = DOMUtils.appendOrSelect(this.svg, `g.tmp-tick`);
+		const tmpTickText = DOMUtils.appendOrSelect(tmpTick, `text`).text(label);
+		const { width, height } = DOMUtils.getSVGElementSize(tmpTickText.node(), { useBBox: true });
+		tmpTick.remove();
+		return { width, height };
```
theiliad, line 348: you can just return `DOMUtils.getSVGElementSize(tmpTickText.node(), { useBBox: true })`

packages/core/src/components/graphs/radar.ts
```diff
@@ -0,0 +1,473 @@
...
+	}
+
+	// Given a flat array of objects, if there are missing data on key,
+	// creates corrisponding data with value = 0
+	normalizeFlatData = (dataset: Array<Datum>) => {
```
theiliad, line 353: not sure what this method does

packages/core/src/components/graphs/radar.ts
```diff
@@ -0,0 +1,473 @@
...
+
+		this.svg = this.getContainerSVG();
+		const { width, height } = DOMUtils.getSVGElementSize(this.parent, { useAttrs: true });
+		if (!width || !height) {
+			return;
+		}
+
+		const data: Array<Datum> = this.model.getData();
+		const displayData: Array<Datum> = this.model.getDisplayData();
+		const groupedData = this.model.getGroupedData();
```
theiliad, line 64: why do we need `data`, `displayData`, `groupedData`, `displayDataNormalized` and `groupedDataNormalized`?

theiliad:
love the animations!

packages/core/src/components/graphs/radar.ts
ilariaventurini: We need `data` to compute the `uniqueKeys` which must always be the same, they must not change if datasets change (so we can't use `displayData`).
`displayData` is used to show only the groups that the user chooses to see by filtering them using the legend.
`groupedData` is used to compute `groupedDataNormalized`.
`displayDataNormalized` and `groupedDataNormalized` are `displayData` and `groupedData` respectively, in which missing data is added with `value = null`. Doing so we can show always the blob, even if some data are missing.

theiliad: seems like `groupedDataNormalized` & `displayDataNormalized` should move into `model.ts` 🙂

theiliad: Ideally called `getGroupedData` can take in either `{ normalize: true }` config or we need a `getNormalizedGroupedData()`

packages/core/src/components/graphs/radar.ts
ilariaventurini: Sometimes there is missing data, for example:

```
const radarData = [
	{ group: "Sugar", key: "London", value: 25 },
	{ group: "Oil", key: "London", value: 6 },
	{ group: "Water", key: "London", value: 12 },
	{ group: "Sugar", key: "Milan", value: 13 },
	{ group: "Oil", key: "Milan", value: 6 },
	{ group: "Water", key: "Milan", value: 28 },
	{ group: "Sugar", key: "Paris", value: 19 },
	{ group: "Oil", key: "Paris", value: 16 },
	{ group: "Water", key: "Paris", value: 10 },
	{ group: "Sugar", key: "New York", value: 11 },
	{ group: "Oil", key: "New York", value: 18 },
	{ group: "Water", key: "New York", value: 8 },
	{ group: "Sugar", key: "Sydney", value: 12 },
	{ group: "Oil", key: "Sydney", value: 16 }
]
```

In that case there are 3 groups (`Sugar`, `Oil` and `Water`) and 5 keys (`London`, `Milan`, `Paris`, `New York` and `Sydney`) but, as you can see, the datum `{ group: "Water", key: "Sydney", value: xx }` is missing. 
If I compute the path for `Water` blob, I will have a shape composed by 4 vertices and not 5 and it will be a bit strange. So I need to create the missing data and fill them with `value = null`.
`normalizeFlatData` and `normalizeGroupedData` do just that: they fill `data` and `displayData` with missing data, setting `value = null`.

I hope it's clearer now :)

packages/core/src/components/graphs/radar.ts
ilariaventurini: Mmmm I don't think so :) I need to calculate and save` { width, height }` before removing the node

packages/core/src/components/graphs/radar.ts
caesarsol: It was a design decision to show missing data as `0`, with an `N/A` displayed in the tooltip! Otherwise, "jumping" a datapoint in a radar makes incorrect values being read.

packages/core/src/components/graphs/radar.ts
ilariaventurini: We tried to make this more generic, but we obtained a function with 4 parameters. Being that it's maybe useless for other charts, we would not refactor it into `model.ts` to avoid creating unnecessary complexity

ilariaventurini:
At the moment the mapping of the column name of the dataset to the angle and value is hardcoded in the code.
I will add a configuration object in radar configuration to make the graph more generic.
Then I will also add some other examples.

packages/core/src/services/angle-utils.ts
```diff
@@ -0,0 +1,63 @@
+export interface Point {
+	x: number;
+	y: number;
+}
+
+export type Angle = number;
+
+interface LabelAlignment {
+	textAnchor: "start" | "middle" | "end"; // *___   __*__   ___*
+	dominantBaseline: "baseline" | "middle" | "hanging"; // __*   --*--   --.
```
theiliad, line 10: these should be ENUMs

packages/core/src/components/graphs/radar.ts
```diff
@@ -0,0 +1,463 @@
...
+				const hoveredElement = select(this);
+				const axisLine = self.parent.select(`.x-axes .x-axis-${Tools.kebabCase(datum)}`);
+				const dots = self.parent.selectAll(`.dots circle.${Tools.kebabCase(datum)}`);
+
+				// Change style
+				axisLine.classed("hovered", true)
+					.attr("stroke-dasharray", "4 4");
+				dots.classed("hovered", true)
+					.attr("opacity", 1)
+					.attr("r", 5);
```
natashadecoste, line 415: I think we should have this in the configuration and set default to 5

packages/core/src/components/graphs/radar.ts
```diff
@@ -0,0 +1,463 @@
...
+			.on("mouseout", null);
+		// Remove legend listeners
+		const eventsFragment = this.services.events;
+		eventsFragment.removeEventListener(Events.Legend.ITEM_HOVER, this.handleLegendOnHover);
+		eventsFragment.removeEventListener(Events.Legend.ITEM_MOUSEOUT, this.handleLegendMouseOut);
+	}
+
+	addEventListeners() {
+		const self = this;
+		const { angle, value } = this.model.getOptions().radar.axes;
```
natashadecoste, line 394:
```suggestion
		const { angle, value } = Tools.getProperty(this.model.getOptions(), "radar", "axes");
```

also it seems `value` isnt being used

packages/core/src/components/graphs/radar.ts
```diff
@@ -0,0 +1,463 @@
...
+			return this.uniqueGroups.map(group => ({ [angle]: key, [groupMapsTo]: group, [value]: null }));
+		}));
+		return Tools.merge(completeBlankData, dataset);
+	}
+
+	// Given a a grouped array of objects, if there are missing data on key,
+	// creates corresponding data with value = null
+	normalizeGroupedData = (dataset: any) => {
+		const options = this.model.getOptions();
+		const { angle, value } = options.radar.axes;
```
natashadecoste, line 352:
```suggestion
		const { angle, value } =  Tools.getProperty(options, "radar", "axes");
```

packages/core/src/components/graphs/radar.ts
```diff
@@ -0,0 +1,463 @@
...
+		this.svg = this.getContainerSVG();
+		const { width, height } = DOMUtils.getSVGElementSize(this.parent, { useAttrs: true });
+
+		const data = this.model.getData();
+		const displayData = this.model.getDisplayData();
+		const groupedData = this.model.getGroupedData();
+		const options = this.model.getOptions();
+		const configuration = Configuration.options.radarChart.radar;
+		const groupMapsTo = options.data.groupMapsTo;
+		const { angle, value } = options.radar.axes;
```
natashadecoste, line 52:
```suggestion
		const configuration = Tools.getProperty(Configuration, "options", "radarChart", "radar");
		const groupMapsTo = Tools.getProperty(options, "data", "groupMapsTo");
		const { angle, value } = Tools.getProperty(options, "radar", "axes");
```
most of the codebase uses the helper function for accesing properties especially when the property paths are longer

packages/core/src/components/graphs/radar.ts
```diff
@@ -0,0 +1,463 @@
...
+		tmpTick.remove();
+		return { width, height };
+	}
+
+	// Given a flat array of objects, if there are missing data on key,
+	// creates corresponding data with value = null
+	normalizeFlatData = (dataset: any) => {
+		const options = this.model.getOptions();
+		const { angle, value } = options.radar.axes;
+		const groupMapsTo = options.data.groupMapsTo;
```
natashadecoste, line 341:
```suggestion
		const { angle, value } = Tools.getProperty(options, "radar", "axes");
		const groupMapsTo = Tools.getProperty(options, "data", "groupMapsTo");
```

packages/core/src/components/graphs/radar.ts
```diff
@@ -0,0 +1,463 @@
...
+		}));
+		return Tools.merge(completeBlankData, dataset);
+	}
+
+	// Given a a grouped array of objects, if there are missing data on key,
+	// creates corresponding data with value = null
+	normalizeGroupedData = (dataset: any) => {
+		const options = this.model.getOptions();
+		const { angle, value } = options.radar.axes;
+		const groupMapsTo = options.data.groupMapsTo;
```
natashadecoste, line 353:
```suggestion
		const groupMapsTo = Tools.getProperty(options, "data", "groupMapsTo");
```

packages/core/src/components/graphs/radar.ts
```diff
@@ -0,0 +1,463 @@
...
+			.style("fill-opacity", group => {
+				if (group.name !== hoveredElement.datum().name) {
+					return opacity.unselected;
+				}
+				return opacity.selected;
+			});
+	}
+
+	handleLegendMouseOut = (event: CustomEvent) => {
+		const { opacity } = Configuration.options.radarChart.radar;
```
natashadecoste, line 374:
```suggestion
		const { opacity } = Tools.getProperty(Configuration, "options", "radarChart", "radar");
```

packages/core/src/components/graphs/radar.ts
```diff
@@ -0,0 +1,463 @@
...
+		const groupMapsTo = options.data.groupMapsTo;
+		return dataset.map(({ name, data }) => {
+			const completeBlankData = this.uniqueKeys.map(k => ({ [groupMapsTo]: name, [angle]: k, [value]: null }));
+			return { name, data: Tools.merge(completeBlankData, data) };
+		});
+	}
+
+	handleLegendOnHover = (event: CustomEvent) => {
+		const { hoveredElement } = event.detail;
+		const { opacity } = Configuration.options.radarChart.radar;
```
natashadecoste, line 362:
```suggestion
		const { opacity } = Tools.getProperty(Configuration, "options", "radarChart", "radar");
```

theiliad:
only issue I find is when height is set to 100 the chart disappears, which I don't think is too much of a bug, at least for now

image

theiliad:
I think wrappers are missing

ilariaventurini:
@theiliad You are right, in that case there is not enough space to draw the radar. I think that radius is negative so the component doesn't render anything because of `if (radius <= 0) { return; }` at line 64.

If you don't precede me, I will make wrappers on Monday.

PR #588, opened 2020-04-22
Fix `Invalid time value` in Firefox

ilariaventurini:
This PR solves the **_RangeError: Invalid time value_** problem that you can see in this [codesandox](https://codesandbox.io/s/carbon-charts-react-x129h) if you open it in Firefox.

This error is thrown in Firefox when passing an invalid string to `new Date`, Chrome instead does not generate errors.
For example:

- Firefox: `new Date("Fri Dec 11 2020 00:00:45 GMT+0100 (Central European Standard Time)9000")` --> `Invalid date`
- Chrome: `new Date("Fri Dec 11 2020 00:00:45 GMT+0100 (Central European Standard Time)9000")` --> valid date: `Fri Dec 11 2020 00:00:45 GMT+0100 (Central European Standard Time)`.

This happens when you want to extend the domain calling the `extendsContinuousDomain` function with dates as domain.
However, this should not happen because if `ScaleType` is `TIME`, the domain is extended with the functions of `date-fns` `subYears`/`addYears`, `subMonths`/`addMonths`, etc ...
In the codesandox example, the user set `spaceToAddToEdges = 0` so it did not enter in this `if (spaceToAddToEdges) {...} `and ran the code below: the `addPaddingInDomain` function renamed as `extendsContinuousDomain`.

I solved the problem making also a mini refactor.

Here are two examples of linechart with time series data: the first one with `spaceToAddToEdges = 1 `(default value) and the second one with `spaceToAddToEdges = 0`.

Screenshot 2020-04-22 at 16 33 34
Screenshot 2020-04-22 at 16 33 16

- [ ] Demos all features
- [ ] Documented/annotated
- [ ] Matches UI/UX specs
- [ ] Accessible


packages/core/src/services/scales-cartesian.ts
```diff
@@ -383,3 +330,54 @@ export class CartesianScales extends Service {
 		return scale;
 	}
 }
+
+function extendsTimeDomain(domain: any, spaceToAddToEdges: number) {
```
theiliad, line 334:
```suggestion
function addSpacingToTimeDomain(domain: any, spaceToAddToEdges: number) {
```

packages/core/src/services/scales-cartesian.ts
```diff
@@ -383,3 +330,54 @@ export class CartesianScales extends Service {
...
+	}
+
+	if (differenceInSeconds(endDate, startDate) > 1) {
+		return [subSeconds(startDate, spaceToAddToEdges), addSeconds(endDate, spaceToAddToEdges)];
+	}
+
+	return [startDate, endDate];
+}
+
+function extendsContinuousDomain([lower, upper]: number[], paddingRatio: number) {
```
theiliad, line 373:
```suggestion
function addSpacingToContinuousDomain([lower, upper]: number[], paddingRatio: number) {
```

ilariaventurini:
@theiliad Do you know what's causing this problem?
image

theiliad:
> @theiliad Do you know what's causing this problem?
> image

ignore that, that's the deploy preview for the svelte package which hasn't been merged yet

caesarsol:
> I was curious to see what would solve the Firefox issue, wasn't able to pinpoint that but looks good 😄

Seems to be just matters of incorrect dates parsing!

theiliad:
> > I was curious to see what would solve the Firefox issue, wasn't able to pinpoint that but looks good 😄
> 
> Seems to be just matters of incorrect dates parsing!

which lines?

ilariaventurini:
> > > I was curious to see what would solve the Firefox issue, wasn't able to pinpoint that but looks good 😄
> > 
> > 
> > Seems to be just matters of incorrect dates parsing!
> 
> which lines?

The problem was that we was summing a date to a number so the result is a string like that one `"Fri Dec 11 2020 00:00:45 GMT+0100 (Central European Standard Time)9000"`.
If you pass this string to a `new Date` in Chorme, it ignores the number in the end and it returns the Date Object corrisponding to 12/11/2020, however, Firefox returns the `Invalid date error`.

PR #598, opened 2020-04-28
Ruler fix

lucafalasco:
### Updates
fix elements higlighting 

Deploy: https://deploy-preview-598--carbon-charts-core.netlify.app

### DIFF
--- a/packages/core/src/components/axes/ruler.ts
+++ b/packages/core/src/components/axes/ruler.ts
@@ -48,24 +48,29 @@ export class Ruler extends Component {
 		 * to only get elements which belong to the same axis coordinate
 		 */
 		const dataPointsMatchingRulerLine: {domainValue: number, originalData: any}[] =
-			scaledData.reduce((accum, currentValue) => {
-				// store the first element of the accumulator array to compare it with current element being processed
-				const sampleAccumValue = accum[0] ? accum[0].domainValue : undefined;
-
-				// if accumulator is not empty and current value is bigger than already existing value in the accumulator, skip current iteration
-				if (sampleAccumValue !== undefined && currentValue.domainValue > sampleAccumValue) {
+			scaledData
+			.filter(d => pointIsWithinThreshold(d.domainValue, x))
+			.reduce((accum, currentValue) => {
+				if (accum.length === 0) {
+					accum.push(currentValue);
 					return accum;
 				}
 
-				// there's a match and currentValue is either less then or equal to already stored values
-				if (pointIsWithinThreshold(currentValue.domainValue, x)) {
-					if (sampleAccumValue !== undefined && currentValue.domainValue < sampleAccumValue) {
-						// there's a closer data point in the threshold area, so reinstantiate array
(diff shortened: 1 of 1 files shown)


packages/core/src/components/axes/ruler.ts
```diff
@@ -50,12 +50,11 @@ export class Ruler extends Component {
 			"[role=graphics-symbol]"
 		);
 		const displayData = this.model.getDisplayData();
-		const domainScale = this.services.cartesianScales.getDomainScale();
 		const rangeScale = this.services.cartesianScales.getRangeScale();
 		const [yScaleEnd, yScaleStart] = rangeScale.range();
 
-		const scaledData: {domainValue: number, originalData: any}[] = displayData.map((d, i) => ({
-			domainValue: this.services.cartesianScales.getDomainValue(d, i),
+		const scaledData: {domainValue: number, originalData: any}[] = displayData.map(d => ({
```
theiliad, line 56: why do we need `scaledData`?
The entire logic seems a bit extensive to me.

Maybe I'm missing something but I think we should just use:
```js
const domainIdentifier = this.services.cartesianScales.getDomainIdentifier();
```

and go through `displayData` filtering out datapoints where `d[domainIdentifier]` matches our domain value that the ruler was hovering through

lucafalasco: we need to call `this.services.cartesianScales.getDomainValue()` on data to compare it with current mouse x position and filter data accordingly, `scaledData` is used basically to avoid having to call getDomainValue multiple times in data and elements manipulations/filtering.
By the way I just pushed a refactor for the reducer, adding an initial filter to only go through points which are within threshold, and also optimized the nearest point matching calcs.

caesarsol: Just an additional note: we need to reason in terms of pixels and not in terms of "domain units" because 1 pixel is the maximum resolution we can have. (i.e. two points on the same x coordinate, even if they are shifted by 0.001px, should be matched together) 
If we reason in terms of domain units, we would lose that possibility. 

And being the threshold calculated in px, we couldn't apply that either.

theiliad: ok, the refactor looks ok for now, let's merge than then I'll take another stab at it later.

I just found a few issues though

packages/core/src/components/axes/ruler.ts
```diff
@@ -64,77 +48,91 @@ export class Ruler extends Component {
...
 
 			ruler.attr("opacity", 1);
 
 			// line snaps to matching point
 			const sampleMatch = dataPointsMatchingRulerLine[0];
-			line.attr("y1", yScaleStart)
-				.attr("y2", yScaleEnd)
-				.attr("x1", sampleMatch.domainValue)
-				.attr("x2", sampleMatch.domainValue);
+			if (orientation === "horizontal") {
```
theiliad, line 123: CartesianOrientations.HORIZONTAL

PR #614, opened 2020-05-14
detect correct main vertical and horizontal axis positions

theiliad:
Just had to optimize the logic there a bit


packages/core/src/services/scales-cartesian.ts
```diff
@@ -214,10 +214,14 @@ export class CartesianScales extends Service {
 
 	protected findMainVerticalAxisPosition() {
 		const options = this.model.getOptions();
-		const axisOptions = Tools.getProperty(options, "axes");
+		const axesOptions = Tools.getProperty(options, "axes");
 
 		// If right axis has been specified as `main`
-		if (Tools.getProperty(axisOptions, AxisPositions.RIGHT, "main") === true) {
+		if (
+			(Tools.getProperty(axesOptions, AxisPositions.LEFT) === null &&
```
zvonimirfras, line 221: Indentation makes it looks like the grouping is opposite from what it is.
Also braces around the && are not actually needed.

theiliad: could you write the if-statement as you're imagining it please? I didn't really understand your comment there. Also we've got #617 which adds prettier

zvonimirfras: I really dislike prettier and don't think we should add it.

zvonimirfras: There are so many reasonable ways to do it, feel free to pick one that suggests the right operator priority.

theiliad: the point of prettier is to get rid of our different opinions

packages/core/src/services/scales-cartesian.ts
```diff
@@ -226,10 +230,14 @@ export class CartesianScales extends Service {
...
-		const axisOptions = Tools.getProperty(options, "axes");
+		const axesOptions = Tools.getProperty(options, "axes");
 
 		// If top axis has been specified as `main`
-		if (Tools.getProperty(axisOptions, AxisPositions.TOP, "main") === true) {
+		if (
+			(Tools.getProperty(axesOptions, AxisPositions.BOTTOM) === null &&
+				Tools.getProperty(axesOptions, AxisPositions.TOP) !== null) ||
+				Tools.getProperty(axesOptions, AxisPositions.TOP, "main") === true
+		) {
```
zvonimirfras, line 240: For maximum verbosity, you could go with
```suggestion
		if ((
				Tools.getProperty(axesOptions, AxisPositions.BOTTOM) === null
				&& Tools.getProperty(axesOptions, AxisPositions.TOP) !== null
			)
			|| Tools.getProperty(axesOptions, AxisPositions.TOP, "main") === true
		) {
```
or something along those lines

PR #617, opened 2020-05-15
add prettier

theiliad:


packages/core/.storybook/main.js
```diff
@@ -14,13 +12,13 @@ module.exports = {
...
-							includePaths: [path.resolve(__dirname + "../src")]
-						}
-					}
-				}
-			]
+							includePaths: [path.resolve(__dirname + "../src")],
+						},
+					},
+				},
+			],
```
cal-smith, line 19: Ew, trailing commas

theiliad: I think those can disabled through prettier configs.

IMO we shouldn't care about any of this stuff, they're all arbitrary preferences that are another reason why Prettier exists

PR #639, opened 2020-05-25
Show pre-selected data groups

### LINKED ISSUE #610: Need ability to set which datasets are selected on initial chart display (rather than all)
#### I am submitting a...

- [x] Feature request
- [ ] Design defect
- [ ] Source code defect
- [ ] Other

### Issue description
Currently it appears that **ALL** datasets in the legend are selected when the chart initially displays.
**Feature request:**
I’d like to be able to set which datasets ( probably via some new property in the ‘options’ ) are selected when the chart initially displays.
**Why is this useful ?**
Adding this feature/option would enable support for restoring/displaying a chart with a specific set of datasets turned-on & turned-off.
**Example:**
See  http://peretz1.fyre.ibm.com:3040/edt/covid_chart.html
Suppose you’re only interested in info about (say) Italy … so in report 1 ( in the left-nav) you select “Italy” in the legend.
And then you go to report 2 … but the chart legend has reset to show all countries (sigh), requiring the user to have to click “Italy” again.  Not optimal.
By adding this feature report 2 could easily be displayed with the legend settings carried over from report 1.
Plus it’s not hard to imagine other use cases where this feature would be useful:
Bookmarking a chart with a specific set of datasets selected.
Dashboard of charts where legend selections in one chart could also be used to drive the display of the other charts.

See example above

sophiiae:
closes #610


packages/core/src/configuration.ts
```diff
@@ -153,8 +153,9 @@ const chart: BaseChartOptions = {
 		loading: false,
 	},
 	color: {
-		scale: null,
+		scale: null
 	},
+	selectedGroups: []
```
theiliad, line 158: should go into the `data` field within options

packages/core/src/components/essentials/legend.ts
```diff
@@ -268,6 +268,7 @@ export class Legend extends Component {
 				const clickedItemData = clickedItem.datum() as any;
 
 				self.model.toggleDataLabel(clickedItemData.name);
+				self.model.updateSelectedGroups();
```
theiliad, line 271: not sure how this needs to happen, and I'm not sure how realistic it is for the dev to pass in `selectedGroups`. what if the chart updates? or the user interacts with the legend?

sophiiae: > not sure how this needs to happen, and I'm not sure how realistic it is for the dev to pass in `selectedGroups`. what if the chart updates? or the user interacts with the legend?

When user check the legend, this will add checked legendItems to selectedGroup, if no legend is checked which means the chart will show all the data groups, then selectedGroup will be updated to an empty array since no legend is pre-selected. I think maybe pre-selectedGroup will be more decent for this option. What do you think ?

And chart updates you mean the data has changed ?

packages/core/src/configuration.ts
sophiiae: Ok, I'll revise that.

packages/core/src/components/essentials/legend.ts
theiliad: yes I'm concerned with any updates really

packages/core/src/model.ts
```diff
@@ -288,6 +288,39 @@ export class ChartModel {
...
+		if (hasDeactivatedItems) {
+			this.state.options.data.selectedGroups = activeItems.map(activeItem => activeItem.name);
+		} else {
+			// If every item is active, clear array
+			this.state.options.data.selectedGroups = [];
+		};
+	}
+
+	checkSelectedGroups(uniqueDataGroups: string[]) {
+		if (this.state.options.data.selectedGroups.length) {
```
theiliad, line 313: options should be grabbed from `this.getOptions()`

packages/core/src/components/essentials/legend.ts
theiliad, line 271: since `updateSelectedGroups` is called every time `toggleDataLabel` is called, would it make sense to call `updateSelectedGroups` inside `toggleDataLabel` instead?

sophiiae: I tried that, but seems the `selectedGroups` only shows former selected legends inside `toggleDataLabel` (no current clicked legend).

theiliad:
That can’t be correct, you can see it being passed into toggleDatalabel
above

theiliad:
@sophiiae could you please add a few test cases for this PR? this seems like something that could easily break down the road

packages/core/src/selectedGroups.spec.ts
```diff
@@ -0,0 +1,47 @@
...
+		const testEnvironment = new TestEnvironment();
+		testEnvironment.render();
+
+		this.chart = testEnvironment.getChartReference();
+		this.testEnvironment = testEnvironment;
+	});
+
+	describe("set default selectedGroups", () => {
+		it("should match the array length in options", function (done) {
+			const numberOfSelectedGroup = this.chart.model.getOptions().data.selectedGroups.length;
```
theiliad, line 22: there's no point in comparing with a `selectedGroups` array that's gone into the library. You'd need to compare it to what you actually provide in options, because that way we can actually detect if at some point we incorrectly modify `selectedGroups` in model

theiliad: I suggest storing the array you include in your options in a variable

sophiiae: Ok, got it.

theiliad:
also, in this demo once you click a checkbox the chart moves up a bit
image

natashadecoste:
when the graph moves up (as per @theiliad comment above) we can see this weird render of "Dataset 1"
image

it seems to be another checkmark path 
image

and on chart resize / rerender the DOM fills up with this misplaced checkmark 
image

theiliad:
> when the graph moves up (as per @theiliad comment above) we can see this weird render of "Dataset 1"
> image
> 
> it seems to be another checkmark path
> image
> 
> and on chart resize / rerender the DOM fills up with this misplaced checkmark
> image

@natashadecoste where in the DOM is this? I can't find a `g.check`

natashadecoste:
@theiliad 
image

look inside any of the `g.legend-item.active` 
line demo: https://deploy-preview-639--carbon-charts-react.netlify.app/?path=/story/line--line-selected-groups

packages/core/src/components/essentials/legend.ts
```diff
@@ -215,9 +215,20 @@ export class Legend extends Component {
 
 				lastYPosition = yPosition;
 
+				// Test if legendItems are placed in the correct direction
+				const testHorizontal = (!legendOrientation ||
```
theiliad, line 219: why do we need an extra check? is there any way to make this work without the extra code?

it seems to be working fine right now

sophiiae: Because the legend has been called twice. `hasDeactivatedItems` is initially true with `selectedGroups` and the check marks aligned vertically in the first render.  And second render won't run since it's `g.check` is not empty.

packages/core/src/components/layout/layout.ts
```diff
@@ -158,17 +158,18 @@ export class LayoutComponent extends Component {
...
 
 			if (d.data.id === "legend") {
-				matchingSVGDimensions = DOMUtils.getSVGElementSize(
+				const svgSize = DOMUtils.getSVGElementSize(
 					select(this),
 					{ useAttrs: true }
 				);
-			}
+				matchingSVGDimensions.height = svgSize.height;
+			};
```
theiliad, line 172:
```suggestion
			}
```

Issue #612, opened 2020-05-13
Charts get wrecked by long labels 


hlyang397:
#### I am submitting a...

- [ ] Feature request
- [ ] Design defect
- [X] Source code defect
- [ ] Demo/documentation defect
- [ ] Other

### Issue description
Charts get wrecked with very long labels, and chart size is limited.

#### Steps to produce the issue
Set long labels in a chart with limited chart size.
ex:
```
export const simpleBarData = [
	{ group: "ec2825604d15b908e98d92defaefcd6600308baf498c435029d10fd9dae551c0", value: 65000 },
	{ group: "8c87b477fc360c3790e4bcff559d393327a2a1809041424aad5d3d025c54cdfc", value: 29123 },
	{ group: "b1ae877ce7cd4e8a5fbd1615b1bd1780057c0774d0cb26adafadabde66e33fb0", value: 35213 },
	{ group: "09e08d54e93d492f0581afe468e3def9ac5af04e5d60f7e88b0212686c523b8b", value: 51213 },
	{ group: "d9b276d3326823b34407f378185e833a3b82f38562d32c6ad046820c5afacbda", value: 16932 }
];
```
#### Current behavior
![image](https://user-images.githubusercontent.com/59426533/81768793-7bf5ab80-950e-11ea-88a1-29c243f475dc.png)

#### Expected behavior
Labels should be truncated, and full label value in tooltip will display when mouse hover.
https://www.carbondesignsystem.com/patterns/overflow-content/#variations
![image](https://user-images.githubusercontent.com/59426533/81769208-97ad8180-950f-11ea-8409-83578cebdb8a.png)

PR #646, opened 2020-06-01
truncate label if too long to avoid graph display issue

linhenry0417:
### Updates
- list
- out
- updates
Issue link: https://github.com/carbon-design-system/carbon-charts/issues/612

- [ ] Demos all features
- [ ] Documented/annotated
- [ ] Matches UI/UX specs
- [ ] Accessible


linhenry0417:
@theiliad 
1. Yes, the tooltip onHover function is available.
2. Demo added to storybook.

theiliad:
> @theiliad
> 
> 1. Yes, the tooltip onHover function is available.
> 2. Demo added to storybook.

Hi, which demo has it been added to? this is what I'm seeing in the area story
image

linhenry0417:
The demo were added to Bar (vertical) and Bar (horizontal) with the name Simple Bar with long label. In my storybook the Area story looks fine though.

linhenry0417:
Screen Shot 2020-06-04 at 12 48 41 AM
I am able to see the full text when hovering in my storybook. I wonder why you couldn't see it.

theiliad:
> Screen Shot 2020-06-04 at 12 48 41 AM
> 
> I am able to see the full text when hovering in my storybook. I wonder why you couldn't see it. I added <title> tag to achieve this. My browser is Firefox.

oh I see. we'd need the charting tooltip to show up rather than the browser `<title>` behavior. I don't think `<title>` shows up on the phone either

theiliad:
Here's an example within the codebase which shows a tooltip onHover for a truncated chart title https://github.com/carbon-design-system/carbon-charts/blob/d97ebe67ab8f9bfe1aef33b92388cac3017c8c23/packages/core/src/components/essentials/title.ts#L26-L67

linhenry0417:
OK. And the user should be able to choose which type of truncation he wants, right? Then I guess we need to add more options to axisOptions, right?

theiliad:
> OK. And the user should be able to choose which type of truncation he wants, right? Then I guess we need to add more options to axisOptions, right?

yes, I'm thinking maybe we can standardize the truncation options.

e.g.
```js
{
 axes: {
   left: {
     truncation: {
        enabled: true,
        type: TruncationTypes.END_LINE
     }
   }
 }
}
```

this way we can use it anywhere:

```js
{
   legend: {
     truncation: {
        enabled: true,
        type: TruncationTypes.END_LINE
     }
   }
}
```

theiliad:
or we can remove enabled and do this:
```js
{
   legend: {
     truncation: {
        type: TruncationTypes.END_LINE
     }
   }
}
```

```js
{
   legend: {
     truncation: {
        type: TruncationTypes.NONE
     }
   }
}
```

packages/core/src/components/axes/axis.ts
```diff
@@ -416,13 +416,15 @@ export class Axis extends Component {
 		// truncate the label if it's too long
 		// only applies to discrete type
 		if (!isTimeScaleType && axisOptions.scaleType) {
-			const lable_data = this.model.getDataGroups();
-			if (lable_data.length > 0) {
+			const labels = this.model.getDataGroups();
```
theiliad, line 419:
```suggestion
			const dataGroups = this.model.getDataGroups();
```

packages/core/src/components/axes/axis.ts
```diff
@@ -416,13 +416,15 @@ export class Axis extends Component {
...
+			if (labels.length > 0) {
 				let label_data_array = [];
-				const first_data = lable_data[0].name;
-				label_data_array.push(first_data);
-				const data_array = lable_data.map(d => d.name);
-				label_data_array = label_data_array.concat(data_array);
+				const first_label = labels[0].name;
```
theiliad, line 422:
```suggestion
				const firstDataGroupName = dataGroups[0].name;
```

packages/core/src/components/axes/axis.ts
```diff
@@ -416,13 +416,15 @@ export class Axis extends Component {
...
+				const label_array = labels.filter(d => d.status === 1).map(d => d.name);
```
theiliad, line 423:
```suggestion
				const activeDataGroups = this.model.getActiveDataGroups();
```

packages/core/src/components/axes/axis.ts
```diff
@@ -413,6 +413,42 @@ export class Axis extends Component {
...
+				container
+					.selectAll("g.ticks g.tick")
+					.html(tick_html);
+				container
+					.selectAll("g.tick text")
+					.data(label_data_array)
+					.text(function(d) {
+						return d.length > 18 ? d.substr(0, 8) + "..." + d.substr(-8) : d;
+					})
+					.append("title")
```
theiliad, line 440: this needs to be switched out to show a chart tooltip onHover

packages/core/src/components/axes/axis.ts
```diff
@@ -413,6 +413,42 @@ export class Axis extends Component {
...
+					`g.axis.${axisPosition} g.ticks g.tick`
+				).html();
+				container
+					.selectAll("g.ticks g.tick")
+					.html(tick_html);
+				container
+					.selectAll("g.tick text")
+					.data(label_data_array)
+					.text(function(d) {
+						return d.length > 18 ? d.substr(0, 8) + "..." + d.substr(-8) : d;
```
theiliad, line 438: values like `18` or `8` need to come from options

packages/core/src/components/essentials/legend.ts
```diff
@@ -51,8 +51,10 @@ export class Legend extends Component {
 		addedLegendItems
 			.append("text")
 			.merge(legendItems.select("text"))
-			.html((d) => d.name)
-			.attr("alignment-baseline", "middle");
+			.html((d) => d.name.length > 18 ? d.name.substr(0, 8) + "..." + d.name.substr(-8) : d.name)
+			.attr("alignment-baseline", "middle")
+			.append("title")
```
theiliad, line 56: this needs to be switched out to show a chart tooltip

packages/core/demo/data/bar.ts
```diff
@@ -92,8 +92,20 @@ export const simpleBarLongLabelOptions = {
 		bottom: {
 			mapsTo: "group",
 			scaleType: "labels",
+			"truncation": {
+				"type": "TruncationTypes.END_LINE",
```
theiliad, line 96:
```suggestion
				"type": TruncationTypes.END_LINE,
```

theiliad: These need to be turned into ENUMs, see this line please:
https://github.com/carbon-design-system/carbon-charts/blob/master/packages/core/src/interfaces/enums.ts#L17

packages/core/src/components/axes/axis.ts
```diff
@@ -415,7 +430,7 @@ export class Axis extends Component {
 
 		// truncate the label if it's too long
 		// only applies to discrete type
-		if (!isTimeScaleType && axisOptions.scaleType) {
+		if (truncationTypeProvided && truncationTypeProvided !== "TruncationTypes.NONE" && !isTimeScaleType && axisOptions.scaleType) {
```
theiliad, line 433: truncationTypeProvided isn't necessary.

we can set truncation by default to `TruncationTypes.NONE` or any other types (I don't suggest defaulting to none, maybe END_LINE by default?

then we also don't need to check whether it's been provided or not

packages/core/src/components/axes/axis.ts
```diff
@@ -435,13 +450,27 @@ export class Axis extends Component {
 					.selectAll("g.tick text")
 					.data(label_data_array)
 					.text(function(d) {
-						return d.length > 18 ? d.substr(0, 8) + "..." + d.substr(-8) : d;
+						if (truncationTypeProvided === "TruncationTypes.MID_LINE") {
```
theiliad, line 453: this entire piece should go into `Tools.ts`, a function that'll take in text + a truncation type, and return truncated text

packages/core/src/components/axes/axis.ts
```diff
@@ -435,13 +450,27 @@ export class Axis extends Component {
...
+						} else if (truncationTypeProvided === "TruncationTypes.END_LINE") {
+							return d.length > truncationThresholdProvided ? d.substr(0, truncationNumCharacterProvided) + "..." : d;
+						}
 					});
 				this.getInvisibleAxisRef()
 					.selectAll("g.tick text")
 					.data(activeDataGroups)
 					.text(function(d) {
-						return d.length > 18 ? d.substr(0, 8) + "..." + d.substr(-8) : d;
+						if (truncationTypeProvided === "TruncationTypes.MID_LINE") {
```
theiliad, line 466: same as above, `Tools.ts`

packages/core/src/interfaces/components.ts
```diff
@@ -49,6 +49,12 @@ export interface LegendOptions {
 		radius?: Number;
 		spaceAfter?: Number;
 	};
+	truncation?: {
```
theiliad, line 52: `truncation` should move out into it's own interface, because ideally we'd want everything to be able to be truncated using the similar configs. once we have an interface for `truncation` we can add the same options into axes for instance

linhenry0417:
Thanks for all the comments @theiliad. I've checked the storybook and the truncation seemed to work fine to me. Feel free to modify the code in the branch if that's easier/faster to you.

theiliad:
> Thanks for all the comments @theiliad. I've checked the storybook and the truncation seemed to work fine to me. Feel free to modify the code in the branch if that'd be easier/faster for you.

thank you, something definitely is broken though
image

linhenry0417:
My storybook runs OK though. Not sure if it was caused by merging master branch.

theiliad:
> My storybook runs OK though. Not sure if it was caused by merging master branch.
> I just updated my branch with latest master branch and re-ran the storybook. It looks fine. Maybe try to re-pull the branch once again? Please let me know if the problem still exists. Thanks.

I haven't pulled in any branches locally, I'm looking at the Netlify deploy previews. Storybook can break for various reasons. I'm suspecting that the dev sever is working but when storybook builds it errors out

theiliad:
@linhenry0417 let me know if you need help on figuring out why storybook won't build

linhenry0417:
@theiliad I thought you were trying to fix the dev server. My storybook runs fine locally with this branch so I'm not sure how to reproduce the problem. How do I fix it?

theiliad:
Again as mentioned above, you’re only running the storybook dev server, but
it seems like this is happening in the final storybook build (the script
can be found in the core package.json)

You’d need to run that script and test what is breaking

linhenry0417:
@theiliad Oh I think I know what you mean now. Now I can run yarn build without error. I've pushed a commit and hopefully this would fix the problem. Could you help to try again? Thanks.

packages/core/src/components/axes/axis.ts
```diff
@@ -42,6 +42,30 @@ export class Axis extends Component {
...
+		);
+		// default config for truncation
+		// set type to END_LINE
+		// set threshold to 12
+		// set number of characters to show to 14
+		if (!truncationTypeProvided) {
+			truncationTypeProvided = TruncationTypes.END_LINE;
+			truncationThresholdProvided = 16;
+			truncationNumCharacterProvided = 14;
+		}
```
natashadecoste, line 68: lets just set these as default configs in `configuration.ts` ?

packages/core/src/components/axes/axis.ts
```diff
@@ -414,6 +438,46 @@ export class Axis extends Component {
...
+				const firstDataGroupName = dataGroups[0].sharedStackKey;
+				const activeDataGroups = dataGroups.map(d => d.sharedStackKey);
+				if (activeDataGroups.length === dataGroups.length) {
+					label_data_array.push(firstDataGroupName);
+				}
+				label_data_array = label_data_array.concat(activeDataGroups);
+				const tick_html = this.getContainerSVG().select(
+					`g.axis.${axisPosition} g.ticks g.tick`
+				).html();
+				container
```
natashadecoste, line 456:
```suggestion

				container
```

packages/core/src/components/axes/axis.ts
```diff
@@ -414,6 +438,46 @@ export class Axis extends Component {
...
 
+		// truncate the label if it's too long
+		// only applies to discrete type
+		if (truncationTypeProvided !== TruncationTypes.NONE && !isTimeScaleType && axisOptions.scaleType === "labels") {
+			const dataGroups = this.model.getDataValuesGroupedByKeys();
+			if (dataGroups.length > 0) {
+				let label_data_array = [];
+				const firstDataGroupName = dataGroups[0].sharedStackKey;
+				const activeDataGroups = dataGroups.map(d => d.sharedStackKey);
+				if (activeDataGroups.length === dataGroups.length) {
```
natashadecoste, line 449:
```suggestion

				if (activeDataGroups.length === dataGroups.length) {
```

packages/core/src/components/axes/axis.ts
```diff
@@ -414,6 +438,46 @@ export class Axis extends Component {
...
+					label_data_array.push(firstDataGroupName);
+				}
+				label_data_array = label_data_array.concat(activeDataGroups);
+				const tick_html = this.getContainerSVG().select(
```
natashadecoste, line 453:
```suggestion
				const tick_html = svg.select(
```

packages/core/src/components/axes/axis.ts
```diff
@@ -414,6 +438,46 @@ export class Axis extends Component {
...
+					.selectAll("g.tick text")
+					.data(label_data_array)
+					.text(function(d) {
+						if (d.length > truncationThresholdProvided) {
+							return Tools.truncateLabel(d, truncationTypeProvided, truncationNumCharacterProvided);
+						} else {
+							return d;
+						}
+					});
+				this.getInvisibleAxisRef()
```
natashadecoste, line 469:
```suggestion

				this.getInvisibleAxisRef()
```

packages/core/src/components/axes/axis.ts
```diff
@@ -414,6 +438,46 @@ export class Axis extends Component {
...
+					label_data_array.push(firstDataGroupName);
+				}
+				label_data_array = label_data_array.concat(activeDataGroups);
+				const tick_html = this.getContainerSVG().select(
+					`g.axis.${axisPosition} g.ticks g.tick`
+				).html();
+				container
+					.selectAll("g.ticks g.tick")
+					.html(tick_html);
+				container
```
natashadecoste, line 459:
```suggestion

				container
```

packages/core/src/components/essentials/legend.ts
```diff
@@ -31,6 +32,32 @@ export class Legend extends Component {
...
+		);
+		// default config for truncation
+		// set type to END_LINE
+		// set threshold to 12
+		// set number of characters to show to 14
+		if (!truncationTypeProvided) {
+			truncationTypeProvided = TruncationTypes.END_LINE;
+			truncationThresholdProvided = 16;
+			truncationNumCharacterProvided = 14;
+		}
```
natashadecoste, line 59: same with this in, can we get it in `configuration.ts`?

packages/core/src/components/essentials/tooltip-radar.ts
```diff
@@ -1,8 +1,12 @@
 import { Tooltip } from "./tooltip";
 import { Tools } from "../../tools";
+import {TooltipTypes} from "../../interfaces";
```
natashadecoste, line 3:
```suggestion
import { TooltipTypes } from "../../interfaces";
```

packages/core/src/components/essentials/tooltip.ts
```diff
@@ -145,6 +148,8 @@ export class Tooltip extends Component {
 		if (type === TooltipTypes.TITLE) {
 			const title = this.model.getOptions().title;
 			return `<div class="title-tooltip"><text>${title}</text></div>`;
+		} else if (type === TooltipTypes.LEGEND) {
+			return `<div class="title-tooltip"><text>${data.name}</text></div>`;
```
natashadecoste, line 152:
```suggestion
			return `<div class="legend-tooltip"><text>${data.name}</text></div>`;
```
in case users want to target the classes separately

packages/core/src/components/essentials/tooltip-radar.ts
```diff
@@ -1,8 +1,12 @@
...
 
 export class TooltipRadar extends Tooltip {
-	getMultilineTooltipHTML(data: any) {
+	getMultilineTooltipHTML(data: any, type: TooltipTypes) {
+		if (type === TooltipTypes.LEGEND) {
+			return `<div class="title-tooltip"><text>${data.product}</text></div>`;
+		}
```
natashadecoste, line 9:
```suggestion
```
if we take this out, it will use `getTooltipHTML` from `tooltip.ts` 👍

packages/core/src/components/essentials/tooltip.ts
natashadecoste: or we can have a truncation tooltip class so we wont need to change this to 
```
		.title-tooltip {
			line-height: 16px;
			font-size: 12px;
			padding: 4px;
			min-width: 20px;
		}
```
to this 

```
		.title-tooltip, .legend-tooltip, .axis-tooltip {
			line-height: 16px;
			font-size: 12px;
			padding: 4px;
			min-width: 20px;
		}

```

packages/core/src/components/essentials/tooltip-scatter.ts
```diff
@@ -7,6 +7,10 @@ export class TooltipScatter extends Tooltip {
 		if (type === TooltipTypes.TITLE) {
 			// the main tooltip component handles title styles
 			return super.getTooltipHTML(datum, type);
+		} else if (type === TooltipTypes.LEGEND) {
+			return `<div class="title-tooltip"><text>${datum.name}</text></div>`;
+		} else if (type === TooltipTypes.AXISLABEL) {
+			return `<div class="title-tooltip"><text>${datum}</text></div>`;
```
natashadecoste, line 13: same as above, we either need to call the classes correctly by the tooltip type or have a trunction class for them all

packages/core/src/components/essentials/tooltip-bar.ts
```diff
@@ -165,23 +177,41 @@ export class TooltipBar extends Tooltip {
...
 	 * @param data associated values for the hovered bar
 	 */
-	getTooltipHTML(data: any) {
+	getTooltipHTML(data: any, type: TooltipTypes) {
+		if (type === TooltipTypes.LEGEND) {
+			const dataVal = data.name;
+			return `<div class="datapoint-tooltip"><p class="label">${dataVal}</p></div>`;
+		} else if (type === TooltipTypes.AXISLABEL) {
+			return `<div class="datapoint-tooltip"><p class="label">${data}</p></div>`;
+		}
```
natashadecoste, line 213: same as above, we either need to call the classes correctly by the tooltip type or have a trunction class for them all

linhenry0417:
@theiliad @natashadecoste I think all of the review comments have been addressed. Please let me know if there's any more questions. I can provide a quick fix. Thanks.

packages/core/src/tools.ts
```diff
@@ -217,6 +218,24 @@ export namespace Tools {
...
 
+		/**
+	 * Truncate the labels
+	 * @export
+	 * @param {any} fullText
+	 * @param {any} truncationType
+	 * @param {any} numCharacter
+	 * @returns Truncated text
+	 */
+	export function truncateLabel(fullText, truncationType, numCharacter) {
```
theiliad, line 229: what if fullText.length is 4 and numCharacter is 10? would we still want to show the `...`?

packages/core/src/interfaces/truncation.ts
```diff
@@ -0,0 +1,19 @@
...
+	 */
+	truncation?: {
+		/**
+		 * truncation type
+		 */
+		type?: string;
+		/**
+		 * truncation threshold
+		 */
+		threshold?: Number;
```
theiliad, line 13: what's the difference between `threshold` and `numCharacter`?

packages/core/src/interfaces/components.ts
```diff
@@ -26,7 +27,7 @@ export interface LayoutComponentChild {
 /**
  * customize the legend component
  */
-export interface LegendOptions {
```
theiliad, line 29: we should probably just include `truncation` inside `LegendOptions` pointing to `TruncationOptions` like so:
```ts
{
  truncation?: TruncationOptions;
}
```

just so we don't extend various interfaces in the future...

packages/core/src/components/essentials/tooltip-scatter.ts
```diff
@@ -7,6 +7,10 @@ export class TooltipScatter extends Tooltip {
 		if (type === TooltipTypes.TITLE) {
 			// the main tooltip component handles title styles
 			return super.getTooltipHTML(datum, type);
+		} else if (type === TooltipTypes.LEGEND) {
+			return `<div class="legend-tooltip"><p class="label">${datum.name}</p></div>`;
+		} else if (type === TooltipTypes.AXISLABEL) {
+			return `<div class="axis-tooltip"><p class="label">${datum}</p></div>`;
```
theiliad, line 13: Tooltip should at some point be refactored to just handle showing a single value rather than being called many different types of tooltips, but this is a working solution for now so 💯

packages/core/src/components/essentials/tooltip-bar.ts
```diff
@@ -149,6 +152,15 @@ export class TooltipBar extends Tooltip {
 			.node()
 			.getBoundingClientRect();
 		const barPosition = hoveredElement.getBoundingClientRect();
+		const { position: axisPosition } = this.configs;
+		const options = this.model.getOptions();
+		const axisOptions = Tools.getProperty(options, "axes", axisPosition);
+		let truncationThresholdProvided = Tools.getProperty(
```
theiliad, line 158: seems like you've set defaults for `threshold` in `Configuration.ts`, so all of these `...Provided` variables need to either be removed or renamed, and any checks for those values should be removed as well since they're always going to exist unless the user sets them to null/undefined.

One check that would be very important is for `truncation.type` so that if it's set to none we do nothing

packages/core/src/components/essentials/legend.ts
```diff
@@ -48,11 +72,24 @@ export class Legend extends Component {
 				return d.status === options.legend.items.status.ACTIVE;
 			});
 
-		addedLegendItems
-			.append("text")
-			.merge(legendItems.select("text"))
-			.html((d) => d.name)
-			.attr("alignment-baseline", "middle");
+		// truncate the legend label if it's too long
+		if (truncationTypeProvided !== TruncationTypes.NONE) {
```
theiliad, line 76:
```js
addedLegendItems
	.append("text")
	.merge(legendItems.select("text"));
```

could move out of the if-statement and be stored in a variable

packages/core/src/components/axes/axis.ts
```diff
@@ -462,6 +538,11 @@ export class Axis extends Component {
 					element: select(this),
 					datum,
 				});
+				if (!isTimeScaleType && axisOptions.scaleType && axisOptions.scaleType === "labels") {
+					self.services.events.dispatchEvent(Events.Tooltip.HIDE, {
```
theiliad, line 542: the tooltip hide event shouldn't need any detail or data being passed into it, is there a specific reason why this'd be needed?

packages/core/demo/data/bar.ts
```diff
@@ -104,6 +139,31 @@ export const simpleHorizontalBarOptions = {
 	},
 };
 
+export const simpleHorizontalBarLongLabelOptions = {
+	title: "Simple horizontal bar with long label (discrete)",
```
theiliad, line 143:
```suggestion
	title: "Simple horizontal bar (truncated labels)",
```

theiliad:
Also could we please add more demos to different chart types for this functionality?

packages/core/src/interfaces/truncation.ts
linhenry0417: Threshold here means the point where truncation applies. For example, if a word is 15 characters long and threshold is 10, it will be truncated. If a word is 5 characters long it will not be truncated. And for that truncated word I can set how many characters to be shown (`numCharacter`). We can just keep the `numCharacter` if you think `threshold` is unnecessary.

packages/core/src/tools.ts
linhenry0417: That's a good point. I'll show the fulltext in that case.

linhenry0417:
> have we tested this functionality on labels that don't need to be truncated?

The donut chart demo is in this case.

theiliad:
> > have we tested this functionality on labels that don't need to be truncated?
> 
> The donut chart demo is in this case.

I see, great thanks. Also I see a couple of new bugs as well:
1. We shouldn't be seeing a tooltip on a non-truncated item
image

2. Items seem to be aligning incorrectly now with the legend checkboxes

linhenry0417:
> > > have we tested this functionality on labels that don't need to be truncated?
> > 
> > 
> > The donut chart demo is in this case.
> 
> I see, great thanks. Also I see a couple of new bugs as well:
> 
>     1. We shouldn't be seeing a tooltip on a non-truncated item
>        image
> 
>     2. Items seem to be aligning incorrectly now with the legend checkboxes

1. Sure. I can fix that.

2. For charts like donut chart, I kept the legend tooltip position the same as it is on the donut chart itself, which is where the mouse is. For others like bar chart it appears right above the label. I think what you mean is we want all to be the same as bar chart?

theiliad:
> 2. For charts like donut chart, I kept the legend tooltip position the same as it is on the donut chart itself, which is where the mouse is. For others like bar chart it appears right above the label. I think what you mean is we want all to be the same as bar chart?

No what I meant is that labels that haven't been truncated shouldn't get a tooltip on hover

packages/core/src/components/essentials/legend.ts
linhenry0417: Yes.

packages/core/src/components/axes/axis.ts
linhenry0417: I can remove it. I saw that other tooltip hide events passed in data but didn't get used as well. I guess we can also clean it up.

linhenry0417:
>     2. Items seem to be aligning incorrectly now with the legend checkboxes
I've checked the storybook https://carbon-design-system.github.io/carbon-charts/react/?path=/story/pie--pie it seems the issue is there already. Maybe we can create a defect for it?

linhenry0417:
I've added a demo for line chart. 
Screen Shot 2020-06-18 at 12 19 20 AM

packages/core/src/interfaces/components.ts
linhenry0417: I tried the code above but it didn't seem to work. For some reason it will not recognize the truncation in the interface. However it works if I extends TruncationOptions..

linhenry0417:
@theiliad I've added another demo for scatter chart. I'm not sure why deploy-preview is failing. It seems it passes and fails from time to time.

theiliad:
> https://carbon-design-system.github.io/carbon-charts/react/?path=/story/bar-horizontal--simple-horizontal-bar-discrete

This link renders properly for me

image

packages/core/demo/data/bar.ts
```diff
@@ -74,6 +74,40 @@ export const simpleBarOptions = {
...
+		left: {
+			mapsTo: "value",
+		},
+		bottom: {
+			mapsTo: "group",
+			scaleType: "labels",
+			truncation: {
+				type: "end_line",
+				threshold: 10,
+				numCharacter: 12
```
theiliad, line 98: can we set these values by default in `mergeDefaultChartOptions` in `tools.ts`?

linhenry0417:
> On my end there are only 2 more changes needed
> 
>     1. setting axis & legend default truncations (axis should be done in `mergeDefaultChartOptions()` in `tools.ts`, and legend in `configuration.ts`)
> 
>     2. The alignment issue in legend items, it seems that the current charting package doesn't have any aligning issues but in the deploy previews for this PR I see the issue

@theiliad I think the above two points were resolved in my last commit. Could you help to take a look? Thanks.

packages/core/demo/data/bar.ts
linhenry0417: Yes.

linhenry0417:
> looks good, just seeing a new bug
> 
> image
> 
> when filtering to this dataset, the line and the axis are showing incorrect info

I'm aware of this. I think this is an issue or bug with the D3 axis rendering library. Even if the data supplied is correct the labels would still display incorrectly after the render function is called. For example I'd need to add the first label again to the front of the data in order for it to display correctly. But I can still try to fix it.

linhenry0417:
> looks good, just seeing a new bug
> 
> image
> 
> when filtering to this dataset, the line and the axis are showing incorrect info

@theiliad This problem should be fixed now. I've verified working via storybook. It'd great if you could help to verify as well. Thanks.

theiliad:
> > looks good, just seeing a new bug
> > image
> > when filtering to this dataset, the line and the axis are showing incorrect info
> 
> @theiliad This problem should be fixed now. I've verified working via storybook. It'd great if you could help to verify as well. Thanks.

looking great, thanks @linhenry0417 💯

packages/core/src/components/axes/axis.ts
```diff
@@ -448,6 +522,12 @@ export class Axis extends Component {
 						datum,
 					}
 				);
+				if (!isTimeScaleType && axisOptions.scaleType && axisOptions.scaleType === "labels" && datum.length > truncationThreshold) {
```
theiliad, line 525: Can these checks be stored in a variable? They're fairly long, and we should use `Tools.getProperty()` to avoid the double checks (`axisOptions.scaleType && axisOptions.scaleType === "labels"`)

packages/core/src/components/axes/axis.ts
```diff
@@ -414,6 +432,51 @@ export class Axis extends Component {
 			container.attr("opacity", 0);
 		}
 
+		// truncate the label if it's too long
+		// only applies to discrete type
+		if (truncationType !== TruncationTypes.NONE && !isTimeScaleType && axisOptions.scaleType === "labels") {
```
theiliad, line 437: Can these checks be stored in a variable? They're fairly long, and we should use Tools.getProperty() to avoid the double checks (axisOptions.scaleType && axisOptions.scaleType === "labels")

also we might not need `!isTimeScaleType` because we're already checking if `axisOptions.scaleType === "labels"`

theiliad: also `"labels"` should say `ScaleTypes.LABELS`

packages/core/demo/data/bar.ts
```diff
@@ -74,6 +74,35 @@ export const simpleBarOptions = {
...
+			mapsTo: "value",
+		},
+		bottom: {
+			mapsTo: "group",
+			scaleType: "labels",
+		},
+	},
+	legend: {
+		truncation: {
+			type: "end_line",
```
theiliad, line 99:
```suggestion
			type: TruncationTypes.END_LINE,
```

packages/core/demo/data/bar.ts
```diff
@@ -104,6 +134,31 @@ export const simpleHorizontalBarOptions = {
...
 };
 
+export const simpleHorizontalBarLongLabelOptions = {
+	title: "Simple horizontal bar (truncated labels)",
+	axes: {
+		left: {
+			mapsTo: "group",
+			scaleType: "labels",
+			truncation: {
+				type: "mid_line",
```
theiliad, line 144:
```suggestion
				type: TruncationTypes.MID_LINE,
```

packages/core/demo/data/bar.ts
```diff
@@ -104,6 +134,31 @@ export const simpleHorizontalBarOptions = {
...
+				numCharacter: 14
+			},
+		},
+		bottom: {
+			mapsTo: "value",
+		},
+	},
+	legend: {
+		truncation: {
+			type: "mid_line",
```
theiliad, line 155:
```suggestion
			type: TruncationTypes.MID_LINE,
```

packages/core/src/components/essentials/legend.ts
```diff
@@ -31,6 +32,34 @@ export class Legend extends Component {
...
+		if (!truncationType) {
+			truncationType = Configuration.legend.truncation.type;
+		}
+		if (!truncationThreshold) {
+			truncationThreshold = Configuration.legend.truncation.threshold;
+		}
+		if (!truncationNumCharacter) {
+			truncationNumCharacter = Configuration.legend.truncation.numCharacter;
+		}
+
```
theiliad, line 62: `legend` options that you've set defaults for in `configuration.ts` are part of the base chart options which are used in merging user-provided options with our defaults

theiliad: image

theiliad: so your default legend values will be there unless the user sets them to `null` or `undefined`, you don't need to set them manually

packages/core/src/components/essentials/legend.ts
```diff
@@ -227,6 +268,15 @@ export class Legend extends Component {
...
 		const options = this.model.getOptions();
+		const legendOptions = Tools.getProperty(options, "legend");
+		let truncationThreshold = Tools.getProperty(
+			legendOptions,
+			"truncation",
+			"threshold"
+		);
+		if (!truncationThreshold) {
+			truncationThreshold = Configuration.legend.truncation.threshold;
+		}
```
theiliad, line 279: legend options that you've set defaults for in configuration.ts are part of the base chart options which are used in merging user-provided options with our defaults

so your default legend values will be there unless the user sets them to null or undefined, you don't need to set them manually

packages/core/src/components/essentials/legend.ts
```diff
@@ -269,10 +319,22 @@ export class Legend extends Component {
...
+					hoveredElement: hoveredItem,
+					type: TooltipTypes.LEGEND,
+				});
+				}
+			})
 			.on("mouseout", function () {
 				const hoveredItem = select(this);
 				hoveredItem.select("rect.hover-stroke").remove();
 
+				self.services.events.dispatchEvent(Events.Tooltip.HIDE, {});
```
theiliad, line 336: tooltip hide event doesn't need details

theiliad:
```suggestion
				self.services.events.dispatchEvent(Events.Tooltip.HIDE);
```

packages/core/src/components/essentials/tooltip-bar.ts
```diff
@@ -13,6 +14,7 @@ import settings from "carbon-components/es/globals/js/settings";
 
 // D3 Imports
 import { select } from "d3-selection";
+import {Configuration} from "tslint";
```
theiliad, line 17: probably a mistaken import?

packages/core/src/configuration.ts
```diff
@@ -108,15 +113,35 @@ export const barChartTooltip: BarTooltipOptions = Tools.merge(
 const axes: AxesOptions = {
 	top: {
 		includeZero: true,
+		truncation: {
```
theiliad, line 116: maybe we can store this in a variable? `standardTruncationOptions`?

packages/core/src/configuration.ts
```diff
@@ -50,6 +50,11 @@ export const legend: LegendOptions = {
 		radius: 6.5,
 		spaceAfter: 4,
 	},
+	truncation: {
+		type: TruncationTypes.END_LINE,
+		threshold: 16,
+		numCharacter: 14,
+	},
```
theiliad, line 57: maybe we can store this in a variable? standardTruncationOptions?

packages/core/demo/data/bar.ts
linhenry0417: I tried to `import { TruncationTypes } from "../../src/interfaces/enums";` in the bar.ts demo data file before. It worked fine in the storybook. However when I ran yarn build it was giving me error `[!] Error: Could not resolve '../../src/interfaces/enums' from dist/demo/data/bar.js`. This is quite strange. I am not sure why.

packages/core/src/components/axes/axis.ts
linhenry0417: Resolved. Thanks.

packages/core/src/components/axes/axis.ts
linhenry0417: Resolved. Thanks.

packages/core/src/components/essentials/legend.ts
linhenry0417: Resolved. Thanks.

packages/core/src/components/essentials/legend.ts
linhenry0417: Resolved. Thanks.

packages/core/src/components/essentials/legend.ts
linhenry0417: Resolved. Thanks.

packages/core/src/components/essentials/tooltip-bar.ts
linhenry0417: Removed. Thanks.

packages/core/src/configuration.ts
linhenry0417: Resolved. Thanks.

packages/core/src/configuration.ts
linhenry0417: Resolved. Thanks.

packages/core/src/components/essentials/tooltip-bar.ts
```diff
@@ -181,7 +189,14 @@ export class TooltipBar extends Tooltip {
 	 * Returns the html for the bar single point tooltip
 	 * @param data associated values for the hovered bar
 	 */
-	getTooltipHTML(data: any) {
+	getTooltipHTML(data: any, type: TooltipTypes) {
+		if (type === TooltipTypes.LEGEND) {
+			const dataVal = data.name;
+			return `<div class="legend-tooltip"><p class="label">${dataVal}</p></div>`;
```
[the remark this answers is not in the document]
linhenry0417: Resolved. Thanks.

Issue #650, opened 2020-06-02
Scatter chart - multi-tooltip doesn't show on hover over data points


navjeets:
#### I am submitting a...

- [ ] Feature request
- [ ] Design defect
- [x] Source code defect
- [ ] Demo/documentation defect
- [ ] Other

#### charts version:
0.30.24
### Issue description
If there are multiple data points, with same x and y values, the multi-point tooltip only shows on hover over grid and not the actual point. This may lead to user not knowing that there are multiple data points especially when AxisTooltipOptions.gridline.enabled is set to false.
#### Steps to produce the issue
Create new Scatter chart with dataset containing multiple records with same x and y values. Observe that the multi-tooltip shows on hover over grid but not the point.
#### Current behavior
If there are multiple points with same x and y values, multi-point tooltip doesn't show on hover over data point.
#### Expected behavior
Multi-point tooltip should show on hover over data points with same x and y values.

theiliad:
According to the specs, the multidata tooltip should only show when hovering on a point if there are other points behind that hovered point which cannot be seen properly.

If you're hovering over just 1 point you should always be seeing the singular tooltip.

navjeets:
Yes. This issue is about the scenario where there are multiple points with same x and y values (i.e. overlapping points) and multidata tooltip doesn't show on hover.

theiliad:
yes, just leaving that piece as an FYI for everyone else that might get to this issue

PR #652, opened 2020-06-02
fix(scatter): multi-tooltip doesn't show on hover for overlapping data points

navjeets:
### Updates
- Updated addEventListeners in scatter.ts file to implement multidata tooltip on hover over overlapping data points.

Closes issue #650 

- [ ] Demos all features
- [ ] Documented/annotated
- [ ] Matches UI/UX specs
- [ ] Accessible


packages/core/src/components/graphs/scatter.ts
```diff
@@ -251,6 +263,96 @@ export class Scatter extends Component {
...
+						type: TooltipTypes.DATAPOINT,
+					});
+				} else {
+					// Show tooltip
+					self.services.events.dispatchEvent(Events.Tooltip.SHOW, {
+						hoveredElement,
+						type: TooltipTypes.DATAPOINT,
+					});
+				}
+
```
theiliad, line 355: a massive amount of logic for something so simple, there has to be an easier way to do this

navjeets: I used the same logic used in showRuler() method in carbon-charts/packages/core/src/components/axes/ruler.ts

Alternatively, I can select all circles with cx and cy within certain threshold of current hovered element to be included in multidata tooltip. That will be a lot simpler than this and effectively be doing the same thing.

navjeets: To further simplify, I can remove the reduce() logic and replace it with "threshold" set to the points.radius value. This way the multidata tooltip will shown for points that overlap.

navjeets: @theiliad I have updated the code to use points.radius as threshold to determine the points to include in multidata tooltip. So if the points are overlapping, they will be included in the tooltip. This allowed me to remove the complex logic in map.reduce call.

theiliad: no there's no need for any that.

we should be able to simply:
1. get domain value of highlighted point
2. get range value of highlighted point
3. find any matching datapoints in `displayData` which don't belong in the same datagroup
4. pass data into tooltip

Then as long as we pass `multiData` into the tooltip it should show those values 
image

navjeets: So we only want to show multidata tooltip if the domain and range values are exact match?

navjeets: @theiliad I have updated the code as per your recommendation. I don't consider datagroup while filtering because in my use case, the multiple points are in same datagroup but are different because of some other attributes.

navjeets: I do want mention that the downside of not using the threshold approach is that its possible that the domain or range values may be different by a small fraction (not visible in UI). So to the user it will still appear that there's 1 point and the UI won't show multidata tooltip on hover over the overlapping points.

navjeets: @theiliad Can you please review the latest changes?

theiliad:
it could also be related to the merge conflicts

PR #656, opened 2020-06-05
chore(core, react, angular, vue, svelte): update demo data

theiliad:


packages/core/demo/data/index.ts
```diff
@@ -184,136 +181,129 @@ let allDemoGroups = [
...
 				data: bubbleDemos.bubbleSkeletonData,
 				chartType: chartTypes.BubbleChart,
-				isDemoExample: true,
+				isDemoExample: false,
 			},
 		],
 	},
 	{
 		title: "Donut",
+		description: "Bubble charts use data points and bubbles to plot measures anywhere along a scale. One measure is plotted along each axis. The size of the bubble represents the third measure. You can use bubble charts to represent financial data or any data where measured values are related.",
```
Donisius, line 278: This doesn't seem to be the right description

packages/core/demo/data/index.ts
```diff
@@ -605,8 +607,9 @@ export const storybookDemoGroups = Tools.clone(allDemoGroups);
 
 // in the demo page we want to show only demos with isDemoExample = true
 export const demoGroups = Tools.clone(allDemoGroups)
+	// remove demoGroup if it's children don't have any demo examples
```
Donisius, line 610:
```suggestion
	// remove demoGroup if its children don't have any demo examples
```

packages/core/demo/data/index.ts
```diff
@@ -605,8 +607,9 @@ export const storybookDemoGroups = Tools.clone(allDemoGroups);
...
+	.filter((demoGroup) => demoGroup.demos.filter(demo => demo.isDemoExample !== false).length > 0)
 	.map((demoGroup) => {
-		demoGroup.demos = demoGroup.demos.filter(demo => demo.isDemoExample);
+		demoGroup.demos = demoGroup.demos.filter(demo => demo.isDemoExample !== false);
```
Donisius, line 613: Can this just be
```suggestion
		demoGroup.demos = demoGroup.demos.filter(demo => !demo.isDemoExample);
```

packages/core/demo/data/index.ts
```diff
@@ -605,8 +607,9 @@ export const storybookDemoGroups = Tools.clone(allDemoGroups);
 
 // in the demo page we want to show only demos with isDemoExample = true
 export const demoGroups = Tools.clone(allDemoGroups)
+	// remove demoGroup if it's children don't have any demo examples
+	.filter((demoGroup) => demoGroup.demos.filter(demo => demo.isDemoExample !== false).length > 0)
```
Donisius, line 611: Can this just be
```suggestion
	.filter((demoGroup) => demoGroup.demos.filter(demo => !demo.isDemoExample).length > 0)
```

packages/core/demo/data/index.ts
theiliad: when `isDemoExample` is `null` or `undefined` we don't want this condition to pass, and also it's safer type-wise this way

packages/core/demo/data/index.ts
theiliad: when isDemoExample is null or undefined we don't want this condition to pass, and also it's safer type-wise this way

packages/core/demo/data/index.ts
theiliad:
```suggestion
```

PR #663, opened 2020-06-10
display timestamp in detail in tooltip if it's time series chart

linhenry0417:
### Updates
- list
- out
- updates

- [ ] Demos all features
- [ ] Documented/annotated
- [ ] Matches UI/UX specs
- [ ] Accessible

### DIFF
--- a/packages/core/src/components/essentials/tooltip-bar.ts
+++ b/packages/core/src/components/essentials/tooltip-bar.ts
@@ -189,8 +189,15 @@ export class TooltipBar extends Tooltip {
 		)
 			? this.model.getOptions().tooltip.valueFormatter(data.value)
 			: data.value.toLocaleString("en");
-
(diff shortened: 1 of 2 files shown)


theiliad:
I don't think we can do this, this might be something you'd just customize manually.

@designertyler

linhenry0417:
@theiliad How do we customize manually? Maybe not all the users would like to have this feature, in this case, what should we do? What if we have an option to enable timestamp display?

theiliad:
> @theiliad How do we customize manually? Maybe not all the users would like to have this feature, in this case, what should we do? What if we have an option to enable timestamp display?

Changes to the tooltip need to be designed, then reviewed by more designers. In your case you could easily customize the tooltip with:
```js
{
  tooltip: {
    customHTML: (data, defaultHTML) => "return something"
  }
}
```

Issue #668, opened 2020-06-15
timeScale option missing for AxisOptions interface


carmiele:
#### I am submitting a...

- [ ] Feature request
- [ ] Design defect
- [x] Source code defect
- [ ] Demo/documentation defect
- [ ] Other

### Issue description
`AxisOptions` is missing the `timeScale` option.

#### Steps to produce the issue
found in latest commit: https://github.com/carbon-design-system/carbon-charts/blob/0402e2a2c3c99b780ebfe6dc9fe1238c3106b002/packages/core/src/axis-chart.ts

#### Current behavior
Missing option to use `timeScale` when typecasting options objects to `AxisOptions` and its inherited interfaces/classes.

#### Expected behavior
Have option to use `timeScale` when typecasting options objects to `AxisOptions` and its inherited interfaces/classes.

theiliad:
@sophiiae could you please take a look?

sophiiae:
> @sophiiae could you please take a look?

Sure.

PR #669, opened 2020-06-15
fix: add missing timescale option in AxisOptions interface

carmiele:
### Updates
- #668 - add missing `timescale` option in `AxisOptions` interface


theiliad:
@sophiiae looks like Carmen has already made the PR

PR #672, opened 2020-06-17
Fix AreaPercentageChart domain issue

sophiiae:


packages/core/src/model.ts
```diff
@@ -52,18 +52,25 @@ export class ChartModel {
...
 			Object.keys(axesOptions).forEach(axis => {
-				if (axesOptions[axis].domain && axesOptions[axis].mapsTo) {
-					const mapsTo = axesOptions[axis].mapsTo;
-					const [start, end] = axesOptions[axis].domain;
-
-					// Filter out data outside domain
-					displayData = displayData.filter((datum) => 
-						datum[mapsTo] >= start && datum[mapsTo] <= end
-					)
+				const mapsTo = axesOptions[axis].mapsTo;
```
Donisius, line 59: Do we still need to check if `axesOptions[axis].mapsTo` exists?

packages/core/src/model.ts
```diff
@@ -52,18 +52,29 @@ export class ChartModel {
...
+
+				if (
```
Donisius, line 61: Before it checked for the existence of `axesOptions[axis].mapsTo` before setting it to mapsTo. Do you think it makes more sense to do that because if mapsTo doesn't exist it won't be used anywhere else?

packages/core/src/model.ts
```diff
@@ -52,18 +52,29 @@ export class ChartModel {
...
+					mapsTo &&
+					axesOptions[axis].domain &&
+					!this.getOptions().percentage
+				) {
```
Donisius, line 65: Assuming prettier isn't formatting it like this maybe:
```suggestion
				if (mapsTo
					&& axesOptions[axis].domain
					&& !this.getOptions().percentage) {
```

packages/core/src/model.ts
sophiiae: I know, but `continue` doesn't work in `forEach`, I need to turn `forEach` to a `for` loop for that.

packages/core/src/model.ts
sophiiae: I don't think it formats in this way.

packages/core/src/model.ts
sophiiae: If the `axesOptions[axis].mapsTo` doesn't exist, then the `const mapsTo` will be undefined. In the `if` check, if mapsTo is undefined, it will skip the rest and won't use `mapsTo` in anywhere.

Donisius: So you can't do something like:
```
	if (axesOptions[axis].mapsTo
		&& axesOptions[axis].domain
		&& !this.getOptions().percentage){
		mapsTo = axesOptions[axis].mapsTo;
        ...etc
	}
```
?

packages/core/src/model.ts
Donisius: okay

packages/core/src/model.ts
```diff
@@ -52,20 +52,32 @@ export class ChartModel {
...
+						displayData = displayData.filter((datum) =>
+							axesOptions[axis].domain.includes(datum[mapsTo])
+						);
+					} else {
+						const [start, end] = axesOptions[axis].domain;
+
+						// Filter out data outside domain
+						displayData = displayData.filter(
+							(datum) =>
+								datum[mapsTo] >= start && datum[mapsTo] <= end
```
Donisius, line 76:
```suggestion
						displayData = displayData.filter((datum) =>
							datum[mapsTo] >= start && datum[mapsTo] <= end
```

sophiiae: It's prettier formatting.

Donisius: I don't get it why wouldn't it do it for the filter above??

theiliad:
update, we're just waiting on a final update here to get add `percentage` as a flag inside each axis rather than the entire chart options

PR #678, opened 2020-06-22
fix(core): stroke dissappears when remapping the dimensions

natashadecoste:
### Updates
- list
- out
- updates

- [ ] Demos all features
- [ ] Documented/annotated
- [ ] Matches UI/UX specs
- [ ] Accessible


packages/core/src/components/graphs/line.ts
```diff
@@ -51,25 +51,27 @@ export class Line extends Component {
...
 
 			data = stackedData.map((d) => ({
-				name: d[0].group,
+				name: d[0][groupMapsTo],
 				data: d.map((datum) => ({
-					date: datum.data.sharedStackKey,
-					group: datum.group,
-					value: datum[1],
+					[domainIdentifier]: datum.data.sharedStackKey,
+					group: datum[groupMapsTo],
```
theiliad, line 71: we don't need to do `groupMapsTo: datum[groupMapsTo]` here?

natashadecoste: 👍  yeah youre right

PR #681, opened 2020-06-24
Fix stacked area chart with uneven data

### LINKED ISSUE #674: Stacked Area Chart - graph with uneven data
#### I am submitting a...

- [ ] Feature request
- [ ] Design defect
- [x] Source code defect
- [ ] Other

### Issue description
The Stacked Area Chart draws the graph with multiple start points when any data in the first dataset has been removed. 

Run Stacked Area Chart Demo, and comment out any data in Dataset 1. 

#### Current behavior

Comment out Jan 1 data in Dataset 1: 
image

Comment out Jan 8 data in Dataset 1: 
image

sophiiae:
closes #674

### DIFF
--- a/packages/core/src/model.ts
+++ b/packages/core/src/model.ts
@@ -169,7 +169,11 @@ export class ChartModel {
 
 		// Sort keys
 		if (scaleType === ScaleTypes.TIME) {
-			stackKeys.sort((a, b) => new Date(a) - new Date(b));
+			stackKeys.sort((a: any, b: any) => {
+				const dateA: any = new Date(a);
+				const dateB: any = new Date(b);
+				return dateA - dateB;
+			});
 		} else if (
 			scaleType === ScaleTypes.LOG ||
 			scaleType === ScaleTypes.LINEAR
--- a/packages/core/src/model.ts
+++ b/packages/core/src/model.ts
@@ -164,12 +164,22 @@ export class ChartModel {
 			(datum) => datum[domainIdentifier]
 		).keys();
 
+		const axisPosition = this.services.cartesianScales.domainAxisPosition; 
+		const scaleType = options.axes[axisPosition].scaleType;
+
 		// Sort keys
-		stackKeys.sort((a, b) => {
-			const dateA: any = new Date(a);
-			const dateB: any = new Date(b);
-			return dateA - dateB;
-		})
+		if (scaleType === ScaleTypes.TIME) {
+			stackKeys.sort((a, b) => {
+				const dateA: any = new Date(a);
+				const dateB: any = new Date(b);
+				return dateA - dateB;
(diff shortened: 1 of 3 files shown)


packages/core/src/model.ts
```diff
@@ -163,6 +163,14 @@ export class ChartModel {
 			displayData,
 			(datum) => datum[domainIdentifier]
 		).keys();
+
+		// Sort keys
+		stackKeys.sort((a, b) => {
+			const dateA: any = new Date(a);
+			const dateB: any = new Date(b);
```
theiliad, line 170: are a & b always going to be strings of date values? will they ever be actual Date objects? do we need a check here?

theiliad:
also, could you please make one of the demos build with uneven data so we have a visual test case for this to always look at.

packages/core/src/model.ts
sophiiae: Since `const stackKeys = map().keys()` return an string array, we don't have to check the type of a & b again.

packages/core/src/model.ts
```diff
@@ -163,6 +163,24 @@ export class ChartModel {
...
+		const axisPosition = this.services.cartesianScales.domainAxisPosition; 
+		const scaleType = options.axes[axisPosition].scaleType;
+
+		// Sort keys
+		if (scaleType === ScaleTypes.TIME) {
+			stackKeys.sort((a, b) => {
+				const dateA: any = new Date(a);
+				const dateB: any = new Date(b);
+				return dateA - dateB;
+			})
```
Donisius, line 176: optional

```suggestion
			stackKeys.sort((a, b) => new Date(a) - new Date(b));
```

otherwise

```suggestion
			stackKeys.sort((a, b) => {
				const dateA: any = new Date(a);
				const dateB: any = new Date(b);
				return dateA - dateB;
			});
```

sophiiae: I tried that, but Type error.

Donisius: Okay, then just the semicolon then :)

theiliad: `(a: any, b: any) => new Date(a) - new Date(b)`

would be the best way

sophiiae: The left and right side of arithmetic operator cannot be Date, so it requires `new Date(a): any` and `new Date(b): any`.

theiliad: ok

can you re-push the change please? I commitd doni's suggestion already

sophiiae: No worries, I did.

theiliad:
> LGTM 👍 One Q - is this sorting something that users could want to override? Would it make sense to expose it in the config?

Hey, it isn't

PR #682, opened 2020-06-24
fix(core): add tooltip to gauge chart

natashadecoste:
I noticed that tooltip was never added in with the gauge chart, it is needed for title truncation

- [ ] Demos all features
- [ ] Documented/annotated
- [ ] Matches UI/UX specs
- [ ] Accessible

### DIFF
--- a/packages/core/src/charts/gauge.ts
+++ b/packages/core/src/charts/gauge.ts
@@ -7,8 +7,7 @@ import { Tools } from "../tools";
 // Components
 import {
 	Gauge,
-	// the imports below are needed because of typescript bug (error TS4029)
-	TooltipPie
+	Tooltip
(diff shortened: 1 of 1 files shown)


theiliad:
I don't see a tooltip in the deploy previews. Why do we need the tooltip anyways? Does it show anything?

natashadecoste:
@theiliad yeah so we do cause otherwise with a long title we wont get the title truncation tooltip! so change the title for guage to a really long one and then you will see the tooltip

theiliad:
> @theiliad yeah so we do cause otherwise with a long title we wont get the title truncation tooltip! so change the title for guage to a really long one and then you will see the tooltip

That's fair

PR #691, opened 2020-07-07
improvement(skeleton): honour loading config to show skeleton chart

### LINKED ISSUE #677: Support showing skeleton UI even when data is passed in
#### I am submitting a...

- [X] Feature request
- [ ] Design defect
- [ ] Source code defect
- [ ] Other

#### charts version:

`0.32.7`

### Issue description

If I pass _true_ to `loading` option and pass data in the `data` prop, the charts crashes. I propse to avoid that behaviour, and continue showing the skeleton UI as long as `loading` is _true_. Regardless of the state of the `data` prop.

The rationale behind this is if I have feature that requires fetching multiple things from the server for different parts of the UI. And the UI loading state is controlled by a single boolean flag that's set to false only when all the requests finish. If the request fetching the charts data finishes first, I'll end up passing _true_ to the `loading` option and passing the charts data as well.

I could use multiple boolean flags as a workaround for each request, but I think it would be a nice improvement if the library can still display the skeleton chart as long as the `loading` option is _true_.

https://github.com/carbon-design-system/carbon-charts/blob/4ecd191a07a9797de898aa95e3eac77b6ac502e4/packages/core/src/components/graphs/skeleton.ts#L35-L39

https://codesandbox.io/s/carbon-charts-loading-data-e6b2x?file=/src/index.js

#### Current behavior

App crashes

#### Expected behavior

The Charts shouldn't crash. It should just continue showing the Skeleton, as if the `loading` option is taking precedence over the `data` prop

natashadecoste:
fix #677

### Updates
- list
- out
- updates

- [ ] Demos all features
- [ ] Documented/annotated
- [ ] Matches UI/UX specs
- [ ] Accessible


packages/core/src/components/axes/axis.ts
```diff
@@ -269,11 +270,12 @@ export class Axis extends Component {
 
 		// Position the axis title
 		// check that data exists, if they don't, doesn't show the title axis
+		const isDataEmpty = this.model.isDataEmpty();
 		if (axisOptions.title) {
 			const axisTitleRef = DOMUtils.appendOrSelect(
 				container,
 				`text.axis-title`
-			).html(this.model.isDataEmpty() ? "" : axisOptions.title);
+			).html(isDataEmpty || isDataLoading ? "" : axisOptions.title);
```
theiliad, line 278:
```suggestion
			).html((isDataEmpty || isDataLoading) ? "" : axisOptions.title);
```

PR #709, opened 2020-07-16
fix(core, angular, vue, react, svelte): debounce ruler renders when more than 100 datapoints

theiliad:


packages/core/src/tools.ts
```diff
@@ -37,6 +39,24 @@ export namespace Tools {
 	export const fromPairs = lodashFromPairs;
 	export const some = lodashSome;
 
+	export function debounceWithD3MousePosition(fn, delay, element) {
```
Donisius, line 42: Would it be possible/better to take advantage of `lodash`'s debounce function instead? https://lodash.com/docs/#debounce

theiliad: no, we already do use lodash, however in this case the d3 `event` object is immediately changed after the event occurs, which will cause the `setTimeout` in the debounce to return `undefined` for `event`

Donisius: okay

PR #711, opened 2020-07-16
adjust the bar width to prevent bars from overlapping

linhenry0417:
### Updates
- list
- out
- updates
Issue link: https://github.com/carbon-design-system/carbon-charts/issues/712

- [ ] Demos all features
- [ ] Documented/annotated
- [ ] Matches UI/UX specs
- [ ] Accessible


theiliad:
> Good idea, haven't seen the deploy previews yet

@linhenry0417 could you please add a demo with a dense amount of bars? like you have in your product

linhenry0417:
Added a demo. I may need to test more to make sure it works fine though.

packages/core/demo/data/bar.ts
```diff
@@ -230,6 +253,22 @@ export const simpleHorizontalBarTimeSeriesOptions = {
 	}
 };
 
+// Vertical simple time series with dense data
+export const simpleHorizontalBarTimeSeriesOptionsDense = {
```
sophiiae, line 257:
```suggestion
export const simpleBarTimeSeriesDenseOptions
```

packages/core/src/components/graphs/bar.ts
```diff
@@ -1,16 +1,21 @@
 // Internal Imports
 import { Component } from "../component";
+import { DOMUtils } from "../../services";
 
 export class Bar extends Component {
 	// Gets the correct width for bars based on options & configurations
 	protected getBarWidth(allDataLabels?: string[]) {
 		const options = this.model.getOptions();
+		const numData = this.model.getData().length;
```
sophiiae, line 9:
```suggestion
		const numberOfData = this.model.getData().length;
```

packages/core/src/components/graphs/bar.ts
```diff
@@ -1,16 +1,21 @@
...
 		const mainXScale = this.services.cartesianScales.getMainXScale();
+		const chartWidth = DOMUtils.getSVGElementSize(this.parent, {
+			useAttrs: true
+		}).width;
 
 		if (!mainXScale.step) {
 			return Math.min(
 				options.bars.maxWidth,
-				(5 / mainXScale.ticks().length) * options.bars.maxWidth
+				chartWidth * 0.5 / numData
```
sophiiae, line 18:
```suggestion
				chartWidth * 0.5 / numberOfData
```

packages/core/src/components/graphs/bar.ts
theiliad: I'd even suggest `numberOfDatapoints`, and shouldn't we be using `this.model.getDisplayData()` here instead?

linhenry0417: Changes made.

linhenry0417:
Actually in our use case we have only one data with different timestamps and values, unlike the demo I added. I guess it's the difference between bar charts and histogram. I'm thinking maybe we can open up a separate issue dealing with legend issue in bar charts?

linhenry0417:
Screen Shot 2020-07-21 at 6 05 06 PM
I tested with several different data and it seems the bars did not overlap as we increase the width like before.

packages/core/src/components/graphs/bar.ts
```diff
@@ -1,16 +1,21 @@
...
 		const mainXScale = this.services.cartesianScales.getMainXScale();
+		const chartWidth = DOMUtils.getSVGElementSize(this.parent, {
+			useAttrs: true
+		}).width;
 
 		if (!mainXScale.step) {
 			return Math.min(
 				options.bars.maxWidth,
-				(5 / mainXScale.ticks().length) * options.bars.maxWidth
+				chartWidth * 0.25 / numberOfDatapoints
```
theiliad, line 18: how did you get to this calculation? could you explain?

linhenry0417: Yes. The maximum width for the bars in the chart can be `width of the chart / numberOfDatapoints` so that they don't overlap with each other. But we need to leave some spaces between the bars and the beginning and end of the chart, which is why I multiply by 0.25 here. How I determined this number was sort of trial and error with various different data. With dense data of minutes of difference (just like my demo), even seconds, any number greater than this would result in overlapping. This seams to work for all the data for me. The original logic in determining was wrong and could result in bar width greater than the chart width.

sophiiae:
@linhenry0417 there are conflicts

linhenry0417:
I thought Eliad would merge it once approved? And he will rebase the code?

sophiiae:
> I thought Eliad would rebase this branch once he merges the PR?

it's better not to have conflicts when it merges.

theiliad:
@linhenry0417 I tried to solve the conflict in the Github UI but seems like there's still an error. Please solve locally

linhenry0417:
@theiliad I've rebased it onto latest master branch. If we are good let's merge this. Thanks.

linhenry0417:
I've also double checked the files changed, looks right to me.

linhenry0417:
Is it ready to be merged?

PR #713, opened 2020-07-17
Custom tick values

### LINKED ISSUE #540: Min option is not working in simple bar charts
#### I am submitting a...
-  Source code defect

#### charts version:
"@carbon/charts": "^0.24.3",
"@carbon/charts-react": "^0.24.3"

### Issue description

Unable to define min tick value in simple bar charts.

We are trying to avoid decimal values in Y axis ticks:
image

We have tried using formatter with math.floor or ceil   however since we have restricted the number of ticks to 5 it looks something like this which is not correct, so if we have any option where we can define domains if the calculated max range is less than 5 it might work

image

#### Current behavior

Round of the ticks value which repeats the value if the ticks decimal value  is  in same range.

#### Expected behavior

Value should not be repeated and it should be a whole number intead of decimal

natashadecoste:
### Updates
- allows users to pass in tickValues([]) if they dont want any ticks on the axis
- can be used to assert tick values on a time, linear, or discrete axis
- the truncation logic needs to also be taken care of by the user (label scaletype is being given custom ticks)

image
```
ticks: {
    customTicks: [0, 1.2, 1.3 , 2]
}
```

closes #540 

- [ ] Demos all features
- [ ] Documented/annotated
- [ ] Matches UI/UX specs
- [ ] Accessible


packages/core/src/components/axes/axis.ts
```diff
@@ -256,6 +264,12 @@ export class Axis extends Component {
 		// Set ticks formatter
 		axis.tickFormat(formatter);
 
+		// prioritize using a custom array of values rather than number of ticks
+		// if both are provided. custom tick values need to be within the domain
+		if (userProvidedTickValues !== null) {
```
sophiiae, line 269:
```suggestion 
		if (userProvidedTickValues) {
```

packages/core/src/components/axes/axis.ts
```diff
@@ -51,6 +51,14 @@ export class Axis extends Component {
 			"number"
 		);
 
+		// user can provide custom ticks to be displayed
+		// ticks need to be in the domain of the axis data
+		const userProvidedTickValues = Tools.getProperty(
+			axisOptions,
+			"ticks",
+			"customTicks"
```
theiliad, line 59:
```suggestion
			"values"
```
??

theiliad:
@natashadecoste did you add a demo?

PR #716, opened 2020-07-21
Fix pie chart alignment

sophiiae:
relates to #700


sophiiae:
> have you checked the demos thoroughly? @sophiiae

Yea, I did.

PR #719, opened 2020-07-24
add treemap graph

theiliad:


packages/core/demo/data/treemap.ts
```diff
@@ -0,0 +1,97 @@
+export const treemapData = [
+	{
+		name: "Oceania",
+		children: [
```
cal-smith, line 4: Open question: are arbitrarily nested children allowed? I'm not super familiar with treemap variants, so I don't know if that would be desirable.

packages/core/src/configuration.ts
```diff
@@ -364,6 +364,15 @@ const radarChart: RadarChartOptions = Tools.merge({}, chart, {
 	}
 } as RadarChartOptions);
 
+/**
+ * options specific to treemap charts
+ */
+const treemapChart: RadarChartOptions = Tools.merge({}, chart, {
```
cal-smith, line 370: Should this be `const treemapChart: TreeMapChartOptions`? Or is there sufficient overlap between tree and radar?

cal-smith (commented):
One small comment, and one open question ... otherwise makes sense to me!

packages/core/demo/data/treemap.ts
theiliad: you could provide those children, however we haven't designed for that yet

theiliad:
Designs already cleared with Jeannie, and colors with Sadek

theiliad:
I've already cleared the designs

cal-smith (approved):

sophiiae:
There's an `invariant violation` in React and it's blank in the Angular version.

cal-smith (approved):

natashadecoste:
should we have a minimum tooltip width? Should it change widths **at all**? 
image
image
^ i think these can be the same widths

sophiiae (approved):

natashadecoste:
Why would the user have to toggle **off** the label on the smaller "portions" rather than have the user toggle the label **on** for the bigger/more important slices? Seems theres always going to be more labels turned off than turned on (in most cases). If the user did nothing, theres more a chance the chart comes out looking messed up rather than if the default was not showing labels and the user **intentionally set those**. 
image
i think its rare that all the smaller children would be 2 character strings like this ^ but you can imagine with real names of places or companies there would be lots cut off

packages/core/src/model-treemap.ts
```diff
@@ -0,0 +1,21 @@
...
+// Internal Imports
+import * as Configuration from "./configuration";
+// import { Tools } from "./tools";
+// import * as colorPalettes from "./services/colorPalettes";
+// import { Events, ScaleTypes } from "./interfaces";
+
+// // D3
+// import { map } from "d3-collection";
+// import { scaleOrdinal } from "d3-scale";
+// import { stack } from "d3-shape";
```
natashadecoste, line 13: we dont need yes?

packages/core/src/charts/treemap.ts
```diff
@@ -0,0 +1,44 @@
+// Internal Imports
+import { TreemapChartModel } from "../model-treemap";
+import { Chart } from "../chart";
+import * as Configuration from "../configuration";
+import { ChartConfig, PieChartOptions } from "../interfaces/index";
```
natashadecoste, line 5:
```suggestion
import { ChartConfig, TreemapChartOptions } from "../interfaces/index";
```

packages/core/src/charts/treemap.ts
```diff
@@ -0,0 +1,44 @@
...
+	// the imports below are needed because of typescript bug (error TS4029)
+	Legend,
+	LayoutComponent,
+	Skeleton
+} from "../components/index";
+
+export class TreemapChart extends Chart {
+	model = new TreemapChartModel(this.services);
+
+	constructor(holder: Element, chartConfigs: ChartConfig<PieChartOptions>) {
```
natashadecoste, line 20:
```suggestion
	constructor(holder: Element, chartConfigs: ChartConfig<TreemapChartOptions>) {
```

natashadecoste (changes requested):

theiliad:
> Why would the user have to toggle **off** the label on the smaller "portions" rather than have the user toggle the label **on** for the bigger/more important slices? Seems theres always going to be more labels turned off than turned on (in most cases). If the user did nothing, theres more a chance the chart comes out looking messed up rather than if the default was not showing labels and the user **intentionally set those**.
> image
> 
> i think its rare that all the smaller children would be 2 character strings like this ^ but you can imagine with real names of places or companies there would be lots cut off

if it's a demo issue we can merge as is and discuss that, then update in a later PR

natashadecoste (approved):

sophiiae (approved):

PR #725, opened 2020-07-27
additional item types in the legend

### LINKED ISSUE #714: [Bubble Chart] Radius Label
Radius label in both legend and tooltips

image (3)

sophiiae:
closes #714


packages/core/src/components/essentials/legend.ts
```diff
@@ -94,7 +93,47 @@ export class Legend extends Component {
 			addedLegendItemsText.html((d) => d.name);
 		}
 
-		this.breakItemsIntoLines(addedLegendItems);
+		const radiusLabel = Tools.getProperty(
```
theiliad, line 96: legend probably doesn't need to know about bubble.

2 ways to fix this:
1. A separate class called `BubbleChartLegend` or something of that sort (extends `Legend`), and in there we add the extra label.
2. Legend component accepts `configs` (similar to `Skeleton` - if you look inside `src/charts/scatter.ts` for isntance) and in there we pass in an "extra legend items" array? I'm assuming we'll have more extra legend items in the future so this idea might work better

natashadecoste:
Should it not expose the value for the radius? It doesnt seem to expose the actual value at all..
image
Also i would think that the **group** tooltip information should still be on the bottom .... not sure about when theres a total... 
image
this seems weird for me, whats the ordering supposed to be???

theiliad:
> Should it not expose the value for the radius? It doesnt seem to expose the actual value at all..
> image
> Also i would think that the **group** tooltip information should still be on the bottom .... not sure about when theres a total...
> image
> this seems weird for me, whats the ordering supposed to be???

yes so on the left it'd say `options.bubble.radiusLabel || "Radius"` and on the right it's show the radius value, and this'll only happen on the singular tooltip

packages/core/src/components/essentials/tooltip-axis.ts
```diff
@@ -84,6 +84,15 @@ export class AxisChartsTooltip extends Tooltip {
 					color: this.model.getStrokeColor(datum[groupMapsTo])
 				}
 			];
+
+			// Add radius label and value
+			const { radiusLabel, radiusMapsTo } = Tools.getProperty(options, "bubble");
+			if (radiusLabel) {
```
theiliad, line 90: is there anyway for the `bubble graph` component to shoot out a flag to show the radius label rather than tooltip checking for it?

theiliad: also, we'd show the radius item in tooltip where bubble graphs exist regardless of whether `radiusLabel` has been set. the label would be `radiusLabel || "Radius"`

packages/core/src/chart.ts
```diff
@@ -126,7 +126,7 @@ export class Chart {
 
 		const legendComponent = {
 			id: "legend",
-			components: [new Legend(this.model, this.services)],
+			components: [new Legend(this.model, this.services, [])],
```
theiliad, line 129: what's the empty array for?

packages/core/src/components/essentials/legend.ts
```diff
@@ -94,7 +94,45 @@ export class Legend extends Component {
 			addedLegendItemsText.html((d) => d.name);
 		}
 
-		this.breakItemsIntoLines(addedLegendItems);
+		const radiusLabel = this.configs.find(item =>
```
theiliad, line 97: configs should be an object, so that we can add more configs in the future to `legend` if wee need to. And the point of adding an `extraLabels` flag to `Legend` would be to remove any `radius` specific logic from `Legend`... You still have everything specific to radius in here

packages/core/src/components/essentials/tooltip-axis.ts
sophiiae: > also, we'd show the radius item in tooltip where bubble graphs exist regardless of whether `radiusLabel` has been set. the label would be `radiusLabel || "Radius"`

`radiusLabel` has a default value `Radius`, so the label would be `Radius` when it's not defined in the options.

sophiiae: > is there anyway for the `bubble graph` component to shoot out a flag to show the radius label rather than tooltip checking for it?
not sure how ..

sophiiae: > is there anyway for the `bubble graph` component to shoot out a flag to show the radius label rather than tooltip checking for it?

not sure how ..

theiliad: > not sure how ..

you'd need to look at how any of the graphs trigger the tooltip to show

packages/core/src/axis-chart.ts
```diff
@@ -65,9 +66,30 @@ export class AxisChart extends Chart {
...
+		if (radiusLabel) {
+			extraLabels.push({
+				type: "radius-label",
+				icon: `<svg width="16px" height="16px" viewBox="0 0 16 16" version="1.1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink">
+							<title>Artboard</title>
+							<g id="Artboard" stroke="none" stroke-width="1" fill="none" fill-rule="evenodd">
+								<circle id="Oval-Copy-33" stroke="#8C8C8C" cx="7" cy="7" r="6.5"></circle>
+								<circle id="Oval-Copy-33" stroke="#8C8C8C" cx="7" cy="10" r="3.5"></circle>
+							</g>
+						</svg>`,
```
natashadecoste, line 85: can we clean this up a bit? 

1. <title> seems to have been carried over from the design file but its not being used correctly here 
2. do we need IDs on the <g> and <circle>? are we using the IDs? and if we are, im sure we could find better names 
3. can we remove some styles that are applied in attributes to the scss? especially the HEX color, stroke widths and stuff?

packages/core/src/components/graphs/bubble.ts
```diff
@@ -10,10 +10,16 @@ import { scaleLinear } from "d3-scale";
...
+	radiusMapsTo;
 
 	getRadiusScale(selection: Selection<any, any, any, any>) {
 		const options = this.model.getOptions();
-		const { radiusMapsTo } = options.bubble;
+		const { radiusMapsTo, radiusLabel } = options.bubble;
+		
+		// Save static variables that other components can access
+		this.radiusLabel = radiusLabel;
+		this.radiusMapsTo = radiusMapsTo;
```
theiliad, line 22: why do we need these stored in `this`?

packages/core/src/components/essentials/tooltip-axis.ts
```diff
@@ -84,6 +84,14 @@ export class AxisChartsTooltip extends Tooltip {
 					color: this.model.getStrokeColor(datum[groupMapsTo])
 				}
 			];
+
+			// Add radius label and value
+			if (e.detail.radiusLabel) {
+				items.push({
+					label: e.detail.radiusLabel,
+					value: datum[e.detail.radiusMapsTo] 
+				})
```
theiliad, line 93: seems like if `getItems()` supports an `additionalItems` array, then bubble can provide an additional radius value to it instead of tooltip knowing about bubble

packages/core/src/components/graphs/bubble.ts
sophiiae: > why do we need these stored in `this`?
I stored them in `this` so it can be accessed in `scatter.ts`. I can't get them in `scatter.ts`

theiliad: @sophiiae this is something we should just grab off of `this.model.getOptions()` rather than storing here

packages/core/src/styles/components/_legend.scss
```diff
@@ -36,4 +36,15 @@ g.#{$prefix}--#{$charts-prefix}--legend {
 	&.clickable g.legend-item:hover {
 		cursor: pointer;
 	}
+
+	g.extra-label {
+		g.icon > svg > g {
```
natashadecoste, line 41:
```suggestion
	g.extra-label g.icon > svg > g {
```

packages/core/src/styles/components/_legend.scss
```diff
@@ -36,4 +36,15 @@ g.#{$prefix}--#{$charts-prefix}--legend {
...
+			stroke: none;
+			stroke-width: 1;
+
+			circle {
+				stroke: #8C8C8C;
```
natashadecoste, line 46: carbon token? 🤔

packages/core/src/components/essentials/legend.ts
```diff
@@ -123,7 +151,8 @@ export class Legend extends Component {
 			this.addEventListeners();
 		}
 
-		const alignment = Tools.getProperty(legendOptions, "alignment");
+		// Set alignment for legend
+		const alignment = Tools.getProperty(legendOptions,"alignment");
```
natashadecoste, line 155:
```suggestion
		const alignment = Tools.getProperty(legendOptions, "alignment");
```

packages/core/src/styles/components/_legend.scss
sophiiae: can't find it in carbon colors.

natashadecoste (approved):

packages/core/src/axis-chart.ts
```diff
@@ -38,14 +39,15 @@ export class AxisChart extends Chart {
 	}
 
 	protected getAxisChartComponents(graphFrameComponents: any[]) {
+		const options = this.model.getOptions();
 		const isZoomBarEnabled = Tools.getProperty(
-			this.model.getOptions(),
-			'zoomBar',
```
natashadecoste, line 43: you might have to try linting? it **should** be changing the double quotes to single quotes now ...

packages/core/src/components/graphs/scatter.ts
```diff
@@ -365,10 +365,16 @@ export class Scatter extends Component {
 					})
 					.classed('unfilled', false);
 
-				// Show tooltip for single datapoint
+				// Show tooltip
+				const bubbleOptions = Tools.getProperty(self.model.getOptions(), "bubble");
```
natashadecoste, line 369:
```suggestion
				const bubbleOptions = Tools.getProperty(self.getOptions(), "bubble");
```

natashadecoste (changes requested):

packages/core/src/axis-chart.ts
```diff
@@ -113,9 +115,36 @@ export class AxisChart extends Chart {
...
+			extraLabels.push({
+				type: "radius-label",
+				icon: `<svg width="16px" height="16px" viewBox="0 0 16 16">
+							<g fill="none" fill-rule="evenodd">
+								<circle cx="7" cy="7" r=${outerRadius}></circle>
+								<circle cx="7" cy="10" r=${innerRadius}></circle>
+							</g>
+						</svg>`,
+				text: radiusLabel
+			});
```
theiliad, line 142: So why would this not be done through the charting options? and we can set the default value in bubble charts through `configuration.ts`

theiliad: maybe @jeanservaas should chime in here. What we're wondering is whether only the library should be able to add additional legend items internally, or whether we'd like to also extend that capability to the users of the library?

theiliad (changes requested):

packages/core/src/axis-chart.ts
jeanservaas: @theiliad my gut is to extend the capability to the users of the library, what do you think?

theiliad: Yes agreed as long as we have clear guidelines and users aren't overusing the feature

theiliad (changes requested):
This needs to become a `add the ability for the legend to take in additional items` rather than `Add radius label for bubble chart` and as part of it we can add the radius bubble to the bubble charts

packages/core/src/axis-chart.ts
```diff
@@ -113,9 +111,22 @@ export class AxisChart extends Chart {
...
+		// Add extra legend items
+		const radiusLabel = Tools.getProperty(options, 'bubble', 'radiusLabel');
+		const additionalLabels = radiusLabel
+			? [
+					{
+						...bubbleConfig.radiusLabel,
+						text: radiusLabel,
+					},
+			  ]
+			: [];
```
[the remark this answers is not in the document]
sophiiae: The thing is, the text of the radius label is customizable. We have two bubble demos, one with `radiusLabel: Surplus` and another `radiusLabel: Radius`. I'm not sure if we should keep this option open.

theiliad:
> This needs to become a `add the ability for the legend to take in additional items` rather than `Add radius label for bubble chart` and as part of it we can add the radius bubble to the bubble charts

additionally, we need to be able to add the additional items anywhere I think. In this example, there are no filterable legend items, however we've got many `informational` legend items in place

image

packages/core/src/components/graphs/scatter.ts
```diff
@@ -365,10 +365,30 @@ export class Scatter extends Component {
...
+							value:
+								datum[
+									Tools.getProperty(
+										bubbleOptions,
+										'radiusMapsTo'
+									)
+								],
+						},
+					],
 				});
```
theiliad, line 392: if this only relates to bubble, then this should happen inside `bubble.ts` not `scatter.ts`

packages/core/src/axis-chart.ts
```diff
@@ -117,9 +115,32 @@ export class AxisChart extends Chart {
...
+		// Add extra legend items
+		const radiusLabel = Tools.getProperty(options, 'bubble', 'radiusLabel');
+		let additionalItems = Tools.getProperty(
+			options,
+			'legend',
+			'additionalItems'
+		);
+		additionalItems.forEach(
+			(item) => (item.icon = Tools.convertSVGString(item.icon))
+		);
```
theiliad, line 127: why is `axis-chart` doing this and not `legend`?

theiliad (changes requested):

packages/core/src/components/essentials/legend.ts
```diff
@@ -132,7 +123,36 @@ export class Legend extends Component {
...
+
+			const addedAdditionalLabelItems = additionalLabelItems
+				.enter()
+				.append('g')
+				.classed('additional-label', true);
+
+			addedAdditionalLabelItems
+				.append('g')
+				.classed('icon', true)
+				.html((d) => d.icon);
```
natashadecoste, line 145: the user would have to add their own aria-label and role to the icon html but maybe add it to the demo code as best practice

natashadecoste (changes requested):

packages/core/src/components/graphs/scatter.ts
sophiiae: `bubble.ts` doesn't have `addEventListeners()`, in order to put it in `bubble.ts` I need to override the function by copying the whole content and add `additionalItems` there.

packages/core/src/components/essentials/legend.ts
```diff
@@ -111,28 +83,73 @@ export class Legend extends Component {
...
+
+		// Update radius label text with option
+		const radiusLabel = Tools.getProperty(
+			this.getOptions(),
+			'bubble',
+			'radiusLabel'
+		);
+		if (radiusLabel) {
+			additionalItemsOption[0].name = radiusLabel;
 		}
```
natashadecoste, line 117: this wont make sense because other charts might use the radius label (ex circle pack)

natashadecoste (changes requested):
A few things:
1. you can override _some of the colors_ for the additional legend items, but not all.. i think we should have the defaults but i think for the most part they should work the same way. `Size` and `Radius` are pretty easy to apply color similarly to `Line`.
2. BubbleChart gets an additional item _by default_ but the legend looks for bubble chart options (see comment in legend.ts) -> this will get messy if we have many charts and decide that legend items are added by default. ... maybe in bubble chart model you could have some code that adds the additional legend item if a user is not overriding it (but allow user to overide it for color options/etc). That way, legend will handle the additional items all in the same way. 
3. I think before we merge,  the demos should only be showing patterns that make sense so we should cleanup the simple bar demo

natashadecoste (dismissed):

packages/core/src/configuration-non-customizable.ts
```diff
@@ -57,10 +57,41 @@ export const legend = {
...
+	size: {
+		config: [
+			{ width: 23, height: 12 },
+			{ width: 13, height: 6 },
+		],
+		color: '#8D8D8D',
+	},
+	quartile: {
+		config: [
+			{ x: 0, y: 0, width: 24, height: 13, color: '#6929C4' },
```
theiliad, line 92: the quartile item type color should match the neutral range background in the boxplot
image

theiliad (changes requested):

packages/core/demo/data/bubble.ts
```diff
@@ -31,6 +31,14 @@ export const bubbleDoubleLinearOptions = {
...
 		radiusLabel: 'Surplus',
 	},
+	legend: {
+		additionalItems: [
+			{
+				type: 'radius',
+				name: 'Surplus',
+			}
+		]
+	}
```
theiliad, line 41: this is redundant, the demo works the same way without it

theiliad (changes requested):

packages/core/demo/data/bubble.ts
sophiiae: This will override the default radius label item with the `name: Surplus`.

theiliad: 👍🏻

natashadecoste:
my only question: for the tooltip - should we keep the group label at the bottom? 
its at the bottom for charts like bar
image
but at the top for charts like treemap 
image


but do we ever want the group label and color indicator the middle? Looks weird in my opinion
image

jeanservaas:
@natashadecoste no we don't want the group label in the middle, should be either at the top or the bottom... I don't think we necessarily have to mandate having it at the bottom...I think there are cases where the top makes sense, but the bottom is probably the most common.

natashadecoste (dismissed):

packages/core/src/tools.ts
```diff
@@ -269,6 +274,47 @@ export namespace Tools {
...
+	/**
+	 * Update legend additional items
+	 * @param {any} defaultOptions
+	 * @param {any} providedOptions
+	 */
+	export function updateLegendAdditionalItems(
+		defaultOptions,
+		providedOptions
+	) {
+		let defaultLegendItems = Tools.getProperty(
```
theiliad, line 286:
```suggestion
		let defaultAdditionalItems = Tools.getProperty(
```

packages/core/src/tools.ts
```diff
@@ -269,6 +274,47 @@ export namespace Tools {
...
+			defaultOptions,
+			'legend',
+			'additionalItems'
+		);
+		const providedLegendItems = Tools.getProperty(
```
theiliad, line 291:
```suggestion
		const userProvidedAdditionalItems = Tools.getProperty(
```

packages/core/src/tools.ts
```diff
@@ -269,6 +274,47 @@ export namespace Tools {
...
+					item.type === LegendItemType.RADIUS
+			);
+
+			providedOptions.legend.additionalItems = Tools.unionBy(
+				defaultLegendItems,
+				providedLegendItems,
+				'name'
+			);
+
+			defaultOptions.legend.additionalItems = [];
```
theiliad, line 314: I think it'd be more readable to call them user-provided rather than just provided

packages/core/src/configuration-non-customizable.ts
```diff
@@ -57,10 +57,41 @@ export const legend = {
...
 		verticalSpace: 24,
 		textYOffset: 8,
+		spaceAfter: 4,
 	},
 	checkbox: {
 		radius: 6.5,
-		spaceAfter: 4,
+	},
+	radius: {
+		config: [
```
theiliad, line 66: what does `config` do? What does it describe? could any of these fields be more detailed?

theiliad (changes requested):

theiliad:
also @sophiiae, could you please add a demo under the `legend` demo group for this?

packages/core/src/configuration-non-customizable.ts
sophiiae: it provides the config data for the icon. I changed it to `iconData` since it's an array.

theiliad (approved):

natashadecoste (approved):

Issue #568, opened 2020-04-09
Error states


caesarsol:
This issue to keep track of the reasoning behind the logic of creating chart error states.
It has been extracted from #407, it was decided to build these on top on Skeleton and Empty states, once these has been finalized.

Design as of now:

![error states](https://user-images.githubusercontent.com/1799710/78885788-78798b00-7a5d-11ea-8cfc-113ceac9d2d1.jpg)

PR #727, opened 2020-07-28
feat: add skeleton error states

ryanomackey:
### Updates
- Adds error states for components with skeletons (see #568)

- [ ] Demos all features
- [ ] Documented/annotated
- [ ] Matches UI/UX specs
- [ ] Accessible


packages/core/src/components/graphs/skeleton.ts
```diff
@@ -301,4 +316,113 @@ export class Skeleton extends Component {
...
+
+		DOMUtils.appendOrSelect(
+			errorContainer,
+			"circle.bx--cc--error-message__icon"
+		);
+
+		errorIcon.exit().remove();
+
+		// Exclamation point
+
```
natashadecoste, line 360:
```suggestion
```

ryanomackey:
> when i hover on orange status icon, it turns purple 🟣

Ha. I almost hate to fix that.

packages/core/demo/data/bar.ts
```diff
@@ -478,6 +478,25 @@ export const simpleBarSkeletonOptions = {
...
+		left: {},
+		bottom: {
+			scaleType: "labels"
+		}
+	},
+	data: {
+		loading: false,
+		error: {
+			title: "No data available",
+			subtitle: "Lorem ipsum dolor sit."
```
theiliad, line 495: Maybe we should use a real subtitle here to make for a better demo? e.g. `Unable to reach application servers at this time`

packages/core/demo/data/line.ts
```diff
@@ -422,3 +422,29 @@ export const lineSkeletonOptions = {
 		loading: true
 	}
 };
+
+// line - error
+export const lineErrorData = [];
```
theiliad, line 427: What happens if data is provided? would it still show the error?

packages/core/src/components/graphs/skeleton.ts
```diff
@@ -28,12 +28,27 @@ export class Skeleton extends Component {
...
+			this.renderSkeleton(false);
+			this.renderErrorMesssage(error);
+		}
+
+		// if neither are true, clean up
+		if (!isDataLoading && !error) {
 			this.removeSkeleton();
+			this.removeErrorMessage();
 		}
 	}
```
theiliad, line 53: shouldn't the error be shown at all times regardless of whether the chart is loading or not? Errors probably won't always just be contained to the data, they could be about anything, and if you don't want an error to show you'd probably just remove it from options...?

packages/core/src/components/graphs/skeleton.ts
```diff
@@ -301,4 +316,113 @@ export class Skeleton extends Component {
...
+			"circle.bx--cc--error-message__icon"
+		);
+
+		errorIcon.exit().remove();
+
+		// Exclamation point
+
+		const exclamationPoint = errorContainer
+			.selectAll("text.bx--cc--error-message__exclamation-point")
+			.data(["!"]);
```
theiliad, line 363: we should be using the correct `carbon-icon` here rather than an `!`

packages/core/src/components/graphs/skeleton.ts
```diff
@@ -301,4 +316,113 @@ export class Skeleton extends Component {
...
+			.selectAll("text.bx--cc--error-message__subtitle")
+			.data([error.subtitle]);
+
+		errorSubtitle
+			.enter()
+			.append("text")
+			.classed("bx--cc--error-message__subtitle", true)
+			.merge(errorSubtitle)
+			.attr("x", "2.65em")
+			.attr("y", height / 2 + 18)
```
theiliad, line 413: what does `/ 2 + 18` do?

packages/core/src/components/graphs/skeleton.ts
```diff
@@ -301,4 +316,113 @@ export class Skeleton extends Component {
...
+		const errorTitle = errorContainer
+			.selectAll("text.bx--cc--error-message__title")
+			.data([error.title]);
+
+		errorTitle
+			.enter()
+			.append("text")
+			.classed("bx--cc--error-message__title", true)
+			.merge(errorTitle)
+			.attr("x", "2.25em")
```
theiliad, line 391: what's `2.25em`?

packages/core/src/components/graphs/skeleton.ts
```diff
@@ -301,4 +316,113 @@ export class Skeleton extends Component {
...
+			.attr("width", width)
+			.attr("height", height);
+
+		// Error icon
+		const errorIcon = errorContainer
+			.selectAll("circle.bx--cc--error-message__icon")
+			.data([
+				{
+					cx: "1em",
+					cy: height / 2 - 5,
```
theiliad, line 338: what does `/ 2 - 5` do?

theiliad:
Also, I'm still seeing the purple hover state in the demos
image

theiliad:
Tyler or Jeannie, could you please review? Do we just need to support error states here or should we support other icons & statuses for other types of messages?

loganmccaul:
@jeanservaas could you take a look at @theiliad's questions?

@ryanomackey could you respond to some of @theiliad's questions?

theiliad:
> @jeanservaas could you take a look at @theiliad's questions?
> 
> @ryanomackey could you respond to some of @theiliad's questions?

I've just msged Jeannie to review. I think overall I'd like to know if we should add support for more icons or other error styling that we might need.

Being able to also add an `action` might be useful, most use cases that I've seen want to show a button (e.g. no activities to visualize right now => `upload some activity data`)

jeanservaas:
@theiliad I would def just start with support icons here. Eventually we want to create a more comprehensive set of status icons (there's def and issue somewhere in the carbon backlog) but it hasn't moved forward yet.

theiliad (changes requested):
Let's fix the react & angular issues and we can do a final code review 💯

PR #728, opened 2020-07-29
Add legend order option

### LINKED ISSUE #587: Donut chart angular - Need support for showing legend in user defined sequence
#### I am submitting a...

- [ ] Feature request
   Need support for showing legend in user defined sequence in Donut chart.

#### Charts version:
    v0.30.14

#### Current behavior
    Right now showing legend in descending order of value.

#### Expected behavior
    Legend must be in user defined sequence.

@theiliad

sophiiae:
closes #587


packages/core/src/model.ts
```diff
@@ -116,6 +114,19 @@ export class ChartModel {
...
+
+		// Sort data in user defined order
+		const legendOrder = Tools.getProperty(
+			this.getOptions(),
+			"legend",
+			"order"
+		);
+		if (legendOrder.length) {
+			sanitizedData.sort((dataA, dataB) => 
+				legendOrder.indexOf(dataA.group) - legendOrder.indexOf(dataB.group)
```
Donisius, line 126: Will this work/should it account for duplicate labels? Like let's say `legendOrder` is `["Restocking", "Misc","Sold", "Qty", "More", "Restocking"]`.

theiliad: I don't think we need to worry about duplicates in the legend, but interested to see what order you'd get from that array

sophiiae: I just tested. It doesn't make any difference with duplicates.

packages/core/src/model.ts
```diff
@@ -116,6 +114,19 @@ export class ChartModel {
...
+			);
+		}
```
theiliad, line 128: this seems to be ordering the entire data instead of just the legend ordering. Hadn't thought about it this way. What will change now? Vs. only changing legend order...?

sophiiae: if we only change the legend order, the chart remains old order, it looks weird for me. 
For example, the legend is `Sold, Misc, Qty, Restocking, More`, 
and the bars display as `Qty, More, Sold, Restocking, Misc`. 
I think they should always keep the same ordering.

theiliad: I'd ask the user who created #587. I think they really just wanted to reverse the order of the legend

sophiiae: ok, I'll change it.

packages/core/demo/data/bar.ts
```diff
@@ -93,6 +93,22 @@ export const simpleBarOptions = {
...
+		left: {
+			mapsTo: "value"
+		},
+		bottom: {
+			mapsTo: "group",
+			scaleType: "labels"
+		}
+	},
+	legend: {
+		order: ["Restocking", "Misc","Sold", "Qty", "More"]
```
theiliad, line 108: what if I pass in `order: ["Restocking", "Misc"]`, what legend items would I get then?

packages/core/src/configuration.ts
```diff
@@ -64,7 +64,8 @@ export const legend: LegendOptions = {
 		spaceAfter: 4
 	},
 	truncation: standardTruncationOptions,
-	alignment: Alignments.LEFT
+	alignment: Alignments.LEFT,
+	order: []
```
theiliad, line 68: should be null

theiliad:
```suggestion
	order: null
```

packages/core/demo/data/bar.ts
natashadecoste: i would expect it to order the legend "Restocking", "Misc" and then **all others** in the default order... 
but it does:  **all others** in the default order, then it does the custom ordered labels "Restocking", "Misc"

sophiiae: I thought about this, but not sure if we need to take this into account.

packages/core/src/components/essentials/legend.ts
```diff
@@ -22,9 +22,28 @@ export class Legend extends Component {
 		);
 		const options = this.model.getOptions();
 		const legendOptions = Tools.getProperty(options, "legend");
+		let dataGroups = this.model.getDataGroups();
+
+		// Sort data in user defined order
+		const legendOrder = Tools.getProperty(legendOptions, "order");
```
theiliad, line 28: separate function (method) maybe?

natashadecoste:
@sophiiae it seems that when you change the options for order, it does not update and change the legend

sophiiae:
> @sophiiae it seems that when you change the options for order, it does not update and change the legend

Do you mean in the demo? It updates when I modify the legend order.

PR #732, opened 2020-07-31
fix: fix tooltip position with zoom bar enabled

hlyang397:
### Updates
- Fix the wrong ruler tooltip y position with zoom bar enabled
  - Add zoom bar height to tooltip y position if zoom bar is enabled

before:
image

after:
image

- [ ] Demos all features
- [ ] Documented/annotated
- [ ] Matches UI/UX specs
- [ ] Accessible

### DIFF
--- a/packages/core/src/components/essentials/tooltip.ts
+++ b/packages/core/src/components/essentials/tooltip.ts
@@ -208,7 +208,7 @@ export class Tooltip extends Component {
 	positionTooltip(e: CustomEvent) {
 		const holder = this.services.domUtils.getHolder();
 		const target = this.tooltip.node();
-		const isZoomBarEnabled = Tools.getProperty(
+		const isTopZoomBarEnabled = Tools.getProperty(
 			this.model.getOptions(),
 			"zoomBar",
 			"top",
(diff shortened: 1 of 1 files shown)


packages/core/src/components/essentials/tooltip.ts
```diff
@@ -207,10 +208,26 @@ export class Tooltip extends Component {
...
+			"enabled"
+		);
 
 		let mouseRelativePos = Tools.getProperty(e, "detail", "mousePosition");
 		if (!mouseRelativePos) {
 			mouseRelativePos = mouse(holder);
+		} else {
+			// if the mouse position is from event (ruler)
+			// we need add zoom bar height
+			if (isZoomBarEnabled) {
```
theiliad, line 224: what if zoom bar gets positioned in the bottom of the chart?

hlyang397: > what if zoom bar gets positioned in the bottom of the chart?
The `isZoomBarEnabled` is actually checking the top zoom bar.
So it won't change y position if zoom bar is in the bottom of the chart.

PR #734, opened 2020-08-04
Combo chart

### LINKED ISSUE #3: Combo Chart
image

### LINKED ISSUE #454: Compound Chart containing both line chart and bar chart
Thia is a  new design/functionality request

charts version: "@carbon/charts-angular": "^0.16.25",

We need a single compound chart component which has both line chart and bar chart for this functionality:

Screen Shot 2019-12-19 at 3 31 55 PM

### LINKED ISSUE #527: Combo Chart
#### I am submitting a...

- [X] Feature request
- [ ] Design defect
- [ ] Source code defect
- [ ] Other

### Issue description

Looking to produce a combo-chart (i.e. a single chart that displays data series in 2 different formats within the same single chart)

The secondary Y-axis independence would be an important feature to maintain ratios amongst the series.

Ideally too our use case allows a stacked bar chart as a percentage

### Screenshot or recording

image

natashadecoste:
#### Chart types available to use in combo charts:
- Line
- Scatter
- Simple Bar
- Stacked Bar
- Grouped Bar
- Area
- Stacked Area

- [ ] Demos all features
- [ ] Documented/annotated
- [ ] Matches UI/UX specs
- [ ] Accessible

closes #454 
closes #527 
closes #3


theiliad:
1. Are there no rulers in combo? and which combinations currently work?
2. I'm assuming that this wouldn't be accessible
image

packages/core/demo/data/combo.ts
```diff
@@ -0,0 +1,130 @@
...
+	},
+	chartTypes: {
+		"simple-bar": [
+			"Qty",
+			"More",
+			"Sold",
+			"Restocking"
+		],
+		"line": [
+			"Line"
```
theiliad, line 33: maybe a better name for the line, `trend` maybe?

packages/core/src/charts/combo.ts
```diff
@@ -0,0 +1,99 @@
...
+		// If the stacked-bar chart is included, stack its data groups
+		if (graphs.includes("stacked-bar")) {
+			this.model.setStackedGroups(chartTypes["stacked-bar"]);
+		}
+
+		// Initialize data, services, components etc.
+		this.init(holder, chartConfigs);
+	}
+
+	getGraphComponents(graph: string) {
```
theiliad, line 75: we need a better more descriptive name here

packages/core/src/charts/combo.ts
```diff
@@ -0,0 +1,99 @@
...
+		const { chartTypes } = this.model.getOptions();
+		return graphComponentsMap[graph].map(Component => new Component(this.model, this.services, {groups: chartTypes[graph]}));
+	}
+
+	getGraphsComponents() {
```
theiliad, line 80: we need a better more descriptive name here

packages/core/src/charts/combo.ts
```diff
@@ -0,0 +1,99 @@
...
+
+export class ComboChart extends AxisChart {
+	model = new ChartModel(this.services);
+
+	constructor(holder: Element, chartConfigs: ChartConfig<ComboChartOptions>) {
+		super(holder, chartConfigs);
+
+		// Warn user if no chartTypes defined
+		if (!chartConfigs.options.chartTypes) {
+			console.warn("No chartTypes defined for the Combo Chart!");
```
theiliad, line 43: so what do we render at that point?

packages/core/src/tools.ts
```diff
@@ -72,8 +76,13 @@ export namespace Tools {
 		providedOptions: any
 	) {
 		defaultOptions = Tools.clone(defaultOptions);
+		const providedChartTypes = providedOptions.chartTypes;
 		const providedAxesNames = Object.keys(providedOptions.axes || {});
 
+		if (providedChartTypes) {
+			defaultOptions.chartTypes = providedChartTypes;
+		}
```
theiliad, line 84: why do we need this here? wouldn't these move over with the `Tools.merge()` call?

packages/core/src/model.ts
```diff
@@ -75,30 +77,34 @@ export class ChartModel {
...
+				(dataGroup) => dataGroup.name === datum[groupMapsTo] && dataGroup.status === ACTIVE
 			);
 		});
 	}
 
-	getDisplayData() {
-		if (!this.get("data")) {
-			return null;
-		}
+	// getDisplayData(groups) {
```
theiliad, line 89: ?

packages/core/src/tools.ts
natashadecoste: there was some weird things happening here under the hood, i changed it quite a bit (here and in 2 other places)

theiliad (changes requested):
combotransitions

is there a lag in the scatter & line transitions?

theiliad (changes requested):
image

the 0 line is irrelevant to the line here and should be removed. It corresponds to an axis that's not visible here

theiliad (changes requested):

jeanservaas:
Hey @natashadecoste 

agree with @theiliad that the dots on the line chart are lagging during category transitions and even toggling between combo chart variants in story book the dots are doing a weird little slide to the right in safari.

One other thing, and this may not be specific to the common chart @theiliad but I'm noticing the y-axis titles are two tight / close to the y-axis labels. They should be 8px away from them on both axes.

packages/core/src/charts/combo.ts
natashadecoste: there was a `getGraphsComponents` and `getGraphComponents` but i changed it so theres just this one so that should be more clear. Do you still want the naming changed more or was that the issue?

natashadecoste:
@jeanservaas  I opened a separate issue for the axes titles: #859 
because it can probably be fixed in parallel to this PR

packages/core/src/services/scales-cartesian.ts
```diff
@@ -319,12 +403,15 @@ export class CartesianScales extends Service {
...
 		const axesOptions = Tools.getProperty(options, "axes");
+		const dualAxes = this.isDualAxes();
 
 		// If top axis has been specified as `main`
 		if (
 			(Tools.getProperty(axesOptions, AxisPositions.BOTTOM) === null &&
 				Tools.getProperty(axesOptions, AxisPositions.TOP) !== null) ||
-			Tools.getProperty(axesOptions, AxisPositions.TOP, "main") === true
+			Tools.getProperty(axesOptions, AxisPositions.TOP, "main") === true ||
+			dualAxes && Tools.getProperty(axesOptions, AxisPositions.BOTTOM, "datasets")
```
theiliad, line 413: couple of points here:
1. `dualAxes && Tools.getProperty(axesOptions, AxisPositions.BOTTOM, "datasets")` should be in brackets
2. What if `datasets` exists & is an empty array?

packages/core/src/services/scales-cartesian.ts
```diff
@@ -118,6 +142,64 @@ export class CartesianScales extends Service {
...
+		}
+		return this.rangeAxisPosition;
+	}
+
+
+	/**
+	 * Returns the position of the domain axis that the datum is mapped with
+	 * @param datum
+	 */
+	getDualAxesDomainPosition(datum) {
```
theiliad, line 192: doesn't this work regardless of whether you have a dual-axis setup or not? Maybe it should just be called `getDatumCorrespondingAxisPosition` or something like that

packages/core/src/services/scales-cartesian.ts
```diff
@@ -118,6 +142,64 @@ export class CartesianScales extends Service {
...
+	// it does not count as dual axes if it just has another axis turned on but is not actually using it to map a dataset
+	determineDualAxes() {
+		const options = this.model.getOptions();
+		const axesOptions = Tools.getProperty(options, "axes");
+
+		if (axesOptions[AxisPositions.LEFT]?.datasets  &&  axesOptions[AxisPositions.RIGHT]
+			|| axesOptions[AxisPositions.RIGHT]?.datasets  &&  axesOptions[AxisPositions.LEFT]
+			|| axesOptions[AxisPositions.TOP]?.datasets  &&  axesOptions[AxisPositions.BOTTOM]
+			|| axesOptions[AxisPositions.BOTTOM]?.datasets  &&  axesOptions[AxisPositions.TOP]) {
+				this.dualAxes = true;
```
theiliad, line 159: is it `this.dualAxes` or `this.isDualAxes`?

Also, what about `determineAxisDuality` for the name?

packages/core/src/services/scales-cartesian.ts
theiliad: maybe I'm misunderstanding what this method is doing, regardless the name is confusing

packages/core/src/model.ts
```diff
@@ -240,12 +267,12 @@ export class ChartModel {
 		}) as any;
 	}
 
-	getStackedData({ percentage } = { percentage: false }) {
+	getStackedData(groups?: string[], { percentage } = { percentage: false }) {
```
theiliad, line 270: `groups` should be part of the original parameter in the function like `percentage`

theiliad: when the function is called we'd want to see `getStackedData({ groups: blah, percentage: blah });`

rather than `getStackedData(blah, blah)`

packages/core/src/model.ts
```diff
@@ -86,23 +84,33 @@ export class ChartModel {
...
 		const { ACTIVE } = Configuration.legend.items.status;
-		const dataGroups = this.getDataGroups();
+		const dataGroups = this.getDataGroups(groups);
 		const { groupMapsTo } = this.getOptions().data;
 
-		const allDataFromDomain = this.getAllDataFromDomain();
+		let allDataFromDomain = this.getAllDataFromDomain();
+
+		if (groups) {
+			allDataFromDomain = allDataFromDomain.filter(item => {
```
theiliad, line 110: shouldn't `getAllDataFromDomain` just take in the `groups`?

packages/core/src/components/axes/zero-line.ts
```diff
@@ -7,35 +7,43 @@ export class ZeroLine extends Component {
 	type = "zero-line";
 
 	render(animate: boolean) {
+		let axisPosition = this.services.cartesianScales.getRangeAxisPosition();
+		if (this.services.cartesianScales.isDualAxes()) {
+			// get the axis that the groups use
+			axisPosition = this.services.cartesianScales.getDualAxesRangePosition(null, this.configs.groups);
```
theiliad, line 13: we don't want to see that `null` in there. Each function ideally would just take a `configs` parameter which'd be a configs object, that'd be way more readable

packages/core/src/components/graphs/area-stacked.ts
```diff
@@ -64,7 +60,7 @@ export class StackedArea extends Component {
 		// D3 area generator function
 		this.areaGenerator = area()
 			// @ts-ignore
-			.x((d) => mainXScale(new Date(d.data.sharedStackKey)))
+			.x((d) => mainXScale(isTimeSeries ? new Date(d.data.sharedStackKey) : d.data.sharedStackKey))
```
theiliad, line 63: please test with just `d.data.sharedStackKey` and see whether this still works.

With the sanitize PR I just merged that should be good enough

packages/core/src/components/graphs/scatter.ts
```diff
@@ -94,7 +93,10 @@ export class Scatter extends Component {
 			.selectAll("circle.dot")
 			.data(
 				scatterData,
-				(datum) => `${datum[groupMapsTo]}-${datum[domainIdentifier]}`
+				(datum) => {
+					const domainIdentifier = this.services.cartesianScales.getDomainIdentifier(datum);
+					return `${datum[groupMapsTo]}-${datum[domainIdentifier]}`;
+				}
```
theiliad, line 99: this might make things a bit slow, wondering if there's a better way to do this since if you have 1000 datapoints this'll run 1000 times

theiliad (changes requested):

packages/core/demo/data/combo.ts
```diff
@@ -0,0 +1,381 @@
...
+		},
+		right: {
+			title: "Temperature (°C)",
+			mapsTo: "temp",
+			datasets: [
+				"Temperature"
+			]
+		}
+	},
+	chartTypes: {
```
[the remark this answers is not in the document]
natashadecoste: will do

sophiiae:
Screen Shot 2020-11-12 at 6 12 47 PM
Screen Shot 2020-11-12 at 6 13 04 PM

Should we hide the irrelevant axis title as well?

natashadecoste:
> Screen Shot 2020-11-12 at 6 12 47 PM
> Screen Shot 2020-11-12 at 6 13 04 PM
> 
> Should we hide the irrelevant axis title as well?

On the dataviz call it was determined that this is ok for combo (and possibly helpful), we keep the titles but hide the axis tick values

packages/core/src/model.ts
natashadecoste: oh yes, good note - thanks 👍

packages/react/src/combo-chart.js
```diff
@@ -0,0 +1,25 @@
+import React from "react";
+
+import { ComboChart as PC } from "@carbon/charts";
```
theiliad, line 3:
```suggestion
import { ComboChart as CC } from "@carbon/charts";
```

packages/core/src/charts/combo.ts
```diff
@@ -0,0 +1,122 @@
...
+				return graphComponentsMap[graph.type].map((Component, i) =>
+					new Component(this.model, this.services, { groups: graph.datasets, id: counter++, options: options }));
+			} else {
+				// user has imported a type or custom component to instantiate
+				options = Tools.merge({}, this.model.getOptions(), graph.options);
+				return new type(this.model, this.services, { groups: graph.datasets, id: counter++, options: options });
+			}
+		}).filter(item => item !== null);
+
+		console.log(Tools.flatten(graphComponents));
```
theiliad, line 95: do we need this?

packages/core/src/tools.ts
```diff
@@ -8,10 +8,12 @@ import {
 
 import {
 	debounce as lodashDebounce,
-	merge as lodashMerge,
+	merge as 	lodashMerge,
```
theiliad, line 11: did you add this or did prettier? I think you might've accidentally, let's remove it

image

packages/core/src/services/scales-cartesian.ts
```diff
@@ -239,26 +303,24 @@ export class CartesianScales extends Service {
...
 
-	getRangeIdentifier() {
+	getRangeIdentifier(data?: any) {
 		const options = this.model.getOptions();
-		const axisOptions = Tools.getProperty(
+		return Tools.getProperty(
 			options,
 			"axes",
-			this.rangeAxisPosition
+			this.getRangeAxisPosition({datum: data}),
```
theiliad, line 321: is `data` called `data` or `datum`? Seems like it should be the same in both places. If it's a datapoint then `datum`, and if it's multiple then `data`

packages/core/src/services/scales-cartesian.ts
```diff
@@ -224,11 +286,13 @@ export class CartesianScales extends Service {
 	}
 
 	getDomainValue(d, i) {
-		return this.getValueThroughAxisPosition(this.domainAxisPosition, d, i);
+		const axisPosition = this.getDomainAxisPosition(d);
```
theiliad, line 289: so some take `{datum: ...}` and some `(d)`? all should be consistent

packages/core/src/services/scales-cartesian.ts
```diff
@@ -103,6 +134,19 @@ export class CartesianScales extends Service {
...
+		const dualAxes = this.isDualAxes();
+		if (dualAxes) {
+			if (orientation === CartesianOrientations.HORIZONTAL) {
+				this.secondaryDomainAxisPosition = this.domainAxisPosition === AxisPositions.LEFT ? AxisPositions.RIGHT : AxisPositions.LEFT;
+				this.secondaryRangeAxisPosition = this.rangeAxisPosition === AxisPositions.BOTTOM ? AxisPositions.TOP : AxisPositions.BOTTOM;
+			} else {
+				this.secondaryDomainAxisPosition = this.domainAxisPosition === AxisPositions.BOTTOM ? AxisPositions.TOP : AxisPositions.BOTTOM;
+				this.secondaryRangeAxisPosition = this.rangeAxisPosition === AxisPositions.LEFT ? AxisPositions.RIGHT : AxisPositions.LEFT;
+			}
+		}
```
theiliad, line 149: shouldn't this happen in `findDomainAndRangeAxes()` or some other method?

packages/core/src/services/scales-cartesian.ts
```diff
@@ -53,14 +53,44 @@ export class CartesianScales extends Service {
...
+			if (axesOptions?.correspondingDatasets && axesOptions.correspondingDatasets.includes(dataset)) {
+				return this.secondaryDomainAxisPosition;
+			}
+		}
 		return this.domainAxisPosition;
 	}
 
-	getRangeAxisPosition() {
+	getRangeAxisPosition({datum = null, groups = null} = {}) {
+		if (this.dualAxes && datum) {
```
theiliad, line 77: if `datum` is falsey, then your `else if (groups && groups.length > 0)` wouldn't run bc `else if (groups && groups.length > 0)` wouldn't be satisfied

packages/core/src/services/essentials/dom-utils.ts
```diff
@@ -137,14 +137,16 @@ export class DOMUtils extends Service {
 		return finalDimensions;
 	}
 
-	static appendOrSelect(parent, query) {
+	static appendOrSelect(parent, query, id?) {
```
theiliad, line 140: a bit weird this new parameter. It's inconsistent with the rest of the parameters and the way this function works

theiliad (changes requested):

packages/core/src/components/graphs/skeleton.ts
```diff
@@ -170,7 +170,7 @@ export class Skeleton extends Component {
 		const height = this.backdrop.attr("height");
 		const width = this.backdrop.attr("width");
 		const ticksNumber = Tools.getProperty(
-			this.model.getOptions(),
+			this.getOptions(),
```
theiliad, line 173: there's no `this.getOptions()` in the Skeleton class, is there? Why would we not get options from the model?

packages/core/src/components/graphs/skeleton-lines.ts
```diff
@@ -29,32 +29,32 @@ export class SkeletonLines extends Skeleton {
 		svg.attr("width", width).attr("height", height);
 
 		const isDataLoading = Tools.getProperty(
-			this.model.getOptions(),
+			this.getOptions(),
```
theiliad, line 32: why not grab options from model?

packages/core/src/components/graphs/scatter.ts
```diff
@@ -166,13 +162,11 @@ export class Scatter extends Component {
 
 	styleCircles(selection: Selection<any, any, any, any>, animate: boolean) {
 		// Chart options mixed with the internal configurations
-		const options = this.model.getOptions();
+		const options = this.getOptions();
```
theiliad, line 165: again, why are we removing `model` here?

packages/core/demo/data/combo.ts
```diff
@@ -0,0 +1,460 @@
...
+		left: {
+			title: "Day of the Week",
+			mapsTo: "date",
+			scaleType: "labels"
+		}
+	},
+	chartTypes: [
+		{
+			type: "simple-bar",
+			options: {},
```
theiliad, line 85: what if `options` isn't included here? Does that cause an error?

packages/core/demo/data/combo.ts
```diff
@@ -0,0 +1,460 @@
...
+			datasets: ["School A"]
+		},
+		{
+			type: "line",
+			options: {},
+			datasets: ["Temperature"]
```
theiliad, line 91: 10 lines above here you've called it `correspondingDatasets`, which one is the correct one?

packages/core/src/charts/combo.ts
```diff
@@ -0,0 +1,123 @@
...
+		// With the user provided options
+		const chartOptions = Tools.mergeDefaultChartOptions(
+			Configuration.options.comboChart,
+			chartConfigs.options
+		);
+
+		// Warn user if no chartTypes defined
+		// Use skeleton chart instead
+		if (!chartConfigs.options.chartTypes) {
+			console.warn("No chartTypes defined for the Combo Chart!");
```
theiliad, line 59: if we consider this a potentially broken chart, then a console.error, if not, console.warn is fine

packages/core/src/charts/combo.ts
```diff
@@ -0,0 +1,123 @@
...
+
+		console.log(Tools.flatten(graphComponents));
+
+		return Tools.flatten(graphComponents);
+	}
+
+	getComponents() {
+		const { chartTypes } = this.model.getOptions();
+		// don't add the regular ruler if stacked ruler is added
+		const stackedRulerEnabled = chartTypes.filter(chartObject => {
```
theiliad, line 102: maybe `array.some` is a better option here https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/some

theiliad (changes requested):

packages/core/src/components/graphs/skeleton-lines.ts
natashadecoste: gets the this.modelgetOptions > merges it with any component specific options then returns it

packages/core/src/components/graphs/scatter.ts
natashadecoste: allows components to have their own styles (that are not global to other graphs)

packages/core/src/components/graphs/skeleton.ts
natashadecoste: its in the component class so ive updated components to use it.
in future if someone is making their own combo using constructors, they can add custom options to a skeleton graph that are applied on top of globals.

packages/core/demo/data/combo.ts
natashadecoste: updated 👍

packages/core/demo/data/combo.ts
natashadecoste: nope, you dont need it
if you dont need anything on top of the carbon charts styles that come with that chart you can put nothing or exclude it

packages/core/src/charts/combo.ts
natashadecoste: updated 👍

packages/core/src/charts/combo.ts
natashadecoste: good note - updated

packages/core/src/services/scales-cartesian.ts
natashadecoste: good call

packages/core/src/charts/combo.ts
```diff
@@ -0,0 +1,125 @@
...
+					return null;
+				}
+				let stacked;
+				options = Tools.merge({}, Configuration.options[`${Tools.camelCase(graph.type)}Chart`], this.model.getOptions(), graph.options);
+				// if we are creating a stacked area, the contained Line chart needs to know it is stacked
+				if (graph.type === ChartTypes.STACKED_AREA) {
+					stacked = true;
+				}
+				return graphComponentsMap[graph.type].map((Component, i) =>
+					new Component(this.model, this.services, { groups: graph.correspondingDatasets, id: counter++, options: options, stacked: stacked }));
```
theiliad, line 93:
```suggestion
					new Component(this.model, this.services, { groups: graph.correspondingDatasets, id: counter++, options, stacked }));
```

theiliad (changes requested):
code looks fine otherwise, demos are broken

packages/core/src/interfaces/charts.ts
```diff
@@ -317,6 +317,20 @@ export interface RadarChartOptions extends BaseChartOptions {
 }
 
 /**
+ * options specific to combo bar charts
+ */
+export interface ComboChartOptions extends AxisChartOptions {
+	chartTypes?: {
```
sophiiae, line 323:
```suggestion
	comboChartTypes?: {
```

packages/core/src/model.ts
```diff
@@ -44,18 +44,23 @@ export class ChartModel {
...
-		const dataGroups = this.getDataGroups();
-
+		const options = this.getOptions();
 		// Remove datasets that have been disabled
-		let displayData = Tools.clone(this.get("data"));
-		const { groupMapsTo } = this.getOptions().data;
+		let allData = this.getData();
+		const dataGroups = this.getDataGroups();
+		const { groupMapsTo } = Tools.getProperty(options, "data");
+		const axesOptions = Tools.getProperty( options, "axes");
```
sophiiae, line 56:
```suggestion
		const axesOptions = Tools.getProperty(options, "axes");
```

packages/core/src/services/scales-cartesian.ts
```diff
@@ -334,24 +386,35 @@ export class CartesianScales extends Service {
...
 			mainHorizontalScaleType === ScaleTypes.TIME
 		) {
-			result.domainAxisPosition = mainHorizontalAxisPosition;
-			result.rangeAxisPosition = mainVerticalAxisPosition;
+			result.primaryDomainAxisPosition = horizontalAxesPositions.primary;
+			result.primaryRangeAxisPosition = verticalAxesPositions.primary;
+			// secondary axes
+			result.secondaryDomainAxisPosition = horizontalAxesPositions.secondary;
+			result.secondaryRangeAxisPosition = verticalAxesPositions.secondary;
 		} else if (
```
sophiiae, line 403:
```suggestion
```
These codes are exactly the same in `else` condition

packages/core/src/components/graphs/area.ts
```diff
@@ -101,10 +112,9 @@ export class Area extends Component {
 				})}`
 			)
 			.node();
+
 		const colorValue = strokePathElement
-			? getComputedStyle(strokePathElement, null).getPropertyValue(
-					"stroke"
-			  )
+			? getComputedStyle(strokePathElement, null).getPropertyValue("stroke" )
```
sophiiae, line 117:
```suggestion
			? getComputedStyle(strokePathElement, null).getPropertyValue("stroke")
```

theiliad (approved):

sophiiae (approved):

Issue #736, opened 2020-08-05
[ZoomBar] fails to render handle-bar


sstone2423:
#### I am submitting a...

- [x] Source code defect

### Issue description

When using the zoomBar. the handle-bar. fails to render with the following error: 

![Screen Shot 2020-08-03 at 12 39 20 PM](https://user-images.githubusercontent.com/25177649/89419941-aa42eb80-d6f7-11ea-99ac-663e588df7f8.png)

#### Steps to produce the issue

Having trouble reproducing as we're using the same configuration as the story demo

#### Current behavior

Throws error and fails to fully render the zoombar

#### Expected behavior

Shouldn't throw error and should fully render the zoomBar

### Screenshot or recording

Screenshot with addons implementation:

![Screen Shot 2020-08-03 at 12 39 07 PM](https://user-images.githubusercontent.com/25177649/89419953-add67280-d6f7-11ea-9807-eef8f36e19ee.png)

Screenshot with exact configuration from storybook zoomBar demo:

![Screen Shot 2020-08-05 at 9 16 21 AM](https://user-images.githubusercontent.com/25177649/89423671-6999a100-d6fc-11ea-9089-0203d14ea956.png)

sstone2423:
Closed via #737

PR #737, opened 2020-08-05
fix: use enter().append() instead of join()

hlyang397:
### Updates
.join() could causes some problems in typescript project.
Use enter().append() instead.

- [ ] Demos all features
- [ ] Documented/annotated
- [ ] Matches UI/UX specs
- [ ] Accessible


sstone2423:
This will close this issue: https://github.com/carbon-design-system/carbon-charts/issues/736

PR #746, opened 2020-08-13
chore(monorepo): update logo

theiliad:
image


theiliad:
[charts_icon.sketch.zip](https://github.com/carbon-design-system/carbon-charts/files/5070991/charts_icon.sketch.zip)

PR #759, opened 2020-08-24
feat: change zoom bar handle cursor to ew-resize

hlyang397:
### Updates
Since the tooltip for zoom bar handle was cancelled, it makes more sense to change the zoom bar handle cursor from pointer to ew-resize.

image

- [ ] Demos all features
- [ ] Documented/annotated
- [ ] Matches UI/UX specs
- [ ] Accessible


packages/core/src/components/axes/zoom-bar.ts
```diff
@@ -304,6 +304,8 @@ export class ZoomBar extends Component {
 		const handleBarXDiff = -handleBarWidth / 2;
 		const handleYBarDiff = (handleHeight - handleBarHeight) / 2;
 
+		const ewHandleCursor = "ew-resize";
```
theiliad, line 307: what's the point of storing this in a variable?

PR #761, opened 2020-08-25
feat: enable chart grid enable option

JennChao:
### Updates
Sometimes user might not want to see chart grid. This PR provide the option of showing or not showing chart grid.

圖片

- [ ] Demos all features
- [ ] Documented/annotated
- [ ] Matches UI/UX specs
- [ ] Accessible


packages/core/src/components/axes/grid.ts
```diff
@@ -15,7 +15,7 @@ export class Grid extends Component {
 		const gridEnable = Tools.getProperty(
```
theiliad, line 15:
```suggestion
		const isGridEnabled = Tools.getProperty(
```

packages/core/src/components/axes/grid.ts
```diff
@@ -15,7 +15,7 @@ export class Grid extends Component {
...
 			this.model.getOptions(),
 			"grid",
-			"enable"
+			"enabled"
 		);
 		if (!gridEnable) {
```
theiliad, line 20:
```suggestion

		if (!gridEnable) {
```

packages/core/src/configuration.ts
```diff
@@ -62,7 +62,7 @@ const legend: LegendOptions = {
  */
 export const grid: GridOptions = {
 	// set enable to false will not draw grid and stroke of grid backdrop
-	enable: true,
+	enabled: false,
```
theiliad, line 65: grid cannot be disabled by default

packages/core/src/styles/components/_grid.scss
```diff
@@ -10,6 +10,16 @@
...
 
+	rect.chart-grid-backdrop-no-stroke {
+		@if $carbon--theme == $carbon--theme--g10 {
+			fill: $ui-01;
+		} @else if $carbon--theme == $carbon--theme--g90 {
+			fill: $carbon--gray-100;
+		} @else {
+			fill: $ui-background;
+		}
+	}
```
theiliad, line 21: why do we need this extra classname?

JennChao: The reason why we need `chart-grid-backdrop-no-stroke` classname is because we don't want to apply stroke around the chart when there is no grid line.

JennChao:
> we should probably allow individual grid enablement flags...
> 
> e.g. `x is active but y is disabled`

x grid and y grid can now be enabled or disabled separately

packages/core/src/configuration.ts
```diff
@@ -61,6 +61,10 @@ const legend: LegendOptions = {
  * Grid options
  */
 export const grid: GridOptions = {
+	// set enable to false will not draw grid and stroke of grid backdrop
+	xGridEnabled: true,
+	yGridEnabled: true,
+
```
natashadecoste, line 67: i think this should go within the x and y configurations liks this 
```
x: {
  enabled: true,
  numberOfTicks: 15
}
```

or maybe "visible" is a better word... since its still enabled/functional

packages/core/src/interfaces/components.ts
```diff
@@ -85,9 +85,11 @@ export interface ThresholdOptions {
 
 export interface GridOptions {
 	y?: {
+		yGridEnabled?: boolean;
```
theiliad, line 88: again, we always will want to keep things consistent by using `enabled`

packages/core/src/components/axes/ruler.ts
```diff
@@ -231,7 +241,9 @@ export class Ruler extends Component {
 		this.backdrop = DOMUtils.appendOrSelect(svg, "svg.chart-grid-backdrop");
 		const backdropRect = DOMUtils.appendOrSelect(
 			this.backdrop,
-			"rect.chart-grid-backdrop"
+			this.isXGridEnabled || this.isYGridEnabled
+				? "rect.chart-grid-backdrop.stroke"
+				: "rect.chart-grid-backdrop"
```
theiliad, line 246: seems like stroke should be a separate configurable no?

packages/core/src/configuration.ts
JennChao: Based on @theiliad 's input. I will use "enabled".

packages/core/src/components/axes/ruler.ts
JennChao: 圖片
A chart with grid but without stroke looks kind of weird. So I  guess we would always want to add stroke when there is grid being enabled. 
And when there is no grid being enabled, we can let the user setup the option to disable stroke? What do you think?

With stroke.
圖片

Without stroke.
圖片

theiliad: actually I think stroke might be something we'd always want to keep unless overriden

@jeanservaas could you please chime in?

JennChao: Hi @theiliad and @jeanservaas,
There is the situation when we don't want to show the stroke. 
圖片

packages/core/src/styles/components/_grid.scss
```diff
@@ -7,6 +7,8 @@
 		} @else {
 			fill: $ui-background;
 		}
+	}
+	rect.stroke {
```
theiliad, line 11:
```suggestion

	rect.chart-grid-backdrop.stroked {
```

packages/core/src/components/axes/grid.ts
```diff
@@ -197,7 +218,9 @@ export class Grid extends Component {
 		this.backdrop = DOMUtils.appendOrSelect(svg, "svg.chart-grid-backdrop");
 		const backdropRect = DOMUtils.appendOrSelect(
 			this.backdrop,
-			"rect.chart-grid-backdrop"
+			isXGridEnabled || isYGridEnabled
+				? "rect.chart-grid-backdrop.stroke"
```
theiliad, line 222:
```suggestion
				? "rect.chart-grid-backdrop.stroked"
```

packages/core/src/components/axes/ruler.ts
```diff
@@ -231,7 +243,9 @@ export class Ruler extends Component {
 		this.backdrop = DOMUtils.appendOrSelect(svg, "svg.chart-grid-backdrop");
 		const backdropRect = DOMUtils.appendOrSelect(
 			this.backdrop,
-			"rect.chart-grid-backdrop"
+			this.isXGridEnabled || this.isYGridEnabled
+				? "rect.chart-grid-backdrop.stroke"
```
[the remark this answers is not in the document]
theiliad: again my feedback here still applies. seems like `Ruler` doesn't need to style the backdrop, just needs to grab it.

this piece seems to be in there by mistake from our previous commits.

Are you able to remove all the styling pieces for the backdrop from `Ruler` and test whether things still work? We should remove any of it that doesn't break charts.

theiliad (changes requested):

packages/core/src/components/axes/ruler.ts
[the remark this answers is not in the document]
JennChao: Removing the code for styling backdrop in `Ruler` component has already been done in https://github.com/carbon-design-system/carbon-charts/pull/765

natashadecoste (approved):

PR #764, opened 2020-08-26
fix: stacked scatter tooltip not show

hlyang397:
- abstract getTooltipData function to allow override
- implement getTooltipData function in scatter-stacked with stacked data

### Updates
The scatter tooltip won't display in stacked charts except first data group because of the different data structure between normal data and stacked data.

### Demo screenshot or recording
Before fix:
image

After fix:
image

### Review checklist (for reviewers only)
- [ ] Demos all features
- [ ] Documented/annotated
- [ ] Matches UI/UX specs
- [ ] Meets the [code style guide](https://github.com/carbon-design-system/carbon-charts/wiki/Code-style-guide)
- [ ] Accessible
- [ ] Mobile first (responsive)
- [ ] RTL support (bidirectional text)
- [ ] Performant (limited bloat)

PR #765, opened 2020-08-26
feat: enable or disable ruler

JennChao:
### Updates
User can enable or disable chart ruler. 
Another PR or fixing tooltip not showing up when ruler is not applied has already been created. https://github.com/carbon-design-system/carbon-charts/pull/764

圖片

- [ ] Demos all features
- [ ] Documented/annotated
- [ ] Matches UI/UX specs
- [ ] Accessible


JennChao:
> wouldn't this PR need to also add some code in `ruler` to disable itself if it's been configured to be disabled?
> 
> Also @jeanservaas do we want to allow disabling the ruler?

@theiliad 
By saying **disable ruler itself**, do you mean not pushing `Ruler` component into `graphFrameComponents` array?
圖片

I actually suggest just have the following code in `Ruler` component. Then we don't need to have duplicated code for modifying all the chart components. 
圖片

packages/core/src/components/axes/ruler.ts
```diff
@@ -24,10 +24,17 @@ export class Ruler extends Component {
 		domainValue: number;
 		originalData: any;
 	}[];
+	isRulerEnabled =  Tools.getProperty(
```
theiliad, line 27: this should go inside `render`, that way it's value will update with future renders

packages/core/src/components/axes/ruler.ts
```diff
@@ -24,10 +24,17 @@ export class Ruler extends Component {
...
+		this.model.getOptions(),
+		"ruler",
+		"enabled"
+	);
 
 	render() {
 		this.drawBackdrop();
-		this.addBackdropEventListeners();
+		if (this.isRulerEnabled) {
```
theiliad, line 35: what if ruler was enabled and now isn't?

packages/core/src/components/axes/ruler.ts
```diff
@@ -24,10 +24,17 @@ export class Ruler extends Component {
...
 		originalData: any;
 	}[];
+	isRulerEnabled =  Tools.getProperty(
+		this.model.getOptions(),
+		"ruler",
+		"enabled"
+	);
 
 	render() {
 		this.drawBackdrop();
```
theiliad, line 34: it seems like `drawBackdrop` is repeated in ruler & grid. Would ruler fully function without it?

packages/core/src/components/axes/ruler.ts
JennChao: backdrop event listener will be remove first and then be created is if needed

packages/core/src/components/axes/ruler.ts
JennChao: ruler needs backdrop since the event listener is applied on backdrop.

theiliad: I get that ruler is dependant on backdrop, however I'm wondering if these lines are needed inside `ruler.ts`

image

JennChao: Actually, I am not sure about it. These lines of code have existed even before I added the ruler option.

theiliad: Yes I'm just asking whether removing them on your end still makes everything work?

packages/core/src/components/axes/ruler.ts
```diff
@@ -26,8 +26,24 @@ export class Ruler extends Component {
...
+
 		this.drawBackdrop();
-		this.addBackdropEventListeners();
+
+		// remove backdrop event listeners
+		this.removeBackdropEventListeners();
+
+		if (isRulerEnabled) {
+			this.addBackdropEventListeners();
+		}
```
theiliad, line 42: so we remove & add event listeners every time? This doesn't sound healthy, there should be a better way to only remove or add when we need to

packages/core/src/components/axes/ruler.ts
JennChao: As what I can see do far, removing code within line 253 to line 261 won't break any thing on my end.

packages/core/src/components/axes/ruler.ts
JennChao: Since d3 doesn't have an api for checking whether a specific eventListener is already existed or not. Removing and then adding event listeners when needed is what we can do for now.

packages/core/src/components/axes/ruler.ts
JennChao: Codes within line 253 to line 261 have already been removed.

packages/core/src/components/axes/ruler.ts
theiliad: again this is the same situation as the other PR, we just need a boolean here to keep track of event listeners

theiliad (changes requested):

packages/core/src/components/axes/ruler.ts
JennChao: A flag, `isEventListenerAdded` has been added.

packages/core/src/components/axes/ruler.ts
```diff
@@ -36,10 +36,30 @@ export class Ruler extends Component {
...
+		);
+
 		this.drawBackdrop();
-		this.addBackdropEventListeners();
+
+		if (isRulerEnabled && !this.isEventListenerAdded) {
+			this.isEventListenerAdded = true;
+			this.addBackdropEventListeners();
+		} else if (!isRulerEnabled && this.isEventListenerAdded) {
+			this.isEventListenerAdded = false;
```
theiliad, line 55: this line should be inside `removeBackdropEventListeners()`

packages/core/src/components/axes/ruler.ts
```diff
@@ -36,10 +36,30 @@ export class Ruler extends Component {
...
+			this.model.getOptions(),
+			"ruler",
+			"enabled"
+		);
+
 		this.drawBackdrop();
-		this.addBackdropEventListeners();
+
+		if (isRulerEnabled && !this.isEventListenerAdded) {
+			this.isEventListenerAdded = true;
```
theiliad, line 52: this line should be inside `addBackdropEventListeners()`

theiliad (changes requested):

theiliad (approved):

natashadecoste (approved):

sophiiae (approved):

PR #766, opened 2020-08-26
feat: provide option to refresh range axis label

hlyang397:
Sometimes data value varies from 10 to 100,000. When zoom bar is moving, the range axis (y-axis) label should be refreshed immediately to avoid extremely small graphs. 

### Updates
- add refreshRangeAxisLabel option (default to false)
- add filterDataForRangeAxisLabel() in zoom service to provide data in zoom domain

Option disabled (default) :
disabled

Option enabled:
enabled

- [ ] Demos all features
- [ ] Documented/annotated
- [ ] Matches UI/UX specs
- [ ] Accessible


packages/core/src/configuration.ts
```diff
@@ -149,7 +149,8 @@ const axisChart: AxisChartOptions = Tools.merge({}, chart, {
 	zoomBar: {
 		top: {
 			enabled: false,
-			type: ZoomBarTypes.GRAPH_VIEW
+			type: ZoomBarTypes.GRAPH_VIEW,
+			refreshRangeAxisLabel: false
```
theiliad, line 153: Not sure if this makes sense here. What if in top you set `refreshRangeAxisLabel` to true and in bottom you set it to false? Seems like this would be a `zoomBar` level config. And it should probably be called `updateRangeAxis` or `realtimeRangeAxis`

hlyang397: @theiliad
Thanks for the prompt review.
I think your comment makes sense.
The option name is changed to `updateRangeAxis` and moved to `ZoomBarsOptions` in 
https://github.com/carbon-design-system/carbon-charts/pull/766/commits/b9abd4c2abf9c738e059a342cc940cb96f71abde

hlyang397: @theiliad 
One more thought, how about moving this configuration to`AxisOptions` ?

theiliad: well that depends. It seems like this is only relevant when you have a zoombar, isn't that the case?

packages/core/src/services/zoom.ts
```diff
@@ -64,4 +100,32 @@ export class Zoom extends Service {
 			extent(zoomBarData, (d: any) => d[domainIdentifier])
 		);
 	}
+
+	// filter out data not inside zoom domain
+	// to get better range value for axis label
+	filterDataForRangeAxis(displayData: object[], stacked = false) {
+		const zoomDomain = this.model.get("zoomDomain");
+		const isUpdateRangeAxis = Tools.getProperty(
```
theiliad, line 108:
```suggestion
		const shouldUpdateRangeAxis = Tools.getProperty(
```

packages/core/src/configuration.ts
hlyang397: Yes, so far this option is required only zoom bar is enabled.

packages/core/src/services/zoom.ts
hlyang397: Updated in https://github.com/carbon-design-system/carbon-charts/pull/766/commits/9688034ff3f124008639b3de35559c84dee307c4.

packages/core/src/services/scales-cartesian.ts
```diff
@@ -416,12 +416,18 @@ export class CartesianScales extends Service {
 		let allDataValues;
 		// If the scale is stacked
 		if (axisOptions.stacked) {
-			const dataValuesGroupedByKeys = this.model.getDataValuesGroupedByKeys();
+			const dataValuesGroupedByKeys = this.services.zoom.filterDataForRangeAxis(
+				this.model.getDataValuesGroupedByKeys(),
+				true
```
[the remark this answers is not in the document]
hlyang397: Good idea. Updated in https://github.com/carbon-design-system/carbon-charts/pull/766/commits/9688034ff3f124008639b3de35559c84dee307c4.

natashadecoste (dismissed):

theiliad:
@hlyang397 how much have you tested this? what if you start with an initial zoom domain that has a smaller y-axis range and then you zoom-out all the way? Will you see the full range at all times?

hlyang397:
> @hlyang397 how much have you tested this? what if you start with an initial zoom domain that has a smaller y-axis range and then you zoom-out all the way? Will you see the full range at all times?

The y-axis range will be updated whenever zoom domain is changing, so it doesn't matter whether the chart starts from a smaller or bigger range.
Please refer to the demo below.

y-axis-update

theiliad:
> > @hlyang397 how much have you tested this? what if you start with an initial zoom domain that has a smaller y-axis range and then you zoom-out all the way? Will you see the full range at all times?
> 
> The y-axis range will be updated whenever zoom domain is changing, so it doesn't matter whether the chart starts from a smaller or bigger range.
> Please refer to the demo below.
> 
> y-axis-update

Yes I'm asking about the case when `refreshRangeAxisLabel` would be set to `false`. What would happen then?

hlyang397:
> 
> Yes I'm asking about the case when `refreshRangeAxisLabel` would be set to `false`. What would happen then?

@theiliad 
Yes, when `updateRangeAxis` is set to `false`, you will see the full range at all times, no matter what initial zoom domain is.
`updateRangeAxis` is default to `false`, so all storybook demos should look same as before.

packages/core/src/interfaces/components.ts
```diff
@@ -109,6 +109,10 @@ export interface ZoomBarsOptions {
 	 * currently only the top position is supported
 	 */
 	top?: ZoomBarOptions;
+	/**
+	 * whether keep updating range axis in real time while zoom domain is changing
+	 */
+	updateRangeAxis?: boolean;
```
theiliad, line 115: I've already commented on this before, not sure where my comment went (usually they disappear if you force-push over the branch and the lines that were commented on get removed).

This seems like it should just be flag under:
```js
{
  axes: {
      "top|left|bottom|right": {
         update: boolean;
      }
  }
}
```

whether it needs to be called `update` or `updateWhenZooming` depends on whether this only applies to zooming or not

theiliad (changes requested):

hlyang397:
@theiliad  I don't think we have any force-push in this PR.  In that case, we should be able to find the force-push record like this:

image

packages/core/src/interfaces/components.ts
hlyang397: Updated in https://github.com/carbon-design-system/carbon-charts/pull/766/commits/a224ab3561811cf0d6454093d2fb3af0fe5ab303.

theiliad:
> @theiliad I don't think we have any force-push in this PR. In that case, we should be able to find the force-push record like this:
> 
> image

yes, what I was saying about this is that if I comment on line X and that line has been removed in a force-pushed commit, then the comment seems to get deleted forever on Github.

hlyang397:
@theiliad  So any other comment on this PR?

theiliad (dismissed):
A bit unhappy that this is really just focusing on the main y-axis and not also the main x-axis, however I can't stand in the way of your progress

sophiiae:
@hlyang397 please fix the conflicts

natashadecoste (dismissed):

hlyang397:
> A bit unhappy that this is really just focusing on the main y-axis and not also the main x-axis, however I can't stand in the way of your progress

@theiliad We don't have vertical zoom bar yet, so I can't test this feature on main x-axis, and that's why I focus on main y-axis for now.

hlyang397:
> @hlyang397 please fix the conflicts

@sophiiae Latest master branch merged. Thanks.

theiliad (dismissed):

natashadecoste (approved):

theiliad (approved):

cal-smith (dismissed):
LGTM!

packages/core/src/services/zoom.ts
```diff
@@ -80,4 +115,37 @@ export class Zoom extends Service {
...
+		const zoomDomain = this.model.get("zoomDomain");
+		const mergedConfigs = Object.assign(
+			{ stacked: false }, // default configs
+			configs
+		);
+		const shouldUpdateRangeAxis = Tools.getProperty(
+			this.model.getOptions(),
+			"axes",
+			this.services.cartesianScales.getMainYAxisPosition(),
+			"updateWhenZooming"
```
theiliad, line 131: I was just testing this and realized that `updateWhenZooming` is available for all axes, however it only takes effect for the main Y-axis. this means that I can set it to false for the main X-axis and it will do nothing

theiliad: we should either just make it available for the range axis like before, or if we are allowing it to be used for all axis then it should achieve something when set to false

theiliad (changes requested):

packages/core/src/services/zoom.ts
hlyang397: @theiliad The domain axis is always changing while zoom domain is changing, so no option is required.
This new option should only works for the range axis (main Y axis).
So I change it back to  `updateRangeAxis` under `zoomBar`.

Updated in https://github.com/carbon-design-system/carbon-charts/pull/766/commits/2cd37ad112dcf86e38faebf6534fb397ed5225f4.

Please let me know if this is not what you expected.

theiliad (approved):

sophiiae (approved):

hlyang397:
@theiliad Not sure what happened to Travis CI,  I can pass the test locally.
Could you help to rerun the build in Travis CI ? Thanks.

PR #767, opened 2020-08-26
feat: enable or disable tooltip

JennChao:
### Updates
Use can enable or disable tooltip.

圖片

- [ ] Demos all features
- [ ] Documented/annotated
- [ ] Matches UI/UX specs
- [ ] Accessible


packages/core/src/components/essentials/tooltip.ts
```diff
@@ -83,7 +88,7 @@ export class Tooltip extends Component {
 				this.positionTooltip(e);
```
theiliad, line 88: why would you position a tooltip that might be hidden?

packages/core/src/components/essentials/tooltip.ts
```diff
@@ -80,10 +85,12 @@ export class Tooltip extends Component {
 				}
 
 				// Position the tooltip
-				this.positionTooltip(e);
+				if (this.isTooltipEnabled) {
```
theiliad, line 88: we probably also wouldn't want to call `getTooltipHTML` or run the logic above if tooltip is disabled

JennChao: please see the updated code

packages/core/src/components/essentials/tooltip.ts
```diff
@@ -19,6 +19,11 @@ export class Tooltip extends Component {
 
 	tooltip: any;
 	positionService = new Position();
+	isTooltipEnabled = Tools.getProperty(
```
theiliad, line 22: this should be somewhere inside in `render()`, where if the user disables the tooltip using an options update, currently the `enabled` value will be unchangeable.

JennChao: Class variable, `isTooltipEnabled` has already been removed.

packages/core/src/components/essentials/tooltip.ts
```diff
@@ -202,7 +209,14 @@ export class Tooltip extends Component {
...
 
 	render() {
-		this.tooltip.classed("hidden", true);
+		const isTooltipEnabled = Tools.getProperty(
+			this.model.getOptions(),
+			"tooltip",
+			"enabled"
+		);
+		if (isTooltipEnabled) {
+			this.tooltip.classed("hidden", true);
```
theiliad, line 218: so if tooltip isn't enabled we won't hide it?

JennChao: No, we don't need to hide tooltip when tooltip is disabled. When tooltip is disabled, we won't even define tooltip in `init`.

JennChao:
> This needs to work in an updatable way.
> 
> It looks like we need a new method in the class which checks for `isTooltipEnabled` and adds the event listeners, or removing them once `isTooltipEnabled` is false.
> 
> That way `render` and `init` can use it



> updatable

Tooltip eventListender are now all updatable since they are added in `render`. Since you want eventListener to be updatable, we can remove `init`

packages/core/src/components/essentials/tooltip.ts
```diff
@@ -214,7 +218,30 @@ export class Tooltip extends Component {
...
+			);
+
+			// Apply html content to the tooltip
+			this.tooltipTextContainer = DOMUtils.appendOrSelect(
+				this.tooltip,
+				"div.content-box"
+			);
+			this.tooltip.style("max-width", null);
+			this.removeTooltipEventListener();
+			this.addTooltipEventListener();
```
theiliad, line 243: so why do we need to remove & add every time? Seems like we just need to store a boolean within the class to keep track of if event listeners are actively attached, and if they are then we don't need to keep removing & adding right?

JennChao: `isEventListenerAdded` is now added as a flag to keep track of whether event listeners are actively attached or not.

packages/core/src/components/essentials/tooltip.ts
```diff
@@ -26,26 +29,42 @@ export class Tooltip extends Component {
...
+		}
+
+		// Position the tooltip
+		this.positionTooltip(e);
+
+		// Fade in
+		this.tooltip.classed("hidden", false);
+	}
+
+	handleShowTooltipVar = (e) => this.handleShowTooltip(e);
```
theiliad, line 61: why the extra function?

packages/core/src/components/essentials/tooltip.ts
```diff
@@ -57,40 +76,34 @@ export class Tooltip extends Component {
...
-					tooltipTextContainer.html(
-						this.model
-							.getOptions()
-							.tooltip.customHTML(data, defaultHTML)
-					);
-				} else {
-					// Use default tooltip
-					tooltipTextContainer.html(defaultHTML);
-				}
+			this.handleShowTooltipVar
```
theiliad, line 79: why wouldn't this just point to `handleShowTooltip` directly?

packages/core/src/components/essentials/tooltip.ts
```diff
@@ -26,26 +29,42 @@ export class Tooltip extends Component {
...
-		this.tooltip.style("max-width", null);
+		// if there is a provided tooltip HTML function call it
+		if (
+			Tools.getProperty(
+				this.model.getOptions(),
+				"tooltip",
+				"customHTML"
+			)
+		) {
+			this.tooltipTextContainer.html(
```
theiliad, line 44: `tooltipTextContainer` seems to be only used here, so why would we initialize it into `this` inside `render`? we should just move it here

packages/core/src/components/essentials/tooltip.ts
```diff
@@ -17,7 +17,10 @@ import { select, mouse } from "d3-selection";
 export class Tooltip extends Component {
 	type = "tooltip";
 
+	// flag for checking whether tooltip event listener is added or not
+	isEventListenerAdded = false;
 	tooltip: any;
+	tooltipTextContainer: any;
```
theiliad, line 23:
```suggestion
```

theiliad (changes requested):

packages/core/src/components/essentials/tooltip.ts
JennChao: The reason why we want the extra function here is to create a named function. That way we can remove the exact same/ identical event listener later. 

圖片

packages/core/src/components/essentials/tooltip.ts
JennChao: `handleShowTooltip` can't be access within `addTooltipEventListener()` function. Only `this.handleShowTooltip` can be access.

packages/core/src/components/essentials/tooltip.ts
JennChao: In order to use named function when adding event listener, I have to put `this.handleShowTooltipVar` as the second parameter of `this.services.events.addEventListener()`. I can't use the following way to add event listener hence I still need `this.tooltipTextContainer` for accessing it in `this.handleShowTooltip` function and then assigned `this.handleShowTooltip` function to a named function `this.handleShowTooltipVar` without passing `e` as parameter to `this.handleShowTooltipVar` when calling it in `this.services.events.addEventListener`.

```
		this.services.events.addEventListener(
			Events.Tooltip.SHOW,
			function(e){
                                ...
                                ...
                        }
		);
```

packages/core/src/components/essentials/tooltip.ts
theiliad: yes I understand what you're trying to achieve here, just not sure why the name function can't be `handleShowTooltip`. You're essentially wrapping a function into another function when you can use the 1st function itself without the need for wrapping.

packages/core/src/components/essentials/tooltip.ts
theiliad: `handleShowTooltip` can itself act as that function, you don't need an extra function to wrap `handleShowTooltip`. Please take a look at some of the graph classes that we have where we do the exact same thing

image

image

There's really no reason for `handleShowTooltipVar`

theiliad (changes requested):

packages/core/src/components/essentials/tooltip.ts
```diff
@@ -26,26 +29,45 @@ export class Tooltip extends Component {
...
+		if (
+			Tools.getProperty(
+				this.model.getOptions(),
+				"tooltip",
+				"customHTML"
+			)
+		) {
+			if (e.detail.content) {
+						const labelHTML = `<div class="title-tooltip">${e.detail.content}</div>`;
+						this.tooltipTextContainer.html(labelHTML);
```
theiliad, line 46: again, `tooltipTextContainer` shouldn't be stored in `this`, there's no point in doing that. If it's really only needed here then we should just move it over here

theiliad (changes requested):

packages/core/src/components/essentials/tooltip.ts
```diff
@@ -207,7 +222,40 @@ export class Tooltip extends Component {
...
+			this.tooltip = DOMUtils.appendOrSelect(
+				holder,
+				`div.${settings.prefix}--${chartprefix}--tooltip`
+			);
+
+			// Apply html content to the tooltip
+			const tooltipTextContainer = DOMUtils.appendOrSelect(
+				this.tooltip,
+				"div.content-box"
+			);
```
theiliad, line 247: do we need this query in here? would it not being there break anything? seems like it's not used

theiliad (changes requested):

theiliad (dismissed):

packages/core/src/components/essentials/tooltip.ts
```diff
@@ -207,7 +222,35 @@ export class Tooltip extends Component {
...
+			"enabled"
+		);
+		if (isTooltipEnabled) {
+			// Grab the tooltip element
+			const holder = select(this.services.domUtils.getHolder());
+			const chartprefix = Tools.getProperty(
+				this.model.getOptions(),
+				"style",
+				"prefix"
+			);
```
natashadecoste, line 237:
```suggestion
		const options = this.model.getOptions();
		const isTooltipEnabled = Tools.getProperty(
			options,
			"tooltip",
			"enabled"
		);
		if (isTooltipEnabled) {
			// Grab the tooltip element
			const holder = select(this.services.domUtils.getHolder());
			const chartprefix = Tools.getProperty(
				options,
				"style",
				"prefix"
			);
```

natashadecoste (changes requested):

theiliad (approved):

natashadecoste (approved):

PR #769, opened 2020-08-27
feat: enable or disable scatter dot on charts except scatter chart

JennChao:
### Updates
Allow user to enable or disable scatter dot in charts except scatter chart.

圖片

- [ ] Demos all features
- [ ] Documented/annotated
- [ ] Matches UI/UX specs
- [ ] Accessible


packages/core/src/components/graphs/scatter-stacked.ts
```diff
@@ -1,11 +1,22 @@
...
 import { Scatter } from "./scatter";
 import { Roles } from "../../interfaces";
+import { Tools } from "../../tools";
 
 export class StackedScatter extends Scatter {
 	type = "scatter-stacked";
 
 	render(animate: boolean) {
+		const isScatterEnabled = Tools.getProperty(this.model.getOptions(), "scatterDotEnabled");
+		console.log("!!! this.configs.ignoreScatterDisabled: ", this.configs.ignoreScatterDisabled);
```
theiliad, line 11: console.logs?

packages/core/src/configuration.ts
```diff
@@ -146,6 +146,7 @@ const axisChart: AxisChartOptions = Tools.merge({}, chart, {
 	axes,
 	timeScale,
 	grid,
+	scatterDotEnabled: true,
```
theiliad, line 149: probably under the `points` flag as `enabled`?

image

packages/core/src/charts/scatter.ts
```diff
@@ -44,7 +44,9 @@ export class ScatterChart extends AxisChart {
 			new TwoDimensionalAxes(this.model, this.services),
 			new Grid(this.model, this.services),
 			new Ruler(this.model, this.services),
-			new Scatter(this.model, this.services),
+			new Scatter(this.model, this.services, {
+				ignoreScatterDisabled: true
```
theiliad, line 48: we probably need a more readable name here. If I was looking at this file for the first time I probably wouldn't understand what this flag would do

packages/core/src/configuration.ts
JennChao: if we move `scatterDotEnabled` under the `points `flag as `enabled`, then only scatter chart can access the config option. However, axis charts, such as line chart and area chart, also need to access this config option as well.

packages/core/src/charts/scatter.ts
JennChao: how about `alwaysEnableScatterDot`?

packages/core/src/configuration.ts
theiliad: Don't think this'd be the case. Any chart that uses scatter should extend scatter chart and their options should do the same

JennChao: 圖片

Sometimes we don't want to have the scatter dot displayed. That's the reason why I setup this `scatterDotEnabled` option for letting the user enable or disable the scatter dot. And please note that, there will always be scatter dot displayed on scatter charts. Only line chart, area chart, and step chart can have the option to disable scatter dot.

theiliad: Yes in the screenshot you have above, `AreaChart` should be extending scatter which means the `enabled` flag would be available in there

JennChao: I think I need to figure out what you mean by saying `extending scatter chart`. Both area chart and scatter chart **extend `AxisChart`** and have **scatter component** in them. That's the reason why we can see the scatter dot. And from my understanding, **scatter component** will only be newed in `AxisChart`.
圖片

From what I have observed above, I think whether we want to show or not show the scatter dot, `Scatter` component, should not be configure in `scatterChart` option. `scatterDotEnabled` should be specify in `axisChart` option so all the axis charts with `Scatter` component can access this config option.

theiliad (changes requested):
This flag should really be under `points`. If a user manually chooses to also disable scatter dots on just a scatter chart, that's just a user error. They will immediately realize that this isn't a great idea and enable the points again. No reason to cause inconsistencies to APIs here

image

packages/core/src/configuration.ts
natashadecoste: what @theiliad is saying is that the `ScatterChartOptions` `extends` the `AxisChartOptions` meaning that all charts that have Scatter dots will have `ScatterChartOptions` **in addition to the**  `AxisChartOptions`.

so if you change ScatterChartOptions like this (charts.ts):
```
/**
 * options specific to scatter charts
 */
export interface ScatterChartOptions extends AxisChartOptions {
	/**
	 * options for the points
	 */
	points?: {
		/**
		 * sets the radius of the point
		 */
		radius: number;
		fillOpacity?: number;
		filled?: boolean;
		enabled?: boolean;
	};
}
```

than you can update how you access the config in both scatter-stacked.ts and scatter.ts:
`const isScatterEnabled = Tools.getProperty(this.model.getOptions(), "points", "enabled");`

natashadecoste (changes requested):

packages/core/src/configuration.ts
JennChao: I got what you were saying. However, if I put the `enabled` option in `points` of `ScatterChartOptions` then only scatter chart will be able to access this configuration. How about line chart? We might also want to disable scatter dot in line chart. 
You can see that there is no way for line chart to access `ScatterChartOptions` because of following few lines of code in `charts/line.ts` file. Line chart can only access `LineChartOptions`. Or maybe you actually want to have the `enabled` option being put within `ScatterChartOptions`, `LineChartOptions`, `BubbleChartOptions`, and `AreaChartOptions`? If that is the case, why don't we just have the `enabled` option put within `AxisChartOptions`?
圖片

theiliad (approved):

PR #770, opened 2020-08-27
feat: create options for tick rotation

hlyang397:
User may like to control if ticks need to be rotated ~while zoom domain is changing~ .
For example, if the tick number is small and the chart is wide enough, maybe tick rotation is not necessary.

### Updates
- new enums and configuration
- control x-axis tick rotation depending on configuration

Always rotate ticks ~while zooming~: 
always

Never rotate ticks ~while zooming~:
never

Rotate ticks when space is not enough: (default)
depending

- [ ] Demos all features
- [ ] Documented/annotated
- [ ] Matches UI/UX specs
- [ ] Accessible


packages/core/src/components/axes/axis.ts
```diff
@@ -65,6 +66,44 @@ export class Axis extends Component {
...
+			if (
+				!rotateWhileZooming || // default to always
+				rotateWhileZooming === TickRotations.ALWAYS
+			) {
+				// if option is set to ALWAYS or not set
+				return true;
+			} else if (rotateWhileZooming === TickRotations.NEVER) {
+				// if option is set to NEVER
+				return false;
+			} else if (rotateWhileZooming === TickRotations.DEPENDING) {
```
theiliad, line 95: this should probably be called `AUTO`

packages/core/src/components/axes/axis.ts
```diff
@@ -472,8 +511,7 @@ export class Axis extends Component {
 					: estimatedTickSize < minTickSize;
 			}
 
-			// always rotate ticks if zoomDomain is changing to avoid rotation flips during zoomDomain changing
-			if (rotateTicks || this.zoomDomainChanging) {
+			if (this.isTickRotationRequired(rotateTicks)) {
```
theiliad, line 514: why would we go through labels and calculate whether any need rotation before knowing whether we are rotating or not?

packages/core/src/components/axes/axis.ts
hlyang397: Updated in https://github.com/carbon-design-system/carbon-charts/pull/770/commits/b08826d54b5d951caef0a37a018920ebb2acfca9.

packages/core/src/components/axes/axis.ts
hlyang397: You are right, if we already know that we need to rotate or not, we could avoid some calculation

Updated in https://github.com/carbon-design-system/carbon-charts/pull/770/commits/f2489e73a5a93494d165e1c23251f39f939b0338.

hlyang397:
@theiliad,

> Couple of points here:
> 
>     * the code here strictly works for bottom & top axis. It should just really factor in which zoom bars are active and act accordingly

Updated in https://github.com/carbon-design-system/carbon-charts/pull/770/commits/bc124bb8840e2158e00a9187d11e9d7db445f629.

> 
>     * we should probably fix the issue of unnecessary tick rotations before fixing this. Seems like ticks rotate at times when they don't need to

I don't think so.
Even the tick rotation function works perfectly without unnecessary tick rotations, we still need this option to avoid the tick rotation flips during zoom domain changing.
So I think unnecessary tick rotations is another issue, but not necessary a dependency of this issue.

theiliad:
we should just move this flag out of the zoom bar options & into each individual axis:

```
{
  axes: {
    bottom: {
       ticks: {
           rotation: .....
       }
    }
  }
}
```

hlyang397:
> again, this seems like it's completely dependant on zooming. Why wouldn't this just work without the zoom bar?

@theiliad ,
Yes, it's completely  dependent on zooming.
I guess you misunderstand this option, so I just update the PR title.
This PR is about the tick rotation `while zooming`, not the tick rotation without zoom bar enabled.
When zoom bar is disabled, tick rotation behavior keeps as it is. It rotates whenever necessary. This option is meaningless.
When zoom bar is enabled, and user is changing zoom domain, this option allows user to decide if 
1. `ALWAYS` keep tick rotation `when zoom domain is changing`
2. `NEVER` let tick rotated (keep flat) `when zoom domain is changing`
3. `AUTO` let tick rotate whenever needed `when zoom domain is changing`

That's why I set the flag name `rotateWhileZooming`, and it has to be related to zoom bar.

If you are thinking about the tick rotation option when zoom domain is NOT changing or zoom bar is disabled, that would be another story.

Hope I make it clear after these explanation.

hlyang397:
> we should just move this flag out of the zoom bar options & into each individual axis:
> 
> ```
> {
>   axes: {
>     bottom: {
>        ticks: {
>            rotation: .....
>        }
>     }
>   }
> }
> ```

The flag is already under ticks of individual axis, but the name is `rotateWhileZooming` for now.
As I mentioned in another comment, this option is about tick rotation `while zooming`, so the name `rotation` may confuse user.
It's up to you, let me know the flag name you like to use.

theiliad (changes requested):
This really should work without being dependant on zooming.

If you've set the config to `auto` then it'll result in the library deciding when to rotate and when not to.

If you've set it to `never` then obviously ticks will never rotate.

If you've set it to `always`, then they'll always be rotated...

Therefore it should just be a `axisOptions.ticks.rotation` flag rather than `axisOptions.ticks. rotateWhileZooming`

hlyang397:
> This really should work without being dependant on zooming.
> 
> If you've set the config to `auto` then it'll result in the library deciding when to rotate and when not to.
> 
> If you've set it to `never` then obviously ticks will never rotate.
> 
> If you've set it to `always`, then they'll always be rotated...
> 
> Therefore it should just be a `axisOptions.ticks.rotation` flag rather than `axisOptions.ticks. rotateWhileZooming`
@theiliad 
The tick rotation option not related to zooming is actually another kind of option.
But this option also works to me for now. Updated in https://github.com/carbon-design-system/carbon-charts/pull/770/commits/e83c395b69c90a86c5a6b980b14b2f5521de38fa.

Since this PR is not related to zoom now, I also update PR title and description.

Let me know if you have any comment.

theiliad (approved):

sophiiae (approved):

natashadecoste (approved):

PR #773, opened 2020-08-28
feat(core): support shift click zooming

hlyang397:
### Updates
- Shift + click in chart could zoom in based on mouse position
- Dispatch ZoomDomain.CHANGE event whenever zoom domain is changed.

shiftClick

- [ ] Demos all features
- [ ] Documented/annotated
- [ ] Matches UI/UX specs
- [ ] Accessible


theiliad (changes requested):
again, with this PR we agreed to switch to this behaviour:
- Zoom onscroll (scrolling up would zoom in, scrolling down would zoom out)
- Pan on drag
- Zoom on shift + drag like you have in there currently

hlyang397:
> again, with this PR we agreed to switch to this behaviour:
> 
>     * Zoom onscroll (scrolling up would zoom in, scrolling down would zoom out)
> 
>     * Pan on drag
> 
>     * Zoom on shift + drag like you have in there currently

@theiliad 
1. This PR is only about the zoom on shift + click. Why this PR can't be simple without the other implementations? The other behaviors should be in another PRs.
2. I don't think we agree the other behaviors, especially the pan on drag.

theiliad:
> > again, with this PR we agreed to switch to this behaviour:
> > ```
> > * Zoom onscroll (scrolling up would zoom in, scrolling down would zoom out)
> > 
> > * Pan on drag
> > 
> > * Zoom on shift + drag like you have in there currently
> > ```
> 
> @theiliad
> 
> 1. This PR is only about the zoom on shift + click. Why this PR can't be simple without the other implementations? The other behaviors should be in another PRs.
> 2. I don't think we agree the other behaviors, especially the pan on drag.

@hlyang397 Pan on drag is what's been decided by design. We really need to implement that

hlyang397:
> > > again, with this PR we agreed to switch to this behaviour:
> > > ```
> > > * Zoom onscroll (scrolling up would zoom in, scrolling down would zoom out)
> > > 
> > > * Pan on drag
> > > 
> > > * Zoom on shift + drag like you have in there currently
> > > ```
> > 
> > 
> > @theiliad
> > 
> > 1. This PR is only about the zoom on shift + click. Why this PR can't be simple without the other implementations? The other behaviors should be in another PRs.
> > 2. I don't think we agree the other behaviors, especially the pan on drag.
> 
> @hlyang397 Pan on drag is what's been decided by design. We really need to implement that, as well as zoom onscroll

@theiliad Yes, I think we should provide these capabilities. But my question is why these can't be another PRs?

theiliad:
> > > > again, with this PR we agreed to switch to this behaviour:
> > > > ```
> > > > * Zoom onscroll (scrolling up would zoom in, scrolling down would zoom out)
> > > > 
> > > > * Pan on drag
> > > > 
> > > > * Zoom on shift + drag like you have in there currently
> > > > ```
> > > 
> > > 
> > > @theiliad
> > > 
> > > 1. This PR is only about the zoom on shift + click. Why this PR can't be simple without the other implementations? The other behaviors should be in another PRs.
> > > 2. I don't think we agree the other behaviors, especially the pan on drag.
> > 
> > 
> > @hlyang397 Pan on drag is what's been decided by design. We really need to implement that, as well as zoom onscroll
> 
> @theiliad Yes, I think we should provide these capabilities. But my question is why these can't be another PRs?

Sure as long as we know that we're going to add those piece then I'm good with merging this

packages/core/src/services/zoom.ts
```diff
@@ -64,4 +65,19 @@ export class Zoom extends Service {
 			extent(zoomBarData, (d: any) => d[domainIdentifier])
 		);
 	}
+
+	handleDomainChange(newDomain) {
+		this.model.set({ zoomDomain: newDomain }, { animate: false });
+		this.services.events.dispatchEvent(Events.ZoomDomain.CHANGE, {
+			newDomain
+		});
```
theiliad, line 73: 👍

packages/core/src/components/axes/grid-brush.ts
```diff
@@ -97,87 +103,76 @@ export class ChartBrush extends Component {
...
+					let leftPoint = clickedX - (width * zoomRatio) / 2;
+					if (leftPoint < 0) {
+						leftPoint = 0;
+					}
+					let rightPoint = clickedX + (width * zoomRatio) / 2;
+					if (rightPoint > width) {
+						rightPoint = width;
+					}
+					// handleZoomDomain assumes max range is [0, width]
+					handleZoomDomain(leftPoint, rightPoint);
```
theiliad, line 173: `handleZoomDomain` might not be the best name, what does this function do? It confuses me a bit too

theiliad: looks like it's updating the zoom domain, maybe `updateZoomDomain`?

theiliad (changes requested):
beyond the requested changes it's good to go

packages/core/src/services/zoom.ts
theiliad: no change needed here btw

packages/core/src/components/axes/grid-brush.ts
hlyang397: I am ok with `updateZoomDomain`.  Updated in https://github.com/carbon-design-system/carbon-charts/pull/773/commits/fb1d4ab541e920409606016791cc45cbe2872896.

theiliad (approved):

cal-smith (approved):
Makes sense to me!

Issue #748, opened 2020-08-14
Missing legend on SimpleBarChart, LineChart, StackedBarChart when loading option is set


kimbaz:
#### I am submitting a...

- [ ] Feature request
- [X] Design defect
- [ ] Source code defect
- [ ] Demo/documentation defect
- [ ] Other

#### charts version:
"@carbon/charts": "0.35.0",
"@carbon/charts-react": "0.35.0",
"d3": "5.16.0"

### Issue description
When setting the option:
```
"data": {
    "loading": true
  },
```
as in https://carbon-design-system.github.io/carbon-charts/react/?path=/story/bar-vertical--stacked-bar-skeleton once the data get loaded the chart legend is missing

#### Steps to produce the issue
Add the option:
```
"data": {
    "loading": true
  },
```
to any of these chart objects [SimpleBarChart, LineChart, StackedBarChart.

#### Current behavior
Chart legend is not displayed

#### Expected behavior
Chart legend should be displayed

kimbaz:
@sophiiae I tested it with
```
"@carbon/charts": "0.36.2",
"@carbon/charts-react": "0.36.2"
```
but I'm still able to reproduce it. Has the fix already been included in the latest release?
Thank you.

sophiiae:
> @sophiiae I tested it with
> 
> ```
> "@carbon/charts": "0.36.2",
> "@carbon/charts-react": "0.36.2"
> ```
> 
> but I'm still able to reproduce it. Has the fix already been included in the latest release?
> Thank you.

The PR is not merged yet. Once it merges, this issue should be closed.

kimbaz:
Looks good. Thank you

WilliamHolmes:
I'm still seeing this issue.
**Component**: `SimpleBarChart`
**Package Versions**:
```
"@carbon/charts": "0.36.2",
"@carbon/charts-react": "0.36.2",
"d3": "6.1.1",
```
![image](https://user-images.githubusercontent.com/5538260/92919723-e705af00-f428-11ea-8030-05705bfd5361.png)

sophiiae:
> I'm still seeing this issue.
> **Component**: `SimpleBarChart`
> **Package Versions**:
> 
> ```
> "@carbon/charts": "0.36.2",
> "@carbon/charts-react": "0.36.2",
> "d3": "6.1.1",
> ```
> 
> ### Loading = true
[...]

PR #776, opened 2020-08-31
Loading and locked option for zoom bar 

hlyang397:
### Updates
- create loading option dedicated to zoom bar
   - instead of using chart data loading option

- create locked option to zoom bar
   - add to zoom bar demo data
   - show zoom bar but not able to manipulate zoom bar handles or chart brush

Locked:
image

- [ ] Demos all features
- [ ] Documented/annotated
- [ ] Matches UI/UX specs
- [ ] Accessible


packages/core/src/services/zoom.ts
```diff
@@ -64,4 +64,22 @@ export class Zoom extends Service {
 			extent(zoomBarData, (d: any) => d[domainIdentifier])
 		);
 	}
+
+	isTopZoomBarLoading() {
```
theiliad, line 68: This should be `isZoomBarLoading(position)`

packages/core/src/services/zoom.ts
```diff
@@ -64,4 +64,22 @@ export class Zoom extends Service {
...
+		return Tools.getProperty(
+			this.model.getOptions(),
+			"zoomBar",
+			"top",
+			"loading"
+		);
+	}
+
+	isTopZoomBarLocked() {
```
theiliad, line 77: This should be `isZoomBarLocked(position)`

hlyang397:
> Why would we need to show a locked zoom bar over the skeleton state of a chart?
> 

When the zoom domain is changed, we will load new data for chart, and the loading takes some time. So we like to change the chart to skeleton state.
In the meanwhile, we hope to avoid the zoom bar skeleton state during data loading (because zoom bar will be exactly the same after data loading), and we don't want user to change the zoom domain again by zoom bar during data loading, so we need to make zoom bar as read only mode (locked).

packages/core/src/services/zoom.ts
hlyang397: Updated in https://github.com/carbon-design-system/carbon-charts/pull/776/commits/9436b0d67a17883273146053a5a6d799dac6c494.

packages/core/src/services/zoom.ts
hlyang397: Updated in https://github.com/carbon-design-system/carbon-charts/pull/776/commits/9436b0d67a17883273146053a5a6d799dac6c494.

theiliad:
This'll need a design review

theiliad:
@jeanservaas have you had a chance to review this?

theiliad:
So I just spoke to Jeannie, and not sure what this PR is trying to achieve.

@hlyang397 / Josh, could you please clarify the intent behind this PR?

Going into the deploy previews I'm seeing a demo where the chart is in a loading state and the zoombar isn't...
image

That doesn't add up, the intent behind the skeleton state is to show a loading state until your data has loaded.

**- If your data hasn't loaded yet, then you don't have an accurate zoombar**
**- If it has and your chart is still showing a skeleton state, why is that?**

hlyang397:
> So I just spoke to Jeannie, and not sure what this PR is trying to achieve.
> 
> @hlyang397 / Josh, could you please clarify the intent behind this PR?
> 
> Going into the deploy previews I'm seeing a demo where the chart is in a loading state and the zoombar isn't...
> image
> 
> That doesn't add up, the intent behind the skeleton state is to show a loading state until your data has loaded.
> 
> **- If your data hasn't loaded yet, then you don't have an accurate zoombar**
> **- If it has and your chart is still showing a skeleton state, why is that?**

@theiliad 
Here is the scenario.
1. In the beginning, user could see all data (in our case, in bar chart) with full time range. Because data count is so large, we group data by time into only 40 groups, which means only 40 bars are displayed in charts.
2. User manipulates zoom bar to a smaller time range, for example, only 10 bars are included.
3. After that we like to provide more detail data for the user, so we need to reload detail data from backend, and then refresh the bar chart to another 40 bars (detail data). That's why we need the skeleton state in chart.
4. However, in the meanwhile, the zoom bar doesn't need to refresh because it will not change. And we don't allow the zoom bar to be changed during the data loading, so it should be locked.

Let's invite designers to have more discussion on this requirement.

theiliad:
> > So I just spoke to Jeannie, and not sure what this PR is trying to achieve.
> > @hlyang397 / Josh, could you please clarify the intent behind this PR?
> > Going into the deploy previews I'm seeing a demo where the chart is in a loading state and the zoombar isn't...
> > image
> > That doesn't add up, the intent behind the skeleton state is to show a loading state until your data has loaded.
> > **- If your data hasn't loaded yet, then you don't have an accurate zoombar**
> > **- If it has and your chart is still showing a skeleton state, why is that?**
> 
> @theiliad
> Here is the scenario.
> 
> 1. In the beginning, user could see all data (in our case, in bar chart) with full time range. Because data count is so large, we group data by time into only 40 groups, which means only 40 bars are displayed in charts.
> 2. User manipulates zoom bar to a smaller time range, for example, only 10 bars are included.
> 3. After that we like to provide more detail data for the user, so we need to reload detail data from backend, and then refresh the bar chart to another 40 bars (detail data). That's why we need the skeleton state in chart.
> 4. However, in the meanwhile, the zoom bar doesn't need to refresh because it will not change. And we don't allow the zoom bar to be changed during the data loading, so it should be locked.
> 
> Let's invite designers to have more discussion on this requirement.

Hmm... I see. Yes if Josh could kindly bring this up on our guild call next week that'd be great. Thanks 🙂

packages/core/src/services/zoom.ts
```diff
@@ -64,4 +64,22 @@ export class Zoom extends Service {
...
+			"loading"
+		);
+	}
+
+	isZoomBarLocked(position) {
+		return Tools.getProperty(
+			this.model.getOptions(),
+			"zoomBar",
+			position,
+			"locked"
```
theiliad, line 82: wouldn't just `options.zoomBar.position.loading === false` mean that it should be locked? do we really need the extra flag?

packages/core/src/components/axes/zoom-bar.ts
```diff
@@ -56,16 +56,17 @@ export class ZoomBar extends Component {
 
 	render(animate = true) {
 		const svg = this.getContainerSVG();
-		const options = this.model.getOptions();
 
-		const isDataLoading = Tools.getProperty(options, "data", "loading");
+		const isTopZoomBarLoading = this.services.zoom.isZoomBarLoading("top");
+		const isTopZoomBarLocked = this.services.zoom.isZoomBarLocked("top");
```
theiliad, line 61: This should be using an ENUM for the position, `AxisPositions.TOP`...

packages/core/src/axis-chart.ts
```diff
@@ -74,7 +77,8 @@ export class AxisChart extends Chart {
 			}
 		};
 
-		if (zoomBarEnabled) {
+		// if all zoom bars are locked, no need to add chart brush
+		if (zoomBarEnabled && !isZoomBarLocked) {
```
theiliad, line 81: don't think we really need a `locked` flag, this can simply be `if (zoomBarEnabled && !isZoomBarLoading) {`

packages/core/src/axis-chart.ts
```diff
@@ -56,6 +56,9 @@ export class AxisChart extends Chart {
 			mainXAxisPosition === AxisPositions.BOTTOM &&
 			mainXScaleType === ScaleTypes.TIME;
 
+		// @todo - should check if zoom bar in all axes are locked
+		const isZoomBarLocked = this.services.zoom.isZoomBarLocked("top");
```
theiliad, line 60: `AxisPositions.TOP`

theiliad (changes requested):

packages/core/src/services/zoom.ts
hlyang397: `loading` means we should display the skeleton of zoom bar.
`locked` means we should display zoom bar as usually, but only read only, no zoom bar handle manipulation allowed.

packages/core/src/axis-chart.ts
hlyang397: Please refer to https://github.com/carbon-design-system/carbon-charts/pull/776#discussion_r491806328, `locked` state is different from `loading` state.

packages/core/src/axis-chart.ts
hlyang397: Updated in https://github.com/carbon-design-system/carbon-charts/pull/776/commits/f9d15696608945a99fa5b1fb7255dee7bf2ba634.

packages/core/src/components/axes/zoom-bar.ts
hlyang397: Updated in https://github.com/carbon-design-system/carbon-charts/pull/776/commits/f9d15696608945a99fa5b1fb7255dee7bf2ba634.

packages/core/demo/data/zoom-bar.ts
```diff
@@ -118,19 +118,47 @@ export const zoomBarLineTimeSeriesInitDomainOptions = addZoomBarToOptions(
...
+export const zoomBarLockedOptions = addZoomBarToOptions(
+	Object.assign(
+		{
+			data: {
+				loading: true
+			},
+			// required for correct spacing after PR 785
+			legend: {
+				enabled: false
+			}
```
theiliad, line 134: why do we need to disable the legend here?

theiliad (changes requested):

packages/core/demo/data/zoom-bar.ts
hlyang397: In PR 785, the logic is changed to legend will not disable automatically when data is loading.
But when legend is enabled while data is loading, the space between zoom bar and skeleton is gone.
I don't know why the code logic is changed, so I have to disable legend here.
https://github.com/carbon-design-system/carbon-charts/pull/785

theiliad: It was changed because when disabling the legend in loading, if you disable loading then there will be no legend

hlyang397: I see.
But if data is loading without disabling legend, the `Legend` and `legendSpacerComponent` components are still added into `fullFrameComponent `, and the `fullFrameComponentDirection` will be `LayoutDirection.COLUMN_REVERSE`.
In that case, the skeleton will be shifted up and the space between zoom bar and skeleton is gone.
That's why I disable legend here.

theiliad: well, is there a solution that works for both cases?

theiliad: Sure

hlyang397: @theiliad After some investigation, I prefer to keep to disabled legend in demo data because:

1. This problem is related to #748 #785 #802 , and I think it's out of this PR scope.
2. #785 looks weird to me. Maybe I didn't get the original issue it likes to resolve, but why we want to create an empty legend and cause the possible space problem ?
3. Both `skeleton` and `legend` are in `full-frame` component in axis-chart.
`skeleton` width and height are set to parent component's size, which means `skeleton` didn't expect `legend` should be displayed while `skeleton` is displayed (data is loading).
If we want to show skeleton and legend at the same time, something needs to change.
If we want to disable legend while showing skeleton (data loading), we may need to roll back #785.
Again, I believe this is out of this PR scope.

hlyang397:
@theiliad Any other comment on this PR?

theiliad (changes requested):

packages/core/demo/data/zoom-bar.ts
hlyang397: Legend option is removed from the demos in https://github.com/carbon-design-system/carbon-charts/pull/776/commits/765ae75541fdc1816c0688cef24a32e4acfc179b.

hlyang397:
@theiliad 
Legend option is removed from the demos in https://github.com/carbon-design-system/carbon-charts/commit/765ae75541fdc1816c0688cef24a32e4acfc179b.
Any other comment for this PR ?

theiliad:
> @theiliad
> Legend option is removed from the demos in [765ae75](https://github.com/carbon-design-system/carbon-charts/commit/765ae75541fdc1816c0688cef24a32e4acfc179b).
> Any other comment for this PR ?

Hi, the issue in question isn't whether the legend disabling option should be in the demo or not, it's really "why do you need to disable the legend for this state?"...

It should just be able to trigger the loading state without the user needing to make a change to the chart

hlyang397:
@theiliad
 > Hi, the issue in question isn't whether the legend disabling option should be in the demo or not, it's really "why do you need to disable the legend for this state?"...
> 
> It should just be able to trigger the loading state without the user needing to make a change to the chart

Yes, user can trigger the loading state by only setting `loading` to `true`. That's it.
User don't need to disable legend.

Disabling legend is just a workaround for other bug, not directly related to this PR. 
So I think you are right, it's confusing to leave `legend` option in the demo, that's why I remove it.

theiliad (approved):

hlyang397:
@natashadecoste @zvonimirfras @jeanservaas 
Could you help to review this PR ? Thanks.

natashadecoste (approved):

PR #777, opened 2020-08-31
Add color palette

sophiiae:


packages/core/src/styles/colors.scss
```diff
@@ -0,0 +1,59 @@
...
+}
+
+@if $carbon--theme == ($carbon--theme--g100 or $carbon--theme--g90) {
+	$current-theme: "dark";
+} @else {
+	$current-theme: "light";
+}
+
+$name: 'color-fill';
+$tokens: (
```
natashadecoste, line 41: are you going to map tokens to this variable based on the number of datagroups?

packages/core/src/styles/colors.scss
```diff
@@ -0,0 +1,59 @@
...
+	}
+}
+
+@if $carbon--theme == ($carbon--theme--g100 or $carbon--theme--g90) {
+	$current-theme: "dark";
+} @else {
+	$current-theme: "light";
+}
+
+$name: 'color-fill';
```
natashadecoste, line 40: do you need name though? i dont think this part would change based on the no. of datasets, the palette chosen or the theme 🤔

packages/core/src/components/graphs/area-stacked.ts
```diff
@@ -67,11 +68,20 @@ export class StackedArea extends Component {
...
 			.data(stackedData, (d) => d[0][groupMapsTo])
-			.attr("fill", (d) => self.model.getFillColor(d[0][groupMapsTo]))
+			.attr("class", "area")
+			.attr(
+				"class",
+				(d) => noProvidedColorScale
+				? `area ${this.model.getColorClass(d[0][groupMapsTo])}`
+				: "area"
+			)
+			.attr("fill", (d) => noProvidedColorScale ? null : self.model.getFillColor(d[0][groupMapsTo]))
```
theiliad, line 84: Only `model` should be looking at `Tools.getProperty(options, "color", "scale")`

packages/core/src/components/graphs/bubble.ts
```diff
@@ -66,17 +69,27 @@ export class Bubble extends Scatter {
...
 			// We need `|| 1` here in case the user doesn't provide radius values in data
 			.attr("r", (d) => radiusScale(d[radiusMapsTo] || 1))
-			.attr("fill", (d) =>
-				this.model.getFillColor(d[groupMapsTo], d[domainIdentifier], d)
+			.attr(
+				"class",
+				(d) => noProvidedColorScale
+					? `dot
+					${this.model.getStrokeColorClass(d[groupMapsTo])} 
+					${this.model.getColorClass(d[groupMapsTo])}
```
theiliad, line 77: Can this be a `this.model.getColorClassNames(["fill", "stroke"], d[groupMapsTo])`?

packages/core/src/components/graphs/gauge.ts
```diff
@@ -121,14 +121,21 @@ export class Gauge extends Component {
...
 			.enter()
 			.append("path")
-			.attr("class", "arc-foreground")
+			.attr(
+				"class",
+				(d) => noProvidedColorScale
+					? `arc-foreground ${self.model.getColorClass(d[groupMapsTo])}`
+					: "arc-foreground"
+			)
+			.attr("fill", (d) => noProvidedColorScale ? null : self.model.getFillColor(d[groupMapsTo]))
```
theiliad, line 136: we need to find a solution to this, it's not ideal to have every graph check for the custom color scale

packages/core/src/components/graphs/radar.ts
```diff
@@ -823,7 +834,7 @@ export class Radar extends Component {
 						.map((datum) => ({
 							label: datum[groupMapsTo],
 							value: datum[valueMapsTo],
-							color: self.model.getStrokeColor(datum[groupMapsTo])
+							class: self.model.getTooltipColorClass(datum[groupMapsTo])
```
theiliad, line 837: is the tooltip color class name any different than the fill color class name?

packages/core/src/configuration.ts
```diff
@@ -154,6 +154,9 @@ const chart: BaseChartOptions = {
 	},
 	color: {
 		scale: null,
+		presetPalette: {
```
theiliad, line 157: is this a pairing? then why not just call it `pairing`?

packages/core/src/index.ts
```diff
@@ -4,9 +4,4 @@ export * from "./charts";
 import * as configurations from "./configuration";
 import * as interfaces from "./interfaces";
 
-import * as colorPalettes from "./services/colorPalettes";
-import { DEFAULT } from "./services/colorPalettes";
-// TODO 1.0 - Remove deprecated API
-const defaultColors = DEFAULT;
-
-export { interfaces, configurations, defaultColors, colorPalettes };
+export { interfaces, configurations };
```
theiliad, line 7: 👍

packages/core/src/interfaces/charts.ts
```diff
@@ -104,6 +104,16 @@ export interface BaseChartOptions {
 		 */
 		scale?: object;
 		/**
+		 * use a carbon dataviz preset color palette
+		 * put the index (selection of which variant)
+		 */
+		presetPalette?: {
```
theiliad, line 110:
```suggestion
		pairing?: number;
```

packages/core/src/model.ts
```diff
@@ -326,7 +329,8 @@ export class ChartModel {
 
 		this.updateAllDataGroups();
 
-		this.setColorScale();
+		this.setUserProvidedColorScale();
+		this.setColorClasses();
```
theiliad, line 333: We should call them all `classNames` instead of `classes`, much more readable

packages/core/src/model.ts
```diff
@@ -578,47 +595,97 @@ export class ChartModel {
...
+		const options = this.getOptions();
+		const userProvidedScale = Tools.getProperty(options, "color", "scale");
+
+		let paletteIndex = Tools.getProperty(
+			options,
+			"color",
+			"presetPalette",
+			"index"
+		);
+		const paletteAvailableChoices = {
```
theiliad, line 638:
```suggestion
		const availableColorPairings = {
```
?

packages/core/src/model.ts
```diff
@@ -578,47 +595,97 @@ export class ChartModel {
...
+			"1-color": 4,
+			"2-color": 5,
+			"3-color": 5,
+			"4-color": 3,
+			"5-color": 2,
+			"14-color": 1
+		};
+		const numberOfDataGroups =
+			this.allDataGroups.length > 5 ? 14 : this.allDataGroups.length;
```
theiliad, line 647: why would we use a fake `numberOfDataGroups` value if more than 5? do we need a comment here to describe the decision?

packages/core/src/model.ts
```diff
@@ -578,47 +595,97 @@ export class ChartModel {
...
+			(dataGroup, index) =>
+				`tooltip-${numberOfDataGroups}-${paletteIndex}-${index + 1}`
+		);
+		const strokeColorClassesNames = this.allDataGroups.map(
+			(dataGroup, index) =>
+				`color-stroke-${numberOfDataGroups}-${paletteIndex}-${
+					index + 1
+				}`
+		);
+
```
theiliad, line 670: is there a difference between `tooltip`, `fill` & `stroke` classnames?

theiliad (changes requested):

jeanservaas:
@theiliad this is cool.

It would be really awesome to have controls to change the color pairings @mjabbink. Right now you can only change it through code and we'll have to plan to add a code tab to the data viz Palette page to provide guidance on how to switch pairings. 

Also @mjabbink we might want to make sure that the pairings we have listed 1st (i.e. the preset palette 1) are the colors we want to have as the defaults. The one and two colors sets are fine. But the 3 and especially 4 color sets are not the most attractive. The four-color one is defaulting to the pairing that has the most red in it right now. @theiliad this is my fault because this is how I organized them. 


### three color pairings

**current:**

image

**better default:**

image



### four color pairings

**current:**
image


**better default:**

image



I noticed something on bubble charts in Safari, when I brought my cursor over the live chart area the 30% opacity within the bubbles disappeared and did not return unless I refreshed

image

The same is happening with scatter:

image
Also on these thresholds... how did we come up with these colors? Are they just meant to be completely configurable? I also noticed the dots above the threshold have a much thicker stroke and are slightly larger, is that intentional?

image


**Completely unrelated to this PR:**

Why are the x-axis labels at a 45 degree angle here?

image

mjabbink:
I agree the current option 1 in 4-color group should not be the default and in fact become option 3.

sophiiae:
@jeanservaas Thank you for pointing out the scatter and bubble chart issue. I have fixed this and updated the default colors as you mentioned. The thresholds dash line color is completely configurable. Those two dots stand out with given thresholds, I didn't make any change for the threshold component in this PR. I believe it's intentional. If the x-axis labels are time series, the labels are rotated with 45 degrees.

packages/core/src/components/essentials/legend.ts
```diff
@@ -64,6 +64,13 @@ export class Legend extends Component {
 			"numCharacter"
 		);
 
+		const paletteIndex = Tools.getProperty(
+			options,
+			"color",
+			"paring",
```
natashadecoste, line 70:
```suggestion
			"pairing",
```

not sure im fully for the naming "pairing" but either way this is the right spelling

packages/core/demo/data/bar.ts
```diff
@@ -90,6 +90,11 @@ export const simpleBarOptions = {
 			mapsTo: "group",
 			scaleType: "labels"
 		}
+	},
+	color: {
+		paring: {
```
natashadecoste, line 95:
```suggestion
		pairing: {
```

packages/core/src/styles/colors.scss
```diff
@@ -0,0 +1,313 @@
...
+			"7": map-get(map-get($ibm-color-map, "green"), 60),
+			"8": map-get(map-get($ibm-color-map, "blue"), 80),
+			"9": map-get(map-get($ibm-color-map, "magenta"), 50),
+			"10": #b28600,
+			//yellow 50, update when available in carbon color
+				"11": map-get(map-get($ibm-color-map, "teal"), 50),
+			"12": map-get(map-get($ibm-color-map, "cyan"), 90),
+			"13": #8a3800,
+			//orange 70, update when available in carbon color
+				"14": map-get(map-get($ibm-color-map, "purple"), 50)
```
natashadecoste, line 122:
```suggestion
			"14": map-get(map-get($ibm-color-map, "purple"), 50)
```

packages/core/src/styles/colors.scss
```diff
@@ -0,0 +1,313 @@
...
+			"3": map-get(map-get($ibm-color-map, "teal"), 70),
+			"4": map-get(map-get($ibm-color-map, "magenta"), 70),
+			"5": map-get(map-get($ibm-color-map, "red"), 50),
+			"6": map-get(map-get($ibm-color-map, "red"), 90),
+			"7": map-get(map-get($ibm-color-map, "green"), 60),
+			"8": map-get(map-get($ibm-color-map, "blue"), 80),
+			"9": map-get(map-get($ibm-color-map, "magenta"), 50),
+			"10": #b28600,
+			//yellow 50, update when available in carbon color
+				"11": map-get(map-get($ibm-color-map, "teal"), 50),
```
natashadecoste, line 118:
```suggestion
			"11": map-get(map-get($ibm-color-map, "teal"), 50),
```

packages/core/src/styles/colors.scss
```diff
@@ -0,0 +1,313 @@
...
+			"3": map-get(map-get($ibm-color-map, "teal"), 60),
+			"4": map-get(map-get($ibm-color-map, "magenta"), 40),
+			"5": map-get(map-get($ibm-color-map, "red"), 50),
+			"6": map-get(map-get($ibm-color-map, "red"), 10),
+			"7": map-get(map-get($ibm-color-map, "green"), 30),
+			"8": map-get(map-get($ibm-color-map, "blue"), 50),
+			"9": map-get(map-get($ibm-color-map, "magenta"), 60),
+			"10": #d2a106,
+			//yellow 40, update when available in carbon color
+				"11": map-get(map-get($ibm-color-map, "teal"), 40),
```
natashadecoste, line 240:
```suggestion
			"11": map-get(map-get($ibm-color-map, "teal"), 40),
```

packages/core/src/styles/colors.scss
```diff
@@ -0,0 +1,313 @@
...
+			"12": map-get(map-get($ibm-color-map, "cyan"), 20),
+			"13": #ba4e00,
+			//orange 60, update when available in carbon color
+				"14": map-get(map-get($ibm-color-map, "purple"), 30)
```
natashadecoste, line 244:
```suggestion
			"14": map-get(map-get($ibm-color-map, "purple"), 30)
```

packages/core/src/configuration.ts
```diff
@@ -154,6 +154,9 @@ const chart: BaseChartOptions = {
 	},
 	color: {
 		scale: null,
+		paring: {
```
natashadecoste, line 157:
```suggestion
		pairing: {
```

packages/core/src/interfaces/charts.ts
```diff
@@ -104,6 +104,16 @@ export interface BaseChartOptions {
 		 */
 		scale?: object;
 		/**
+		 * use a carbon dataviz preset color palette
+		 * put the index (selection of which variant)
+		 */
+		paring?: {
```
natashadecoste, line 110:
```suggestion
		pairing?: {
```

packages/core/src/model.ts
```diff
@@ -454,6 +464,35 @@ export class ChartModel {
...
+			Object.keys(userProvidedScale).length > 0
+		);
+	}
+
+	getColorClassName(types: string[], group: any, originalClassName?: any) {
+		if (this.colorIsProvided()) {
+			return originalClassName;
+		}
+
+		const colorParingTag = this.colorClassNames(group);
```
natashadecoste, line 485:
```suggestion
		const colorPairingTag = this.colorClassNames(group);
```

packages/core/src/model.ts
```diff
@@ -454,6 +464,35 @@ export class ChartModel {
...
+		let className = originalClassName;
+		types.forEach(
+			(type) =>
+				(className = originalClassName
+					? `${className} ${type}-${colorParingTag}`
+					: `${type}-${colorParingTag}`)
```
natashadecoste, line 491:
```suggestion
					: `${type}-${colorPairingTag}`)
```

packages/core/src/model.ts
```diff
@@ -584,47 +623,71 @@ export class ChartModel {
...
 	}
+
+	/*
+	 * Color palette
+	 */
+	protected setColorClasses() {
+		let paringIndex = Tools.getProperty(
+			this.getOptions(),
+			"color",
+			"paring",
```
natashadecoste, line 657:
```suggestion
			"pairing",
```

packages/core/src/model.ts
```diff
@@ -584,47 +623,71 @@ export class ChartModel {
...
 		this.colorScale = scaleOrdinal()
 			.range(colorRange)
 			.domain(this.allDataGroups);
 	}
+
+	/*
+	 * Color palette
+	 */
+	protected setColorClasses() {
+		let paringIndex = Tools.getProperty(
```
natashadecoste, line 654:
```suggestion
		let pairingIndex = Tools.getProperty(
```

natashadecoste (changes requested):
it doesn't seems that you can choose a bigger palette than the number of datagroups.
What if i have 2 expected datagroups (ex days of the weekend) but I know that i might sometimes have 3 (a long weekend) and i'd like the colors to stay consistent in the chart/dashboard. 

could it make sense to allow someone to pick a 3 color palette for 2 datagroups in this case? Otherwise it seems the colors are going to jump around since there isnt a 2 color palette that extends to being a 3 color

packages/core/src/interfaces/charts.ts
```diff
@@ -104,6 +104,20 @@ export interface BaseChartOptions {
...
 		scale?: object;
 		/**
+		 * use a carbon dataviz preset color palette
+		 * put the index (selection of which variant)
+		 */
+		pairing?: {
+			/**
+			 * the number of data groups/categories, only if differ from given data
+			 */
+			numberOfGroups?: number;
```
natashadecoste, line 114: I'd rather this was called colorGroups, colorVariants or something that makes it more intuitive that its the number of **colors** rather than groups that you are overriding in the chart. Since i think you could just **choose** to use a bigger colorset for visual parity with the rest of the dashboard/page,  it makes this config seem a bit misleading to me.

natashadecoste (changes requested):

packages/core/src/interfaces/charts.ts
sophiiae: I know it's not a good name for this config, I'd like to see if anyone comes up with a better idea for this. I thought about `colorGroups`, it seems more like the input is an array of string, not an integer. It might be a bit confused about the type.

natashadecoste: variants? numberOfVariants? 🤔

jeanservaas (approved):
I'm gonna approve this now because I really want to get it in, and then we can change the order of the color groupings, which will also need to be updated on the site and in the master charts file — into another PR.

packages/core/src/interfaces/charts.ts
sophiiae: I'll take `numberOfVariants`.

packages/core/src/interfaces/charts.ts
```diff
@@ -104,6 +104,20 @@ export interface BaseChartOptions {
 		 */
 		scale?: object;
 		/**
+		 * use a carbon dataviz preset color palette
+		 * put the index (selection of which variant)
+		 */
+		pairing?: {
+			/**
+			 * the number of data groups/categories, only if differ from given data
```
natashadecoste, line 112:
```suggestion
			 * the number of color variants in the palette (defaults to using the number of data groups in the given data)
```

natashadecoste (changes requested):
What do we do if they are using custom colors? 
If i add custom color scale (on a line chart demo), it only colors those datasets -> it doesnt affect their **legend items**

image

And then there's the case of if they put less custom colors than there is datagroups - do we default them to carbon palette? or use a carbon palette for the colors that are remaining (not defined)? 

it turns out like this (below)
image

same legend problem but also the chart **looks correct** but is mapping 2 datasets to the 2 colors when the chart needs 4 color variants. I argue this is worse than a chart warning or erroring out because its just looking really false

packages/core/src/components/graphs/area.ts
```diff
@@ -76,7 +76,17 @@ export class Area extends Component {
 			.selectAll("path.area")
 			.data(groupedData, (group) => group.name);
 
-		if (isGradientAllowed) {
+		// The fill value of area has been overwritten, get color value from stroke color class instead
+		const areaElement = document.getElementsByClassName(
+			this.model.getColorClassName(["stroke"], groupedData[0].name)
+		)[0];
```
natashadecoste, line 82: why not use a d3 selection??

natashadecoste (approved):

packages/core/src/components/graphs/area.ts
theiliad: yes please update this

packages/core/src/components/essentials/tooltip.ts
```diff
@@ -174,15 +174,18 @@ export class Tooltip extends Component {
 		} else {
 			const items = this.getItems(e);
 			const formattedItems = this.formatItems(items);
+			const useColor = this.model.colorIsProvided();
```
theiliad, line 177: we need a more readable name here

packages/core/src/components/essentials/legend.ts
```diff
@@ -80,6 +87,15 @@ export class Legend extends Component {
...
+			.attr("class", (d, i) => {
+				if (paletteIndex) {
+					return this.model.getColorClassName(
+						["fill"],
+						d.name,
+						"checkbox"
+					);
+				}
+			})
 			.style("fill", (d) => {
```
theiliad, line 99: so we need both `class` & `fill`?

theiliad (changes requested):

packages/core/src/model.ts
```diff
@@ -584,47 +623,81 @@ export class ChartModel {
...
+		);
+		if (!numberOfVariants || numberOfVariants < this.allDataGroups.length) {
+			numberOfVariants = this.allDataGroups.length;
+		}
+
+		let pairingIndex = Tools.getProperty(colorPairingOptions, "index");
+		const availableColorPairings = {
+			"1-color": 4,
+			"2-color": 5,
+			"3-color": 5,
```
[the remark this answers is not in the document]
sophiiae: they have the same number of color pairs.

packages/core/src/components/essentials/legend.ts
sophiiae: we need `class` for color palette and `fill` for custom colors when user provides the color scale.

packages/core/src/model.ts
```diff
@@ -454,6 +465,46 @@ export class ChartModel {
...
+		const userProvidedScale = Tools.getProperty(
+			this.getOptions(),
+			"color",
+			"scale"
+		);
+		const dataGroups = this.getDataGroups();
+
+		if (
+			userProvidedScale == null ||
+			Object.keys(userProvidedScale).length < dataGroups.length
```
theiliad, line 478: what if I'm using the color palette and I provide a color scale that changes just 1 color? Would that be disregarded here?

packages/core/src/model.ts
```diff
@@ -454,6 +465,46 @@ export class ChartModel {
 		return this.colorScale;
 	}
 
+	isCustomColorValid() {
```
theiliad, line 468:
```suggestion
	isUserProvidedColorScaleValid() {
```

packages/core/src/model.ts
```diff
@@ -1,8 +1,8 @@
 // Internal Imports
 import * as Configuration from "./configuration";
 import { Tools } from "./tools";
-import * as colorPalettes from "./services/colorPalettes";
 import { Events, ScaleTypes } from "./interfaces";
+import { ColorClassNameTypes } from "./interfaces/enums";
```
theiliad, line 5:
```suggestion
import { Events, ScaleTypes, ColorClassNameTypes } from "./interfaces";
```

theiliad (changes requested):

packages/core/src/model.ts
sophiiae: Yes, if user only provides partial color scale, those colors will be assigned to the components repeatedly. Currently, we only apply colors either by class or color fill. I can make another PR later for this so user can partially override the color.

theiliad (approved):

PR #780, opened 2020-08-31
fix(meter) render max width bar for values over 100

natashadecoste:
### Updates
- meter will display a text value > 100 but will make the bar max width
- changed some code because it seems the status classes were not getting cleared, just added to

- [ ] Demos all features
- [ ] Documented/annotated
- [ ] Matches UI/UX specs
- [ ] Accessible


packages/core/src/model-meter.ts
```diff
@@ -52,17 +52,21 @@ export class MeterChartModel extends ChartModel {
 		const options = this.getOptions();
 		const dataValue = this.getDisplayData().value;
 
+		// use max value if the percentage is bigger than 100%
+		const boundedValue = dataValue > 100 ? 100 : dataValue;
```
theiliad, line 56: does this result in the incorrect value being shown in the percentage portion of the title?

natashadecoste: nope! its so that if the num > 100, it will still get the status for the max (100)

theiliad (dismissed):

sophiiae (approved):

theiliad (approved):

PR #781, opened 2020-09-01
Linear gradient option for area chart

JennChao:
### Updates
Allow user to have option of having gradient for area charts. (not including stacked area charts).

圖片

- [ ] Demos all features
- [ ] Documented/annotated
- [ ] Matches UI/UX specs
- [ ] Accessible


theiliad:
This requires a design review initially

jeanservaas:
Hi @JennChao 

After looking at this with @theiliad and @mjabbink it's hard to see this (even though it's really pretty), it's hard to see how it adds value or clarity to the chart — i.e. it seems purely aesthetic. 

Also, the fading gradient detracts from the purpose of the area chart, which is to expose trends by allowing users to see the overlap between several line charts. 

We're trying only to use gradients (continuous color) where they serve a specific purpose. For instance, in the continuous legend (where there's not space for stepped color) or in heat maps where the stepped color palette does not offer enough depth to represent the data. We could also explore gradient usage in areas where, say a line or bar chart is exceeding a maximum use cap and entering a danger zone etc.

JennChao:
> Hi @JennChao
> 
> After looking at this with @theiliad and @mjabbink it's hard to see this (even though it's really pretty), it's hard to see how it adds value or clarity to the chart — i.e. it seems purely aesthetic.
> 
> Also, the fading gradient detracts from the purpose of the area chart, which is to expose trends by allowing users to see the overlap between several line charts.
> 
> We're trying only to use gradients (continuous color) where they serve a specific purpose. For instance, in the continuous legend (where there's not space for stepped color) or in heat maps where the stepped color palette does not offer enough depth to represent the data. We could also explore gradient usage in areas where, say a line or bar chart is exceeding a maximum use cap and entering a danger zone etc.

Hi @jeanservaas @theiliad,
The reason why I added the option for the user to enable or disable gradient is because we do need the gradient style in our product. As you can see in the following screenshot. Currently, we are using another third party chart library called `ECharts` in our project. We gradually want to replace `ECharts` with `Carbon Chats ` in our project. That's why we come over here to have this gradient PR created.
圖片


From my perspective, this is only an option for having gradient style or not. By default, gradient won't be applied to area chart. Only when user wants the gradient style, user can set `enabled` to `true`, then there 
will be gradient. So probably we can still have the gradient option added to Carbon Chart?

jeanservaas:
@JennChao 

We talked about this in the guild meeting today and both @theiliad and I are OK with using the gradient in the situation you're showing above as long as there's only one color/one data set and no overlaying of gradients. @theiliad also wants to prevent the ability to add multiple gradients to a line chart in implementation.

jeanservaas (dismissed):

JennChao:
> @JennChao
> 
> We talked about this in the guild meeting today and both @theiliad and I are OK with using the gradient in the situation you're showing above as long as there's only one color/one data set and no overlaying of gradients. @theiliad also wants to prevent the ability to add multiple gradients to a line chart in implementation.

Hi @theiliad and @jeanservaas 
Now with my newly committed code, gradient style will only be applied when there is only one dataset and the `gradientEnabled` option is true.

packages/core/src/services/essentials/gradient-utils.ts
```diff
@@ -0,0 +1,75 @@
+// Internal Imports
+import { Service } from "../service";
+import { DOMUtils } from "../";
+
+export class GradientUtils extends Service {
+	static appendLinearGradient(svg, id, x1, x2, y1, y2, stopAttrList) {
```
theiliad, line 6: can this be a `configs` flag?
```suggestion
	static appendLinearGradient(configs) {
```

theiliad: this way where you're calling `appendLinearGradient` we won't just see 6 unnamed parameters, instead we'll see

```js
appendLinearGradient({
  svg: "blah",
  id: "blah",
  ............................
})
```

much more readable

theiliad: and we should just rename `stopAttrList` to `stops`.

They're not really `attrs` that are being passed in, rather they are actual stop definitions that we use to apply attrs to our element.

packages/core/src/components/graphs/area.ts
```diff
@@ -37,18 +43,51 @@ export class Area extends Component {
 				.x((d, i) => cartesianScales.getDomainValue(d, i))
 				.y0(cartesianScales.getRangeValue(0))
 				.y1((d, i) => cartesianScales.getRangeValue(d, i));
+			minAndMax = this.services.cartesianScales.getMainYScale().domain();
```
theiliad, line 46: If you're setting `minAndMax` to a domain, shouldn't `minAndMax` just be called `domain`?

packages/core/src/components/graphs/area.ts
```diff
@@ -37,18 +43,51 @@ export class Area extends Component {
...
+		if (isGradientAllowed) {
+			groupedData.forEach((dataset) => {
+				GradientUtils.appendLinearGradient(
+					this.parent,
+					dataset.name.replace(" ", "") + "_" + this.gradient_id,
+					"0%",
+					"0%",
+					"0%",
+					"100%",
+					GradientUtils.getStopArray(
```
theiliad, line 83: `getStopArray` should just receive `minAndMax` instead and it'd talk to `need3Stops` internally

packages/core/src/interfaces/charts.ts
```diff
@@ -192,6 +192,10 @@ export interface LineChartOptions extends ScatterChartOptions {
  * options specific to area charts
  */
 export interface AreaChartOptions extends AxisChartOptions {
+	/**
+	 * enable gradient or not
+	 */
+	gradientEnabled?: boolean;
```
theiliad, line 198: this should be a configurable under `color`.

```js
color: {
        gradient: {
            enabled: true
        }
}
```

packages/core/src/components/graphs/area.ts
```diff
@@ -37,18 +43,51 @@ export class Area extends Component {
...
 
 		// Update the bound data on area groups
 		const groupedData = this.model.getGroupedData();
+
+		// Should gradient style be applicable
+		const isGradientAllowed =
+			groupedData
+			&& groupedData.length === 1
+			&& Tools.getProperty(this.model.getOptions(), "gradientEnabled");
+
```
theiliad, line 63:
```suggestion
if (groupedData.length > 1 && Tools.getProperty(this.model.getOptions(), "gradientEnabled")) {
  console.error("Gradients can only be enabled when having 1 single dataset");
}
```

theiliad (changes requested):

packages/core/src/services/essentials/gradient-utils.ts
```diff
@@ -0,0 +1,76 @@
...
+	static need3Stops(domain) {
+		if (domain[0] < 0 && domain[1] > 0) {
+			return true;
+		}
+	}
+
+	static getOffsetRatio(dataset) {
+		const minAndMax = [
+			Math.min.apply(Math, dataset.map(function(o) { return o.value; })),
+			Math.max.apply(Math, dataset.map(function(o) { return o.value; }))
```
theiliad, line 32: multiple issues in the above 2 lines:
- you're hard-coding `.value`, however that field is overridable using `mapsTo` in the options. We cannot do that since the code will break for users who use `mapsTo`
- running `Math.min` & `Math.max` on a large data array can be very performance heavy, is this information something that we can grab from the model, or any of the axis domains?

packages/core/src/configuration.ts
```diff
@@ -139,7 +139,8 @@ const chart: BaseChartOptions = {
 		selectedGroups: []
 	},
 	color: {
-		scale: null
+		scale: null,
+		gradientEnabled: false
```
theiliad, line 143: again, we need this to be:
```js
{
  color: {
    gradient: {
       enabled: false
    }
  }
}
```

since we possibly will have to add more flags/customizables in the future for gradients.

packages/core/src/components/graphs/area.ts
```diff
@@ -37,18 +43,56 @@ export class Area extends Component {
...
+		// Should gradient style be applicable
+		const isGradientAllowed =
+			groupedData
+			&& groupedData.length === 1
+			&& Tools.getProperty(this.model.getOptions(), "color", "gradientEnabled");
+
+		if (groupedData.length > 1 && Tools.getProperty(this.model.getOptions(), "color", "gradientEnabled")) {
+			console.error("Gradients can only be enabled when having 1 single dataset");
+		}
 		const areas = svg
```
theiliad, line 67:
```suggestion

		const areas = svg
```

packages/core/src/components/graphs/area.ts
```diff
@@ -2,12 +2,16 @@
 import { Component } from "../component";
 import * as Configuration from "../../configuration";
 import { CartesianOrientations, Events } from "../../interfaces";
+import { GradientUtils } from "../../services";
+import { Tools } from "../../tools";
 
 // D3 Imports
 import { area } from "d3-shape";
+import { select } from "d3";
```
theiliad, line 10: importing something from `d3` rather than the specific d3 package will cause all of `d3` to be imported and there will be no tree-shaking available at that point.

I believe this should say `import { select } from "d3-selection";`

packages/core/src/components/graphs/area.ts
```diff
@@ -37,18 +43,56 @@ export class Area extends Component {
...
+			&& groupedData.length === 1
+			&& Tools.getProperty(this.model.getOptions(), "color", "gradientEnabled");
+
+		if (groupedData.length > 1 && Tools.getProperty(this.model.getOptions(), "color", "gradientEnabled")) {
+			console.error("Gradients can only be enabled when having 1 single dataset");
+		}
 		const areas = svg
 			.selectAll("path.area")
 			.data(groupedData, (group) => group.name);
+		if (!this.parent.selectAll("defs linearGradient").empty()) {
```
theiliad, line 70:
```suggestion

		if (!this.parent.selectAll("defs linearGradient").empty()) {
```

packages/core/src/components/graphs/area.ts
```diff
@@ -37,18 +43,56 @@ export class Area extends Component {
...
+			this.parent
+				.selectAll("defs linearGradient")
+				.each(function() {
+					this.parentNode.remove();
+				});
+		}
+		if (isGradientAllowed) {
```
theiliad, line 77:
```suggestion

		if (isGradientAllowed) {
```

packages/core/src/services/essentials/gradient-utils.ts
```diff
@@ -0,0 +1,76 @@
...
+				color: color,
+				opacity: "0.6"
+			},
+			{
+				offset: "80%",
+				color: color,
+				opacity: "0"
+			}
+		];
+		if (is3Stops) {
```
theiliad, line 52:
```suggestion

		if (is3Stops) {
```

packages/core/src/services/essentials/gradient-utils.ts
```diff
@@ -0,0 +1,76 @@
...
+					color: color,
+					opacity: "0.6"
+				}
+			];
+		}
+		return stops;
+	}
+
+
+
```
theiliad, line 75:
```suggestion
```

packages/core/src/services/essentials/gradient-utils.ts
```diff
@@ -0,0 +1,76 @@
...
+					offset: "100%",
+					color: color,
+					opacity: "0.6"
+				}
+			];
+		}
+		return stops;
+	}
+
+
```
theiliad, line 74:
```suggestion
```

packages/core/src/services/essentials/gradient-utils.ts
```diff
@@ -0,0 +1,76 @@
...
+				{
+					offset: "100%",
+					color: color,
+					opacity: "0.6"
+				}
+			];
+		}
+		return stops;
+	}
+
```
theiliad, line 73:
```suggestion
```

theiliad (changes requested):

theiliad (approved):

sophiiae (approved):

packages/core/src/components/graphs/area.ts
```diff
@@ -37,19 +42,70 @@ export class Area extends Component {
...
 			areaGenerator
 				.x0(cartesianScales.getRangeValue(0))
 				.x1((d, i) => cartesianScales.getRangeValue(d, i))
 				.y((d, i) => cartesianScales.getDomainValue(d, i));
+			domain = this.services.cartesianScales.getMainXScale().domain();
 		}
 
 		// Update the bound data on area groups
 		const groupedData = this.model.getGroupedData();
+
```
natashadecoste, line 56: i'd just suggest using a variable for `this.model.getOptions()` so you only get the options from the model once in this block

natashadecoste (dismissed):

theiliad (changes requested):
image

theiliad (changes requested):

theiliad (approved):
💯  🎉

sophiiae (approved):

PR #782, opened 2020-09-01
feat: hide or show axis

JennChao:
### Updates
Allow user to hide or show axis.

圖片

- [ ] Demos all features
- [ ] Documented/annotated
- [ ] Matches UI/UX specs
- [ ] Accessible


JennChao:
@theiliad 
This PR is for the need in DE.
As you can see from the screenshot. We need a chart only with sparkline. No axis should be displayed.

圖片

packages/core/src/components/axes/axis.ts
```diff
@@ -162,6 +162,9 @@ export class Axis extends Component {
 			svg,
 			`g.axis.${axisPosition}`
 		);
+		if (!Tools.getProperty(options, "axes", axisPosition, "visible")) {
+			container.style("display", "none");
+		}
```
theiliad, line 167: Do we need to render an axis if it's invisible? Seems like we're just hiding it

JennChao: We still need to render part of the axis or the following line of code will break in `TwoDimensionalAxes` component.
圖片

In order to make sure all the charts still work properly, I think just hiding the axis is a better idea.

theiliad: we can keep pieces that we need, but I'm certain that we don't have to run all of the axis rendering logic when they're hidden. If we need an invisible axis ref initialized, then we keep that piece running, or we get that method to ignore the `invisibleAxisRef` when it doesn't need it

JennChao: Now in `Axis` component, we only render needed part of codes when axis is disabled. We still need to render codes related to invisible axis and zoomDomain for later usage in `TwoDimensionalAxes` component and `ZoomBar` component.

packages/core/src/components/axes/axis.ts
```diff
@@ -205,6 +205,10 @@ export class Axis extends Component {
 			scale.domain(zoomDomain);
 		}
 
+		if (!Tools.getProperty(options, "axes", axisPosition, "visible")) {
+			return;
+		}
```
theiliad, line 210: This is great, but we're still running logic that doesn't need running when we've `return`ed.

- `tickHeight`
- `fakeTick`
- `truncationNumCharacter` and all the truncation-related info
- and more

Please do a sweep above the if-statement and move below the if-statement anything that we wouldn't need if we return because of the `visible` flag.

Also, we need to add `.attr("aria-hidden", true)` to the axis ref if it's not visible.

Ideally we also wouldn't create a 2nd `invisible` axis ref and just use the original one as the invisible ref since it is exactly that

theiliad (changes requested):

theiliad:
> 圖片

Also there seems to be no demos being added with this PR. I suggest adding this demo to the storybook, and with the rest of your PRs it should fully look like what you have above

JennChao:
> > 圖片
> 
> Also there seems to be no demos being added with this PR. I suggest adding this demo to the storybook, and with the rest of your PRs it should fully look like what you have above

@theiliad I will have a storybook for sparkline demo after all the PRs have been merged. I don't think we need a storybook for enabling or disabling axis of chart since normally users use chart with axis. Only our DE sparkline requires a chart without axis.

packages/core/src/components/axes/axis.ts
theiliad: we've still got all of these things left to do^^

theiliad (changes requested):

packages/core/src/components/axes/axis.ts
```diff
@@ -182,6 +141,50 @@ export class Axis extends Component {
 			.attr("aria-hidden", true)
 			.attr("aria-label", `invisible ${axisPosition} ticks`);
 
+		if (!Tools.getProperty(options, "axes", axisPosition, "visible")) {
```
theiliad, line 144:
```suggestion
        const isAxisVisible = Tools.getProperty(options, "axes", axisPosition, "visible");
		if (!isAxisVisible) {
```

theiliad: Also, as I've requested in my 2 previous change requests, we need to add `.attr("aria-hidden", isAxisVisible);` to the axis element

theiliad: Also, any code that sets the domain of the scale should be running here before the check for visibility, for instance you're skipping this piece
image

which means for a zoomable axis you actually won't be able to zoom in if the axis is invisible.

Please make sure that all the axis related functionality still works with an invisible axis.

theiliad (changes requested):

packages/core/src/components/axes/axis.ts
JennChao: When axis is disabled, we shouldn't let the user to zoom in. The main purpose of zooming in is to have have more detail information about the data. However, if the user has already disabled axis, the chart would no longer serve as a detail provider. So I don't think providing zoom functionality to chart without axis is a good idea. What do you think? @theiliad

theiliad (approved):
@JennChao would custom domains break if axis is invisible?

image

Would any of the existing functionality break if axes are invisible? Just checking for the last time, thanks.

JennChao:
> @JennChao would custom domains break if axis is invisible?
> 
> image
> 
> Would any of the existing functionality break if axes are invisible? Just checking for the last time, thanks

According to what I have tested so far, nothing would break

theiliad (approved):

sophiiae (approved):

PR #787, opened 2020-09-03
fix the defect of label tooltip not showing when using custom tooltip

linhenry0417:
The tooltip for axis labels is not rendering when a custom tooltip HTML is provided in options.

When the user provides a custom tooltip HTML for the chart, the tooltip to show the full truncated text in the vertical axis is not rendering:
This PR is to fix this.
imagefffff


linhenry0417:
@theiliad @sophiiae This should be a quick fix to the issue described above. Thanks.

theiliad (approved):

sophiiae (approved):

natashadecoste (approved):

PR #789, opened 2020-09-07
Sparkline loading

JennChao:
### Updates
specify option for sparking loading state in line chart. Sparkline loading state can only be used when there is only one dataset.

sparkline-loading

- [ ] Demos all features
- [ ] Documented/annotated
- [ ] Matches UI/UX specs
- [ ] Accessible


theiliad (changes requested):
Hi, just a bit confused here.

You're building out a loading state for sparkline graph types, now in the demo you've got full axis title and tick and legend items.

The point of the loading state is that your application data has not loaded yet, therefore you do not know your axis values, your legend items or what your line will look like.

At which point I don't see a reason for deviating from the traditional skeleton state that we have for axis charts.

We can definitely rediscuss this on our guild calls (possibly Josh can bring the topic up if we've missed something here).

I'm going to close for now.

PR #790, opened 2020-09-07
Sparkline demo

JennChao:
### Updates
This PR is created for demo only. You can run storybook for more detail.
sparkline loading demo: path=/story/line--sparkline-loading
sparkline with data demo: path=/story/area--area-time-series-as-sparkline

圖片

圖片

- [ ] Demos all features
- [ ] Documented/annotated
- [ ] Matches UI/UX specs
- [ ] Accessible


jeanservaas:
@JennChao 

I'm wondering if the gradient can be slightly subtler / less intense. In the preview it looks pretty heavy. I'm also wondering what it would look like on a less tall line chart.

image

JennChao:
> @JennChao
> 
> I'm wondering if the gradient can be slightly subtler / less intense. In the preview it looks pretty heavy. I'm also wondering what it would look like on a less tall line chart.
> 
> Just to be clearer, I'd end the gradient bit sooner and more gradually to give it a lighter touch. Also I just want to make sure that whoever on the security team provided design specs for this will get to have a look at the demo and add a review too.
> 
> image

Hi @jeanservaas 
I adjusted the gradient offset value and opacity. The gradient style is now less intense. 
圖片


The following screenshot is how the line chart would look like when the height is half of the height of the screenshot you provided.
圖片

And I had a meeting with Josh, the designer of IBM Security for Data Explorer, he is fine with our current implementation.

jeanservaas (commented):
Hey @JennChao 

Sorry don't want to hold you guys up. I think the new color intensity is good. What I was referring to is the length of the gradient fade too:


image


I was wondering in code, if you could stretch the fade a little further so it truncates more subtley like the example above... rather than in that abrupt horizontal line. This might requiring pulling the gradient down a little further below the spark line. You actually have to treat it like a transparent area shape, rather than just a line.

Let me know if this is possible. If not, and your designer is OK with the demo I will approve.

theiliad:
@JennChao do we still need this PR?

theiliad:
Confirmed with @JennChao, this is no longer needed

PR #717, opened 2020-07-22
Tool bar above zoom bar and shift-clicking on chart

JennChao:
### Updates
- For [#380](https://github.com/carbon-design-system/carbon-charts/issues/380)

### Demo screenshot or recording
圖片

### Review checklist (for reviewers only)
- [ ] Demos all features
- [ ] Documented/annotated
- [ ] Matches UI/UX specs
- [ ] Meets the [code style guide](https://github.com/carbon-design-system/carbon-charts/wiki/Code-style-guide)
- [ ] Accessible
- [ ] Mobile first (responsive)
- [ ] RTL support (bidirectional text)
- [ ] Performant (limited bloat)

PR #791, opened 2020-09-08
Toolbar 

hlyang397:
### Updates
- Based on PR #717 
- Toolbar
  - zoom-in, zoom-out button
    - disabled state
  - overflow menu button
    - reset zoom
    - disabled state
- Toolbar demo group

image

- [ ] Demos all features
- [ ] Documented/annotated
- [ ] Matches UI/UX specs
- [ ] Accessible


packages/core/src/configuration.ts
```diff
@@ -151,11 +154,22 @@ const axisChart: AxisChartOptions = Tools.merge({}, chart, {
...
+	} as ZoomBarsOptions,
+	toolbar: {
+		enabled: false,
+		overflowMenuItems: {
+			resetZoom: {
+				enabled: false,
+				text: "Reset zoom"
+			} as ToolbarOverflowMenuItemOptions
+		} as ToolbarOverflowMenuItems
+	} as ToolbarOptions
```
theiliad, line 172: So what's been discussed with Josh Kimmel and the rest of the dataviz guild would work a bit differently than specified here.

**The user would need a way of defining how many icons to show and how many to move into the overflow**.

This means that each item should have a name & icon, and the items can no longer be called `overflow` items but more so just `controls` in general.

That way the user can define a max number (e.g. 2) which means that the zoom in & zoom out buttons would show up as icons and everything else would go inside the overflow menu.

Additionally, the user needs to have a way of changing the order (e.g. define the order as `zoom in`, `zoom out`, `reset zoom`, `make fullscreen`, `print` and choose the limit as 3 which'll mean they'll only see zoom in, zoom out & reset zoom as icons)

Not sure if this is the best route to go but it's what design is asking for. I'll tag @jeanservaas here for feedback

theiliad (changes requested):

jeanservaas:
Yes I think this is what we agreed upon in the call... The user can basically arrange the controls they way they want, but we have some best practice suggestions

**The things we strongly encourage:**

1) if you have two or three controls, they will appear as icons in the upper right corner, not in an overflow. 
2) If you have four or more controls, 2+ controls can appear in the overflow menu

**Things we definitely want to avoid:**

— Having only 1 item in an overflow menu
 — Splitting up dependent controls like zoom up/zoom down (i.e. don't ever have a situation where one is in an overflow menu and one is not)

hlyang397:
@theiliad @jeanservaas 
I have updated this PR based on the new spec you mentioned.
Please help to review it, thanks.

theiliad:
> @theiliad @jeanservaas
> I have updated this PR based on the new spec you mentioned.
> Please help to review it, thanks.

Hi, it seems like deploy previews aren't working and Travis is failing. Might be some errors that we'd need to fix in the code.

You should be seeing some errors when running or building storybook locally.

hlyang397:
> Hi, it seems like deploy previews aren't working and Travis is failing. Might be some errors that we'd need to fix in the code.
> 
> You should be seeing some errors when running or building storybook locally.

@theiliad 
Thanks for catching this.
The local storybook is good, but I import `src` folder in storybook demo folder, and that won't work in `dist/`.
I have fixed this in latest commit.
Please let me know if you find any other issue.

theiliad (changes requested):
We need the controls in the toolbar to be keyboard accessible (user should be able to tab through), and possibly proper aria labeling

hlyang397:
> 💯
> We need the controls in the toolbar to be keyboard accessible (user should be able to tab through), and possibly proper aria labeling. Please see the legend items as an example in the demos

@theiliad Can Carbon Charts includes Carbon components as library ?
Toolbar controls (button, overflow menu) are highly related to carbon design system, and have dependency on carbon components. It makes no sense to reinvent the wheel.

theiliad:
> > 💯
> > We need the controls in the toolbar to be keyboard accessible (user should be able to tab through), and possibly proper aria labeling. Please see the legend items as an example in the demos
> 
> @theiliad Can Carbon Charts includes Carbon components as library ?
> Toolbar controls (button, overflow menu) are highly related to carbon design system, and have dependency on carbon components. It makes no sense to reinvent the wheel.

`carbon-components` are HTML components, Carbon Charts components are written in SVG. You can't really embed HTML inside SVG components (there are hacks but not recommended).

Overflow menu can be used from `carbon-components`, since it's an overlay element, however anything that is inside the chart container (AKA inside the chart container SVG element) can only be written in SVG

packages/core/src/styles/components/_toolbar.scss
```diff
@@ -0,0 +1,80 @@
...
+				}
+				rect {
+					fill: $disabled-03;
+				}
+				path {
+					fill: $disabled-03;
+				}
+				polygon {
+					fill: $disabled-03;
+				}
```
theiliad, line 63:
```suggestion
				circle, rect, path, polygon {
					fill: $disabled-03;
				}
```

packages/core/src/styles/components/_toolbar.scss
```diff
@@ -0,0 +1,80 @@
...
+				rect {
+					fill: $icon-02;
+				}
+				path {
+					fill: $icon-02;
+				}
+				polygon {
+					fill: $icon-02;
+				}
```
theiliad, line 43:
```suggestion
				circle, rect, path, polygon {
					fill: $icon-02;
				}
```

packages/core/src/styles/components/_toolbar.scss
```diff
@@ -0,0 +1,80 @@
...
+					.toolbar-button-background {
+						stroke: $icon-02;
+						stroke-width: 2;
+						stroke-dasharray: 2;
+					}
+				}
+			}
+
+			&-icon {
+				pointer-events: visible;
```
theiliad, line 31:
```suggestion
				pointer-events: visible;

```

packages/core/src/styles/components/_toolbar.scss
```diff
@@ -0,0 +1,80 @@
...
+					fill: $icon-02;
+				}
+				&:hover {
+					cursor: pointer;
+					pointer-events: visible;
+				}
+			}
+
+			&-icon--disabled {
+				pointer-events: auto;
```
theiliad, line 51:
```suggestion
				pointer-events: auto;

```

packages/core/src/styles/components/_toolbar.scss
```diff
@@ -0,0 +1,80 @@
...
+@import "./../vendor/carbon-components/src/components/overflow-menu/overflow-menu";
+
+g.#{$prefix}--#{$charts-prefix}--toolbar {
+	@include carbon--motion(standard, expressive);
+
+	.toolbar-container {
+		.toolbar-button {
+			fill: $ui-background;
+			cursor: pointer;
+			&:hover {
```
theiliad, line 11:
```suggestion

			&:hover {
```

packages/core/src/configuration.ts
```diff
@@ -170,11 +173,30 @@ const axisChart: AxisChartOptions = Tools.merge({}, chart, {
...
 		top: {
 			enabled: false,
 			type: ZoomBarTypes.GRAPH_VIEW
 		}
-	} as ZoomBarsOptions
+	} as ZoomBarsOptions,
+	toolbar: {
+		enabled: false,
+		maxIcons: 3,
+		controlsInOrder: [
```
theiliad, line 185:
```suggestion
		controls: [
```

packages/core/src/configuration.ts
```diff
@@ -170,11 +173,30 @@ const axisChart: AxisChartOptions = Tools.merge({}, chart, {
...
+			} as ToolbarControl,
+			{
+				type: ToolbarControlTypes.ZOOM_OUT,
+				text: "Zoom out"
+			} as ToolbarControl,
+			{
+				type: ToolbarControlTypes.RESET_ZOOM,
+				text: "Reset zoom"
+			} as ToolbarControl
+		]
```
theiliad, line 198: does this default to anything?
What if someone doesn't provide a `text` value for `ToolbarControlTypes.ZOOM_IN`? would it default to `Zoom in`?

packages/core/demo/data/toolbar.ts
```diff
@@ -0,0 +1,58 @@
...
+		top: {
+			enabled: true
+		}
+	};
+	if (configs) {
+		if (configs.titlePostifx) {
+			options.title += configs.titlePostifx;
+		}
+		if (configs.maxIcons) {
+			options.toolbar.maxIcons = configs.maxIcons;
```
theiliad, line 21:
```suggestion
			options.toolbar.numberOfIcons = configs.numberOfIcons;
```

theiliad (changes requested):

packages/core/src/styles/components/_toolbar.scss
hlyang397: Updated in https://github.com/carbon-design-system/carbon-charts/pull/791/commits/56b3c20d8ac2990a06938ac3cd80d0d54c8bde20.

packages/core/src/styles/components/_toolbar.scss
hlyang397: Updated in https://github.com/carbon-design-system/carbon-charts/pull/791/commits/56b3c20d8ac2990a06938ac3cd80d0d54c8bde20.

packages/core/src/styles/components/_toolbar.scss
hlyang397: Updated in https://github.com/carbon-design-system/carbon-charts/pull/791/commits/56b3c20d8ac2990a06938ac3cd80d0d54c8bde20.

packages/core/src/styles/components/_toolbar.scss
hlyang397: Updated in https://github.com/carbon-design-system/carbon-charts/pull/791/commits/56b3c20d8ac2990a06938ac3cd80d0d54c8bde20.

packages/core/demo/data/toolbar.ts
hlyang397: Updated in https://github.com/carbon-design-system/carbon-charts/pull/791/commits/f6a4a5976a3106e01e7907c24c082647153b9858.

packages/core/src/configuration.ts
hlyang397: Updated in https://github.com/carbon-design-system/carbon-charts/pull/791/commits/f6a4a5976a3106e01e7907c24c082647153b9858.

packages/core/src/configuration.ts
hlyang397: No, I expected user to provide the required `text`.
But in latest https://github.com/carbon-design-system/carbon-charts/pull/791/commits/ad3c9b0b6f6955f93c3daac07210106060739be9, I check the `text`, and use `type` if `text` is not available.

BTW, I also remove the default ToolbarControls in configuration because if will provide extra control setting if user provides less control configuration. (We use `Tools.merge` in `model.setOptions`)

jeanservaas (commented):
@theiliad quick comment about the spacing:

the toolbar icons seem to be riding a little bit low, they should be vertically centered with the Chart title. Otherwise looks good.

image

hlyang397:
> @theiliad quick comment about the spacing:
> 
> the toolbar icons seem to be riding a little bit low, they should be vertically centered with the Chart title. Otherwise looks good.
> 

@theiliad Does LayoutComponent support vertical alignment center ?

hlyang397:
> 💯
> We need the controls in the toolbar to be keyboard accessible (user should be able to tab through), and possibly proper aria labeling. Please see the legend items as an example in the demos

@theiliad The keyboard support is included in latest code, please help to review it, thanks.

jeanservaas (approved):
LGTM! Thanks for the update

packages/core/src/components/axes/toolbar.ts
```diff
@@ -0,0 +1,495 @@
...
+		// assume the overflow icon has right alignment in layout
+		this.overflowMenuX = width - overflowMenuWidth;
+		this.overflowMenuY = parentY + VERTICAL_ALIGNMENT_Y + buttonSize;
+		const container = DOMUtils.appendOrSelect(svg, "svg.toolbar-container")
+			.attr("width", "100%")
+			.attr("height", Configuration.toolbar.height)
+			.attr("opacity", 1);
+
+		// clean children first
+		container.html(null);
```
theiliad, line 108: hmmmm... so the toolbar cleans everything everytime? Why not render more efficiently through d3?

theiliad (changes requested):
added comments above

packages/core/src/services/zoom.ts
```diff
@@ -152,4 +153,143 @@ export class Zoom extends Service {
...
+		if (
+			currentZoomDomain &&
+			maxZoomDomain &&
+			currentZoomDomain[0].valueOf() === maxZoomDomain[0].valueOf() &&
+			currentZoomDomain[1].valueOf() === maxZoomDomain[1].valueOf()
+		) {
+			return true;
+		} else {
+			return false;
+		}
```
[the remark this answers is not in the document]
hlyang397: Updated in https://github.com/carbon-design-system/carbon-charts/pull/791/commits/4c136d2b9735cc4b8ff817bc8279b03915ef7675.

packages/core/src/services/zoom.ts
```diff
@@ -152,4 +153,143 @@ export class Zoom extends Service {
...
+			this.model.getOptions(),
+			"zoomBar",
+			"minZoomRatio"
+		);
+		// if current zoom domain is already smaller than minZoomRatio
+		if (currentZoomDomainPeriod / maxZoomDomainPeriod < minZoomRatio) {
+			return true;
+		} else {
+			return false;
+		}
```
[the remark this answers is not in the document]
hlyang397: Updated in https://github.com/carbon-design-system/carbon-charts/pull/791/commits/4c136d2b9735cc4b8ff817bc8279b03915ef7675.

packages/core/src/interfaces/components.ts
```diff
@@ -122,6 +123,40 @@ export interface StackedBarOptions extends BarOptions {
...
+	/**
+	 * the maximum toolbar controls to be displayed as icons
+	 * controls more than this number will appear in the overflow menu
+	 * minimum is 1. (all toolbar controls are in overflow menu)
+	 */
+	maxIcons?: number;
+	/**
+	 * toolbar controls which will be displayed following the array order
+	 */
+	controlsInOrder?: ToolbarControl[];
```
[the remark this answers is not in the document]
hlyang397: Thanks for catching this. I wonder if there is any tool could help to catch this error.
Updated in https://github.com/carbon-design-system/carbon-charts/pull/791/commits/4c136d2b9735cc4b8ff817bc8279b03915ef7675.

packages/core/demo/data/toolbar.ts
```diff
@@ -0,0 +1,70 @@
...
+			text: "Zoom out"
+		},
+		{
+			type: "Reset zoom",
+			text: "Reset zoom"
+		}
+	];
+
+	if (configs) {
+		if (configs.titlePostifx) {
```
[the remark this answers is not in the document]
hlyang397: Thanks for catching this.
Updated in https://github.com/carbon-design-system/carbon-charts/pull/791/commits/4c136d2b9735cc4b8ff817bc8279b03915ef7675.

packages/core/src/interfaces/components.ts
```diff
@@ -122,6 +123,40 @@ export interface StackedBarOptions extends BarOptions {
...
+	/**
+	 * is the toolbar visible or not
+	 */
+	enabled?: boolean;
+	/**
+	 * the maximum toolbar controls to be displayed as icons
+	 * controls more than this number will appear in the overflow menu
+	 * minimum is 1. (all toolbar controls are in overflow menu)
+	 */
+	maxIcons?: number;
```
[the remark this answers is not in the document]
hlyang397: Updated in https://github.com/carbon-design-system/carbon-charts/pull/791/commits/4c136d2b9735cc4b8ff817bc8279b03915ef7675.

packages/core/demo/data/toolbar.ts
```diff
@@ -0,0 +1,70 @@
...
+			options.title += configs.titlePostifx;
```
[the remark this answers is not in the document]
hlyang397: Updated in https://github.com/carbon-design-system/carbon-charts/pull/791/commits/4c136d2b9735cc4b8ff817bc8279b03915ef7675.

packages/core/demo/data/toolbar.ts
```diff
@@ -0,0 +1,70 @@
...
+			type: "Zoom in",
+			text: "Zoom in"
+		},
+		{
+			type: "Zoom out",
+			text: "Zoom out"
+		},
+		{
+			type: "Reset zoom",
+			text: "Reset zoom"
```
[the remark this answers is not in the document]
hlyang397: No, `text` field is not required and it's defaulted to `type`.
This is to demo how to change the toolbar control text.
The other story shows that it also works without `text` field.
Let me know if you like to remove the `text` field in demo data as well.
Update interface in https://github.com/carbon-design-system/carbon-charts/pull/791/commits/1c7e6cae39b7460cd6a918c8645de84e77b2251c.

packages/core/src/components/axes/toolbar.ts
```diff
@@ -0,0 +1,495 @@
...
+	isOverflowMenuOpen() {
+		return (
+			this.overflowMenu
+				.selectAll("ul.bx--overflow-menu-options--open")
+				.size() > 0
+		);
+	}
+
+	// show/hide overflow menu
+	showOverflowMenu(show: boolean) {
```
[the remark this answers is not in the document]
hlyang397: @theiliad 
Yes, we use the boolean parameter to decide if we should show or hide the overflow menu.
I am not pretty sure what's your preference.
Do you like to have two separate functions (`showOverflowMenu` and `hideOverflowMenu`) ?
Or any other preferred function name ?

packages/core/src/components/axes/toolbar.ts
```diff
@@ -0,0 +1,495 @@
...
+					overflowMenuItems[nextItemIndex].id +
+						this.overflowMenuItemId
+				)
+				.focus();
+		}
+	}
+
+	toggleOverflowMenu() {
+		if (this.isOverflowMenuOpen()) {
+			// hide overflow menu
```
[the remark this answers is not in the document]
hlyang397: Please refer to https://github.com/carbon-design-system/carbon-charts/pull/791#discussion_r504465304.

packages/core/src/components/axes/toolbar.ts
```diff
@@ -0,0 +1,495 @@
...
+			return {
+				buttonList: controlList
+			};
+		} else {
+			return {
+				// leave one button for overflow button
+				buttonList: controlList.splice(0, numberOfIcons - 1),
+				overflowMenuItemList: controlList
+			};
+		}
```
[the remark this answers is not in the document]
hlyang397: Updated in https://github.com/carbon-design-system/carbon-charts/pull/791/commits/4c136d2b9735cc4b8ff817bc8279b03915ef7675.

packages/core/src/components/axes/toolbar.ts
```diff
@@ -0,0 +1,495 @@
...
+			"toolbar",
+			"controls"
+		);
+		const controlList = [];
+		controls.forEach((control) => {
+			let controlConfig;
+			switch (control.type) {
+				case ToolbarControlTypes.ZOOM_IN:
+					if (isZoomBarEnabled) {
+						controlConfig = this.getZoomInConfig();
```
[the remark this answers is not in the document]
hlyang397: @theiliad 
I don't get your point.
Even you have a `getControlConfigs(control: ToolbarControlTypes)` method, you still need a `switch` in the method to return respective config based on `ToolbarControlTypes`.
Could you provide a pseudo code or example ?

packages/core/src/components/axes/toolbar.ts
hlyang397: Yes, we need to clean buttons if data is loading, or update toolbar controls if configuration is changed.
We could use d3, but I don't think it could help to reduce the code complexity because the buttons and overflow menu items are more related to configurations instead of simple data, and for each configuration we need to create complex html component instead of SVG.
Pleasee let me know if you still think we should change to use d3.

hlyang397:
> added comments above

@theiliad I have updated PR based on your comments. Please help to review again, thanks.

packages/core/src/components/axes/toolbar.ts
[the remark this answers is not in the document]
theiliad: Hi, yes what I'm trying to avoid is a `getZoomInConfig()` call for a small object return. This'll result in too many methods in the future.

If we just have a `getControlConfigs()` that includes all the configs and returns the 1 we care about, then it's simpler

theiliad (changes requested):

packages/core/src/components/axes/toolbar.ts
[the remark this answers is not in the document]
hlyang397: Updated in https://github.com/carbon-design-system/carbon-charts/pull/791/commits/bcee77bd714f8fab66157934d5178c6660e59e06.
Please let me know if this is not what you expected, thanks.

hlyang397:
@theiliad Any other comment on this PR?

packages/core/src/components/axes/toolbar.ts
```diff
@@ -0,0 +1,489 @@
...
+		for (let i = currentItemIndex - 1; i >= 0; i--) {
+			const previousOverflowMenuItem = overflowMenuItems[i];
+			if (!previousOverflowMenuItem.shouldBeDisabled()) {
+				previousItemIndex = i;
+				break;
+			}
+		}
+		// only if previous enabled menu item found
+		if (previousItemIndex < overflowMenuItems.length) {
+			document
```
theiliad, line 248: can this be done with a d3 selection?

packages/core/src/components/axes/toolbar.ts
```diff
@@ -0,0 +1,489 @@
...
+			const nextOverflowMenuItem = overflowMenuItems[i];
+			if (!nextOverflowMenuItem.shouldBeDisabled()) {
+				nextItemIndex = i;
+				break;
+			}
+		}
+		// only if next enabled menu item found
+		if (nextItemIndex > -1) {
+			document
+				.getElementById(
```
theiliad, line 270: can this be done with a d3 selection?

packages/core/src/components/axes/toolbar.ts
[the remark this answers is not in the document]
theiliad: My concern is the name of the function. Looking at the name `showOverflowMenu` I'd assume that calling it would `show` the overflow menu. Maybe we need to call it `updateOverflowMenu`

packages/core/src/components/axes/toolbar.ts
```diff
@@ -0,0 +1,489 @@
...
+
+		// size of toolbar button with background
+		const buttonSize = Configuration.toolbar.buttonSize;
+		const parentY = parseFloat(this.parent.node().getAttribute("y"));
+
+		const svg = this.getContainerSVG();
+
+		// TODO -- adjust toolbar Y position to align title component
+		// before layout component supports vertical alignment center
+		const VERTICAL_ALIGNMENT_Y = -6;
```
theiliad, line 81:
```suggestion
		const Y_OFFSET = -6;
```

theiliad (changes requested):

packages/core/src/components/axes/toolbar.ts
hlyang397: Updated in https://github.com/carbon-design-system/carbon-charts/pull/791/commits/ffb41bcd7af6b329ba8ba5538c427491ad32453a.

packages/core/demo/data/toolbar.ts
```diff
@@ -0,0 +1,70 @@
...
+		}
+	];
+
+	if (configs) {
+		if (configs.titlePostfix) {
```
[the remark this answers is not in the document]
hlyang397: Updated in https://github.com/carbon-design-system/carbon-charts/pull/791/commits/ffb41bcd7af6b329ba8ba5538c427491ad32453a.

packages/core/src/components/axes/toolbar.ts
hlyang397: Updated in https://github.com/carbon-design-system/carbon-charts/pull/791/commits/70f749775ac447223e68fdf3e387b11e8cd07297.

packages/core/src/components/axes/toolbar.ts
hlyang397: Updated in https://github.com/carbon-design-system/carbon-charts/pull/791/commits/70f749775ac447223e68fdf3e387b11e8cd07297.

hlyang397:
@theiliad  Updated based on your comments.
Please help to review again, thanks.

theiliad (approved):

natashadecoste (approved):
Nice
💯 for documention too

sophiiae (approved):

PR #793, opened 2020-09-09
feat: keep initialZoomDomain in model to detect change

hlyang397:
### Updates
When user changes the initialZoomDomain option, the chart zoom domain should be updated accordingly.

No UI change.

- [ ] Demos all features
- [ ] Documented/annotated
- [ ] Matches UI/UX specs
- [ ] Accessible


packages/core/src/components/axes/zoom-bar.ts
```diff
@@ -128,6 +114,41 @@ export class ZoomBar extends Component {
 			// add value 0 to the extended domain for zoom bar area graph
 			this.compensateDataForDefaultDomain(zoomBarData, defaultDomain);
 
+			const initialZoomDomainInModel = this.model.get(
```
theiliad, line 117: what about calling them new and old?

theiliad (changes requested):

packages/core/src/components/axes/zoom-bar.ts
hlyang397: Good point, updated in https://github.com/carbon-design-system/carbon-charts/pull/793/commits/9c1323a6a6807c1c6e3469569bed5069fcbefd6f.
Thanks.

theiliad (approved):

sophiiae (approved):

natashadecoste (approved):

PR #829, opened 2020-10-02
fix(core): fix division by zero & timescale issues

theiliad:
Fixes 2 issues:
- Timescale was missing parsing of JS-dates in 1 place
- Division by zero was being allowed in the stacked area chart


natashadecoste:
🥳

natashadecoste (approved):

cal-smith (approved):

PR #833, opened 2020-10-06
fix(core): fix Legend and Tooltip group color

icaro1508:
Fix Legend and Tooltip Group colors to be based on the getFillColor option as well

### Updates
- Update Legend fill color function to take into consideration custom fill and stroke colors
- Update Tooltip 'Group' fill color function to take into consideration custom fill and stroke colors
- Fixes issue #828 
### Demo screenshot or recording
image

### Review checklist (for reviewers only)
- [ ] Demos all features
- [ ] Documented/annotated
- [ ] Matches UI/UX specs
- [ ] Meets the [code style guide](https://github.com/carbon-design-system/carbon-charts/wiki/Code-style-guide)
- [ ] Accessible
- [ ] Mobile first (responsive)
- [ ] RTL support (bidirectional text)
- [ ] Performant (limited bloat)

PR #834, opened 2020-10-06
fix(core): fix Legend and Tooltip group color

### LINKED ISSUE #828: Group legend and group tooltip colors don't match getFillColor defined color
#### I am submitting a...

- [ ] Feature request
- [ ] Design defect
- [x] Source code defect
- [ ] Other

#### charts version:

@carbon/charts: "0.38.0"
@carbon/charts-react: "0.38.0"

### Issue description

Group color in tooltip and legend doesn't match the color provided by the `BaseChartOptions.getFillColor`

- Create a new CRA project
- Install `carbon-components` and `@carbon/charts-react`
- Import chart styles
- Create new Component that uses some chart (I only tested the `SimpleBarChart` component)
- Set `getFillColor` function to return a static value (in the CodeSandbox example `rgb(110, 50, 201)`)
- Hover over bars for tooltip or check group color in legend

#### Current behavior

Group legend and group in tooltip colors don't match fill color defined by `getFillColor` function

#### Expected behavior

Group legend and group in tooltip colors **should** match fill color defined by `getFillColor` function

### Screenshot or recording

bug_tooltips

### Link to Running Demo

[CodeSandBox](https://codesandbox.io/s/dreamy-volhard-sxx0s)

icaro1508:
Fix Legend and Tooltip Group colors to be based on the getFillColor option as well

fix #828

### Updates
- Update Legend fill color function to take into consideration custom fill and stroke colors
- Update Tooltip 'Group' fill color function to take into consideration custom fill and stroke colors

demo_picture

- [ ] Demos all features
- [ ] Documented/annotated
- [ ] Matches UI/UX specs
- [ ] Accessible

### DIFF
--- a/packages/core/src/components/essentials/legend.ts
+++ b/packages/core/src/components/essentials/legend.ts
@@ -82,7 +82,7 @@ export class Legend extends Component {
 			.attr("ry", 1)
 			.style("fill", (d) => {
 				return d.status === Configuration.legend.items.status.ACTIVE
-					? this.model.getStrokeColor(d.name)
+					? this.model.getColorForGroup(d.name)
 					: null;
 			})
 			.classed("active", function (d, i) {
--- a/packages/core/src/components/essentials/legend.ts
+++ b/packages/core/src/components/essentials/legend.ts
@@ -82,7 +82,7 @@ export class Legend extends Component {
 			.attr("ry", 1)
 			.style("fill", (d) => {
 				return d.status === Configuration.legend.items.status.ACTIVE
-					? this.model.getColorForGroup(d.name)
+					? this.model.getFillColor(d.name)
 					: null;
 			})
 			.classed("active", function (d, i) {
@@ -231,8 +231,8 @@ export class Legend extends Component {
 
 					if (
 						startingPoint +
-							spaceNeededForCheckbox +
-							legendItemTextDimensions.width >
+						spaceNeededForCheckbox +
+						legendItemTextDimensions.width >
 						svgDimensions.width
 					) {
 						lineNumber++;
--- a/packages/core/src/model.ts
+++ b/packages/core/src/model.ts
@@ -454,16 +454,6 @@ export class ChartModel {
 		return this.colorScale;
 	}
 
-	getColorForGroup(group: any, key?: any, data?: any) {
-		const options = this.getOptions()
-		const shouldUseFillColorForGroup = options.getFillColor && !options.getStrokeColor
-		if (shouldUseFillColorForGroup) {
-			return this.getFillColor(group, key, data)
-		} else {
-			return this.getStrokeColor(group, key, data)
-		}
-	}
-
 	/**
 	 * For charts that might hold an associated status for their dataset
 	 */
@@ -581,7 +571,7 @@ export class ChartModel {
 		// Get group status based on items in selected groups
 		const getStatus = (groupName) =>
 			!options.data.selectedGroups.length ||
-				options.data.selectedGroups.includes(groupName)
+			options.data.selectedGroups.includes(groupName)
 				? ACTIVE
 				: DISABLED;
 
--- a/packages/core/src/model.ts
+++ b/packages/core/src/model.ts
@@ -454,6 +454,16 @@ export class ChartModel {
 		return this.colorScale;
 	}
 
+	getColorForGroup(group: any, key?: any, data?: any) {
+		const options = this.getOptions()
+		const shouldUseFillColorForGroup = options.getFillColor && !options.getStrokeColor
+		if (shouldUseFillColorForGroup) {
+			return this.getFillColor(group, key, data)
+		} else {
+			return this.getStrokeColor(group, key, data)
+		}
+	}
+
 	/**
 	 * For charts that might hold an associated status for their dataset
 	 */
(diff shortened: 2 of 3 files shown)


icaro1508:
I opened PR #833 but that had a unecessary double negative in it, since i Didn't really want to do another commit just for that i reopened the original PR without that, sorry if I did something wrong.

packages/core/src/model.ts
```diff
@@ -454,6 +454,16 @@ export class ChartModel {
 		return this.colorScale;
 	}
 
+	getColorForGroup(group: any, key?: any, data?: any) {
```
theiliad, line 457: We already have a `getFillColor` in model that'll call `options.getFillColor` if it exists

packages/core/src/model.ts
```diff
@@ -454,6 +454,16 @@ export class ChartModel {
...
+		const options = this.getOptions()
+		const shouldUseFillColorForGroup = options.getFillColor && !options.getStrokeColor
+		if (shouldUseFillColorForGroup) {
+			return this.getFillColor(group, key, data)
+		} else {
+			return this.getStrokeColor(group, key, data)
```
theiliad, line 463: `getFillColor` is used in places where we'd like to determine the `fill` value of an element, it'd be misleading if the `fill` value is set by a stroke color value that's being provided. We probably don't need this line.

image

packages/core/src/components/essentials/legend.ts
```diff
@@ -82,7 +82,7 @@ export class Legend extends Component {
 			.attr("ry", 1)
 			.style("fill", (d) => {
 				return d.status === Configuration.legend.items.status.ACTIVE
-					? this.model.getStrokeColor(d.name)
+					? this.model.getColorForGroup(d.name)
```
theiliad, line 85: this should've been a `getFillColor` call, not sure why it was `getStrokeColor`

theiliad (changes requested):
Hi 👋 

Thank you for contributing this piece, however I'm not sure what this PR is achieving. Could you please walk me through the use-case here? Thanks

theiliad:
The `getStrokeColor` calls in legend look like a mistake. They should've been `getFillColor` calls instead.

`getStrokeColor` possibly should only be used in scatter & bubble graphs

icaro1508:
> Hi 
> 
> Thank you for contributing this piece, however I'm not sure what this PR is achieving. Could you please walk me through the use-case here? Thanks

Gladly! the problem, as described in issue #828 is related to the user of `getStrokeColor` for `Legend` and `Tooltip` components, especially when dealing with graphs that don't necessarily have a stroke color (e.g.: Bar graphs).

Use case is: whoever is using the component defines a custom fillColor for a graph that doesn't have a stroke and that should reflect both on the Group color in the tooltip component and in the checkbox for the graph legend.

The provided code was more a fix related to these cases, falling back to the original behaviour if a `getStrokeColor` function is provided.

I did it this way because after looking through the code I couldn't really find a simple way to define if a graph should take into consideration a provided stroke color or not when painting the tooltip and legend.

icaro1508:
> The `getStrokeColor` calls in legend look like a mistake. They should've been `getFillColor` calls instead.
> 
> `getStrokeColor` possibly should only be used in scatter & bubble graphs

I wholeheartedly agree :) just couldn't find a way to make it work without messing around with the code too much and I was a bit scared since it's my first PR ever, if you have any suggestions I'd be more than glad to implement them and refactor what I did!

My idea was a way to determine for the different types of graphs (when creating the model instance) if it should use the fill color or stroke color for the components modified by the PR.

packages/core/src/model.ts
icaro1508: Makes sense, I realized this after opening the PR that in the cases where it would need to use the Stroke Color and the chart options had a fillColor defined it wouldn't really work, any ideas on how to overcome this?

theiliad:
> > Hi
> > Thank you for contributing this piece, however I'm not sure what this PR is achieving. Could you please walk me through the use-case here? Thanks
> 
> Gladly! the problem, as described in issue #828 is related to the user of `getStrokeColor` for `Legend` and `Tooltip` components, especially when dealing with graphs that don't necessarily have a stroke color (e.g.: Bar graphs).
> 
> Use case is: whoever is using the component defines a custom fillColor for a graph that doesn't have a stroke and that should reflect both on the Group color in the tooltip component and in the checkbox for the graph legend.
> 
> The provided code was more a fix related to these cases, falling back to the original behaviour if a `getStrokeColor` function is provided.
> 
> I did it this way because after looking through the code I couldn't really find a simple way to define if a graph should take into consideration a provided stroke color or not when painting the tooltip and legend.

Yes I think the real bug here is that nothing but the stroke color for bubble & scatter should be calling `getStrokeColor`, if we do switch everything else back to calling `getFillColor`, then the issues should be resolved. Does that work?

icaro1508:
> 
> Yes I think the real bug here is that nothing but the stroke color for bubble & scatter should be calling `getStrokeColor`, if we do switch everything else back to calling `getFillColor`, then the issues should be resolved. Does that work?

Should work! should I open another PR with just this or do you want to do this yourself since it's a really simple change? At your service :)

icaro1508:
> Edit: actually, my main concern around this is that, from the examples for the bubble and scatter graphs, the fill color seems to have an alpha or at least be considerably less opaque than the stroke color, can't this create some weird looking legend components?

Just checked the code on this and it doesn't look like it'll be an issue, updated the PR changing the calls for `getStrokeColor` in Legend and Tooltip to use `getFillColor` instead.

theiliad (approved):

sophiiae (approved):

PR #837, opened 2020-10-12
Sparkline loading

JennChao:
### Updates
Loading state for sparkline

圖片

- [ ] Demos all features
- [ ] Documented/annotated
- [ ] Matches UI/UX specs
- [ ] Accessible


packages/core/demo/data/line.ts
```diff
@@ -39,6 +72,42 @@ export const lineOptions = {
...
+	legend: {
+		enabled: false
+	},
+	tooltip: {
+		enabled: false
+	},
+	points: {
+		enabled: false
+	},
+	sparklineLoading: true
```
theiliad, line 108: can this work the same way as other components? Essentially you should be using the same loading flag here as a bar or line chart & it would load up the correct loading state based on the chart type.

theiliad (changes requested):

packages/core/demo/data/line.ts
JennChao: `sparklineLoading` option is now being set within `data`. Just like how we set the loading state for other types of chart.

packages/core/demo/data/line.ts
```diff
@@ -39,6 +72,44 @@ export const lineOptions = {
...
+		enabled: false
+	},
+	tooltip: {
+		enabled: false
+	},
+	points: {
+		enabled: false
+	},
+	data: {
+		sparklineLoading: true
```
theiliad, line 109: so I'm wondering why `sparklineLoading` would be a separate entity? why can't sparkline just read the `loading flag`? sounds like this loading behaviour can be used for both line & sparkline graphs? correct @jeanservaas?

theiliad (changes requested):

packages/core/demo/data/line.ts
JennChao: The `loading` entity and `sparklineLoading` entity should be different. If we set `loading` to true, skeleton chart will be displayed. Only if we set `sparklineLoading` to true, we will be able to see loading sparkline. Unless we no longer want to show skeleton for line chart then we can just use `loading` as the flag to check whether we want to display loading sparkline or not. What do you think? @theiliad

theiliad: I'll respond in the change request

theiliad (changes requested):
@JennChao currently it seems like you're using the line graph to trigger the loading state.

This won't be a scalable solution.

We'd need the skeleton component to handle this state.

What I suggest is:
1. We create a `skeleton-line.ts` component that extends `skeleton.ts`.
2. `skeleton-line.ts` checks your grid & gradient options the same way that the line graph does, in order to determine whether you have a sparkline graph or not.
3. `skeleton-line.ts` will render either the grid loading state (regular line chart) or the sparkline loading state

This way we won't need a separate `sparklineLoading` flag. Nor would we need a `allowSparklineLoading` in the line chart...

jeanservaas (approved):
I'm fine with this from a design perspective!

JennChao:
> @JennChao currently it seems like you're using the line graph to trigger the loading state.
> 
> This won't be a scalable solution.
> 
> We'd need the skeleton component to handle this state.
> 
> What I suggest is:
> 
>     1. We create a `skeleton-line.ts` component that extends `skeleton.ts`.
> 
>     2. `skeleton-line.ts` checks your grid & gradient options the same way that the line graph does, in order to determine whether you have a sparkline graph or not.
> 
>     3. `skeleton-line.ts` will render either the grid loading state (regular line chart) or the sparkline loading state
> 
> 
> This way we won't need a separate `sparklineLoading` flag. Nor would we need a `allowSparklineLoading` in the line chart...


@theiliad 
I have an idea about leveraging the existing `Skeleton` component to create loading sparkine. The following is what I plan to do. Please take a look and let me know whether you like it or not.

There is one method called `renderSkeleton` in `Skeleton.ts` component. I plan to add the following `else if` block to render the loading sparkline UI.
圖片

And I can create a new method called `renderSparklineSkeleton` in `Skeleton.ts` to handle loading sparkline.
圖片

I modify both `drawYGrid` and `drawBackdrop` in `Skeleton.ts` so we can have the correct UI for loading sparkline.
圖片
圖片

In `charts/line.ts`, we can add the following lines of code so `Skeleton.ts` component can tell whether it should render loading sparkline or chart grids keleton by reading the value of `this.configs.skeleton`.
圖片

In `enums.ts`, we have to add `SPARKLINE = "sparkline"`.
圖片

In `charts.ts`, we can add the following entity to `LineChartOptions` interface. If the user wants to show loading sparkline instead of grids skeleton, user can simply specify `sparklineEnabled: true` in the config.
圖片

With all the modification I just mentioned, we no longer need to add any other changes to `graphs/line.ts` component.


And this is what the loading sparkline will look like.
sparkline_cc_style

theiliad:
> > @JennChao currently it seems like you're using the line graph to trigger the loading state.
> > This won't be a scalable solution.
> > We'd need the skeleton component to handle this state.
> > What I suggest is:
> > ```
> > 1. We create a `skeleton-line.ts` component that extends `skeleton.ts`.
> > 
> > 2. `skeleton-line.ts` checks your grid & gradient options the same way that the line graph does, in order to determine whether you have a sparkline graph or not.
> > 
> > 3. `skeleton-line.ts` will render either the grid loading state (regular line chart) or the sparkline loading state
> > ```
> > 
> > 
> > This way we won't need a separate `sparklineLoading` flag. Nor would we need a `allowSparklineLoading` in the line chart...
> 
> @theiliad
> I have an idea about leveraging the existing `Skeleton` component to create loading sparkine. The following is what I plan to do. Please take a look and let me know whether you like it or not.
> 
> There is one method called `renderSkeleton` in `Skeleton.ts` component. I plan to add the following `else if` block to render the loading sparkline UI.
> 圖片
> 
> And I can create a new method called `renderSparklineSkeleton` in `Skeleton.ts` to handle loading sparkline.
> 圖片
> 
> I modify both `drawYGrid` and `drawBackdrop` in `Skeleton.ts` so we can have the correct UI for loading sparkline.
> 圖片
> 圖片
> 
> In `charts/line.ts`, we can add the following lines of code so `Skeleton.ts` component can tell whether it should render loading sparkline or chart grids keleton by reading the value of `this.configs.skeleton`.
> 圖片
> 
> In `enums.ts`, we have to add `SPARKLINE = "sparkline"`.
> 圖片
> 
> In `charts.ts`, we can add the following entity to `LineChartOptions` interface. If the user wants to show loading sparkline instead of grids skeleton, user can simply specify `sparklineEnabled: true` in the config.
> 圖片
> 
> With all the modification I just mentioned, we no longer need to add any other changes to `graphs/line.ts` component.
> 
> And this is what the loading sparkline will look like.
> sparkline_cc_style

Hi, again we want to avoid any extra flags in options.

Just the fact that the grid & axes are invisible should be enough for the skeleton to not show the grid and render just the line.

I still am recommending a `skeleton-line.ts` class

packages/core/src/interfaces/charts.ts
```diff
@@ -86,14 +86,18 @@ export interface BaseChartOptions {
...
 		loading?: boolean;
 		/**
 		 * options related to pre-selected data groups
 		 * Remains empty if every legend item is active or dataset doesn't have the data groups.
 		 */
 		selectedGroups?: string[];
+		/**
+		 * used to simulate data loading in sparkline way
+		 */
+		sparklineLoading?: boolean;
```
theiliad, line 100:
```suggestion
```

theiliad: Also just confirming again that there's no reference of `sparklineLoading` in the code?

theiliad (changes requested):

packages/core/src/interfaces/charts.ts
JennChao: `sparklineLoading` has been removed.

theiliad (approved):

JennChao:
@jeanservaas @natashadecoste @zvonimirfras 
Please help to review this PR to see whether it can be merged or not. Thanks a lot!

sophiiae (approved):

natashadecoste (approved):

Issue #835, opened 2020-10-06
Add in confidence interval chart 


shaziajk:
### Issue description
Add in a confidence interval chart (screenshot below of an example confidence interval graph). When given the confidence interval range for a point be able to plot this onto a chart. This will look similar to what is already available for a line chart in terms of design but would have an included filled area around the line indicating the confidence interval 

The input data would match that of other charts but with the additional fields for the confidence interval e.g. [
  {
    "group": "Dataset 1",
    "date": "2019-01-01T00:00:00.000Z",
    "value": 10,
    "CI-Min": 5,
    "CI-Max": 10
  },.....

### Screenshot or recording
![image (5)](https://user-images.githubusercontent.com/44590418/95228680-8c792c00-07f7-11eb-8370-3d78d7dbd1d4.png)

![image](https://user-images.githubusercontent.com/44590418/95225501-c6e0ca00-07f3-11eb-999a-4cfa6ca1022b.png)

ptlm500:
Regarding rendering confidence interval for discrete data with some form of interpolation function for "smoothing", as in:

![image](https://user-images.githubusercontent.com/16822085/95229044-027d9300-07f8-11eb-8515-f9078687bba2.png)

I would propose that the same interpolation function is applied to the upper and lower bounds of the confidence interval, so the filled background displaying the confidence interval is "smoothed" in the same way.

Issue #723, opened 2020-07-27
Combo chart [area + line]


natashadecoste:
#### I am submitting a...

- [x] Feature request
- [ ] Design defect
- [ ] Source code defect
- [ ] Demo/documentation defect
- [ ] Other

## Issue description
Need a bounded area chart to display a range (actual, expected, etc) over time with a Line chart that could show actual value (or historic value etc). A regular Area + Line case is below as well. 

#### bounded area might be similar to:

![image](https://user-images.githubusercontent.com/14351335/88562805-25711700-cfff-11ea-8f07-2aaace830ac3.png)

#### area + line use case: 

![image](https://user-images.githubusercontent.com/14351335/93376376-81019900-f827-11ea-800c-82328abf1945.png)

PR #844, opened 2020-10-19
feat(confidence interval component): implement confidence interval ch…

shaziajk:
…art component & update demos

Implement a new confidence interval chart component and all the changes needed to the axis so they
are aware of the confidence interval. Update the demos to include a new one for this chart and
update the zoom bar demo to include the confidence interval.

### Updates
- Update to add in the confidence interval as a component
- Update demos to demonstrate chart 
- It follows closely with the area-chart implementation.
- Issue Number: https://github.com/carbon-design-system/carbon-charts/issues/835

Screenshot 2020-10-19 at 16 27 28
Screenshot 2020-10-19 at 16 27 37
Screenshot 2020-10-19 at 16 27 48

- [ ] Demos all features
- [ ] Documented/annotated
- [ ] Matches UI/UX specs
- [ ] Accessible


jeanservaas (commented):
Hey all, 

I think this can work but @theiliad had a great point about accessibility. Right now I know that this is not the correct palette choice for the dark theme, but we're working on fixing our palette options upstream.

I think in a situation like this though, if you're using a 30% opacity of the stroke color for the confidence interval enhancement, it will not be accessible unless you also put another stroke around it. 

I think you could do something like this to make it accessible but not make the line too obtrusive:

that's a .5 px stroke with dashed line (2px dash, 2px gap)


image

One other way to consider (@shixiedesign provided this inspiration) is to consider using handles rather than coloring the whole area. But this would be a different approach and would require some more exploration, whereas the other idea would be a relatively quick fix.

image

shaziajk:
Hey @jeanservaas @theiliad 
I've added in the additional line around the outside, I didn't feel the handles would fit in with how a traditional confidence interval generally looks.
Screenshots are included below of how it looks, on the dark colour pallet though it might possible be a tiny bit tricky to see still.

Screenshot 2020-10-22 at 12 33 00
Screenshot 2020-10-22 at 12 33 07
Screenshot 2020-10-22 at 12 33 21

shaziajk:
I also had another question, does the new graph have to be manually added into the react, angular etc repos? as currently all the changes are limited to core
thanks!

theiliad:
> I also had another question, does the new graph have to be manually added into the react, angular etc repos? as currently all the changes are limited to core
> thanks!

yes if you've added a new chart type, then it'll have to be added

shaziajk:
Thanks @theiliad this should now all be added in!

theiliad (commented):
@shaziajk I'm wondering why you made this into a chart type rather than a component that can be added through charting options?

Is there a specific reason why we'd want to keep these as just chart types? Any benefits? Downsides?

shaziajk:
@theiliad I understand what you mean! I wasn't sure of which way it would fit better, but I decided chart as it won't be able to be used with any of the other charts if it was just a component to be added through charting options other then for the line chart. So thats why I went as having at as its own chart as it didn't make sense to have it as just an option for that reason, and I didn't want complicate things if it was a separate component to be added in 

But if you guys prefer I can work on getting it changed to an option if thats a better fit?

theiliad (changes requested):

packages/core/src/components/graphs/confidence-interval.ts
```diff
@@ -0,0 +1,154 @@
...
+		const { cartesianScales, curves } = this.services;
+
+		const orientation = cartesianScales.getOrientation();
+
+		const areaGenerator = area().curve(this.services.curves.getD3Curve());
+
+		// Update the bound data on area groups
+		const groupedData = this.model.getGroupedData();
+
+		const confidence = groupedData && groupedData.length === 1;
```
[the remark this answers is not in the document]
shaziajk: Yep that makes sense! the console warning should now be in

theiliad (changes requested):
My feedback today is mainly around `patternizing` the confidence behaviour.

If this is something that'll be available through a flag and applicable to all cartesian chart types, then what are the implications.

For instance, with stacked bar charts we realized the need for `stacked data` to become a `thing` that the library understands in general, as in it was no longer just something that belonged to the stacked bar chart, but rather something that's accessible by future chart or graph types that have a `stacked` nature...

Is this something we need to look at here?

shaziajk:
> My feedback today is mainly around `patternizing` the confidence behaviour.
> 
> If this is something that'll be available through a flag and applicable to all cartesian chart types, then what are the implications.
> 
> For instance, with stacked bar charts we realized the need for `stacked data` to become a `thing` that the library understands in general, as in it was no longer just something that belonged to the stacked bar chart, but rather something that's accessible by future chart or graph types that have a `stacked` nature...
> 
> Is this something we need to look at here?

Yep that is a fair point and I agree it would be useful to do for future charts or graphs which may need a confidence behaviour. I will look into it and try and add this in

theiliad:
> > My feedback today is mainly around `patternizing` the confidence behaviour.
> > If this is something that'll be available through a flag and applicable to all cartesian chart types, then what are the implications.
> > For instance, with stacked bar charts we realized the need for `stacked data` to become a `thing` that the library understands in general, as in it was no longer just something that belonged to the stacked bar chart, but rather something that's accessible by future chart or graph types that have a `stacked` nature...
> > Is this something we need to look at here?
> 
> Yep that is a fair point and I agree it would be useful to do for future charts or graphs which may need a confidence behaviour. I will look into it and try and add this in

yes, or possibly separating some of the `confidence-only` features and checks that don't apply to other graphs. E.g. we wouldn't want axis-related checks or code to run when using a pie chart

shaziajk:
Hey @theiliad 
I've separated out the confidence-only features so its a separate thing that should now be able to be understood and used by any potential future components that also have a confidence type behaviour. This is mainly around generating the scaled points needed for the confidence. This would also mean it wouldn't run for charts that it doesn't apply to. 

Just wanted to check if im on the right track with that i've done?

packages/core/src/components/graphs/confidence-interval.ts
```diff
@@ -0,0 +1,208 @@
...
+			Events.Legend.ITEM_HOVER,
+			this.handleLegendOnHover
+		);
+
+		// Un-highlight area on legend item mouseouts
+		eventsFragment.addEventListener(
+			Events.Legend.ITEM_MOUSEOUT,
+			this.handleLegendMouseOut
+		);
+	}
```
theiliad, line 34: we probably don't need these since we're only visualizing 1 datagroup

packages/core/src/components/graphs/confidence-interval.ts
```diff
@@ -0,0 +1,208 @@
...
+		const areaGenerator = area().curve(this.services.curves.getD3Curve());
+
+		// Update the bound data on area groups
+		const groupedData = this.model.getGroupedData();
+
+		const confidence = groupedData && groupedData.length === 1;
+
+		if (!confidence) {
+			console.warn(
+				"Confidence Intervals can only be shown when having 1 single dataset"
```
theiliad, line 52:
```suggestion
				`Confidence Intervals can only be shown when having 1 single datagroup, you've supplied ${groupedData.length}`
```

theiliad: Also, if they provide multiple datagroups, what do they see?

packages/core/src/components/graphs/confidence-interval.ts
```diff
@@ -0,0 +1,208 @@
...
+			); // eslint-disable-line no-console
+		}
+
+		if (orientation === CartesianOrientations.VERTICAL) {
+			areaGenerator
+				.x((d, i) => cartesianScales.getDomainValue(d, i))
+				.y0((d, i) =>
```
theiliad, line 59: can we move these into a variable? e.g.

```js
const yAccessor = confidence ? cartesianScales.getConfidenceScaledValues : cartesianScales.getRangeValue;
```

and then calling `yAccessor(d, i)` etc...

packages/core/src/components/graphs/confidence-interval.ts
```diff
@@ -0,0 +1,208 @@
...
+						? cartesianScales.getConfidenceScaledValues(d, i)[0]
+						: cartesianScales.getRangeValue(d, i)
+				)
+				.x1((d, i) =>
+					confidence
+						? cartesianScales.getConfidenceScaledValues(d, i)[1]
+						: cartesianScales.getRangeValue(d, i)
+				)
+				.y((d, i) => cartesianScales.getDomainValue(d, i));
+			domain = this.services.cartesianScales.getMainXScale().domain();
```
theiliad, line 83:
```suggestion
			domain = cartesianScales.getMainXScale().domain();
```

packages/core/src/components/graphs/confidence-interval.ts
```diff
@@ -0,0 +1,208 @@
...
+		// Remove legend listeners
+		const eventsFragment = this.services.events;
+		eventsFragment.removeEventListener(
+			Events.Legend.ITEM_HOVER,
+			this.handleLegendOnHover
+		);
+		eventsFragment.removeEventListener(
+			Events.Legend.ITEM_MOUSEOUT,
+			this.handleLegendMouseOut
+		);
```
theiliad, line 206:
```suggestion
```

packages/core/src/components/graphs/confidence-interval.ts
theiliad, line 34:
```suggestion
```

theiliad (changes requested):
almost good to go 💯

packages/core/src/components/graphs/confidence-interval.ts
shaziajk: This change should now be in!

packages/core/src/components/graphs/confidence-interval.ts
shaziajk: This change should now be in!

packages/core/src/components/graphs/confidence-interval.ts
shaziajk: This change should now be in!

packages/core/src/components/graphs/confidence-interval.ts
shaziajk: Something similar to what you've been suggested should be in!

packages/core/src/components/graphs/confidence-interval.ts
shaziajk: The change suggested has been implemented. If multiple datagroups are provided tho you just see the lines without the confidence attached, as pictured below. Is that ok?
Screenshot 2020-11-27 at 15 44 17

packages/core/src/components/graphs/confidence-interval.ts
shaziajk: Yep makes sense and has been removed!

packages/core/src/components/graphs/confidence-interval.ts
```diff
@@ -0,0 +1,208 @@
...
+	};
+
+	handleLegendMouseOut = (event: CustomEvent) => {
+		this.parent
+			.selectAll("path.area")
+			.transition(
+				this.services.transitions.getTransition("legend-mouseout-area")
+			)
+			.attr("opacity", Configuration.area.opacity.selected);
+	};
```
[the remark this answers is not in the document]
shaziajk: This has all now been removed!!

packages/core/src/interfaces/charts.ts
```diff
@@ -316,6 +316,26 @@ export interface RadarChartOptions extends BaseChartOptions {
 	};
 }
 
+/**
+ * options specific to confidence interval charts
+ */
+export interface ConfidenceIntervalChartOptions extends ScatterChartOptions {
```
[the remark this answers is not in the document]
shaziajk: Yep makes sense for it to extend linechartoptions and should now be updated to this.

packages/core/src/components/graphs/confidence-interval.ts
theiliad: I was thinking maybe the 1st one just shows up with the confidence, but I guess this works too

packages/core/src/components/graphs/confidence-interval.ts
```diff
@@ -49,38 +33,25 @@ export class ConfidenceInterval extends Component {
...
 		if (!confidence) {
 			console.warn(
-				"Confidence Intervals can only be shown when having 1 single dataset"
+				`Confidence Intervals can only be shown when having 1 single datagroup, you've supplied ${groupedData.length}`
 			); // eslint-disable-line no-console
 		}
 
+		const upperConfidenceBound = (d, i) => confidence ? cartesianScales.getConfidenceScaledValues(d, i)[0] : cartesianScales.getRangeValue;
+		const lowerConfidenceBound = (d, i) => confidence ? cartesianScales.getConfidenceScaledValues(d, i)[1] : cartesianScales.getRangeValue;
+
```
theiliad, line 42: very nice 💯

theiliad (dismissed):

packages/core/src/services/scales-cartesian.ts
```diff
@@ -409,6 +421,23 @@ export class CartesianScales extends Service {
 			);
 		} else if (scaleType === ScaleTypes.TIME) {
 			allDataValues = displayData.map((datum) => datum[mapsTo]);
+			// If the scale has confidence
+		} else if (confidence && options.axes) {
```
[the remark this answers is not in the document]
shaziajk: So for this, I tried to make it so time scale is also used for confidence but it became a bit tricky as the current code base doesn't allow the time scale to be classed as range which is what confidence is calculated on (if that makes sense). It then was making the confidence method slightly convoluted trying to account for the use case of having confidence on the time scale. so atm confidence wouldn't apply in cases where where `scaleType === ScaleTypes.TIME`. But it is something I can work on to implement if you feel it should be in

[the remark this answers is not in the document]
theiliad: I see time series demos in your deploy previews. Seems like it's working?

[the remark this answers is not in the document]
shaziajk: yeah it works for when the `min` `max` of confidence isnt time based, if that makes sense?

packages/core/demo/data/confidence-interval.ts
```diff
@@ -0,0 +1,61 @@
...
+
+export const confidenceIntervalTimeSeriesOptions = {
+	title: "Confidence Interval (time series)",
+	legend: {
+		enabled: false
+	},
+	confidence : {
+		upperBoundMapsTo: "max",
+		lowerBoundMapsTo: "min" 
+   	},
```
cal-smith, line 17: Is there a reason this couldn't be made generic for any axis chart? It's not uncommon to see error bars on scatter plots or bar charts .... I don't entirely understand why it needs to be a whole new chart?

cal-smith (changes requested):
Requesting changes, but more of a general comment. Just want to make sure we're approaching this problem from the right angle.

I see this has been discussed previously, but we're still adding a new chart type that effectively just mimics a line chart - why not add the functionality to line directly?

packages/core/demo/data/confidence-interval.ts
shaziajk: @cal-smith I think adding it directly to line chart would work, but in terms of adding it to all axis charts Im not sure if this would work for all chart types. For example area chart is also an axis chart but further adding a confidence interval on to this would get messy. In this case error bars would be needed as an alternate to confidence interval. Which is why i think @theiliad suggested separating out the confidence behaviour to be used for future graphs.

natashadecoste: I was wondering this as well, is it not better suited to be an option for an area chart to be bounded with min/max values **or we have a carbon-charts `BoundedAreaChart`** similar to stacked area but then combine it with a `Line` chart (or multiple line charts) via `ComboChart`? 

FYI the bounded area + line issue is here:
https://github.com/carbon-design-system/carbon-charts/issues/723

shaziajk: @natashadecoste a `BoundedAreaChart` sounds good! I think that is essentially what is happening now as the confidence interval component is currently a bounded area and the line is just drawn on by calling the line chart component (in a similar way to what the area chart is currently doing). But yeah making it an additional option for the area chart to be bounded or making a `BoundedAreaChart` similar to stacked i think might be better then having it available to all axes type charts

cal-smith: Interesting ... We definitely don't have to add it to everything _now_, but as long as we have the behaviour for confidence intervals/error bars in a central spot, and we can add it as needed to _existing_ chart types we should be good. It seems like the main barrier to adding it to more axis/Cartesian charts is primarily design related, so I'd still want to make sure that we aren't coding ourselves into a corner.

@natashadecoste's suggestion seems reasonable as well if we need a way to expose this as a first pass. I'd probably lean towards making it a config, but that's totally up to You @shaziajk, @theiliad, and @natashadecoste

theiliad: So maybe then `confidence` needs to also be changed to `bounds` => { upper, lower }

theiliad: the reason we wouldn't want to make it just a config is that a bar chart for instance wouldn't really benefit from the bounds showing up behind the bars

shaziajk: ok sounds cool would it make more sense to follow the making option for an area chart to be bounded with min/max values suggested by @natashadecoste and then chance confidence to bounds => { upper, lower }? 
rather then make a whole `BoundedAreaChart` which follow a similar approach to stacked?

jeanservaas:
@shaziajk thanks so much for this work. The side stroke is no big deal, I'm happy to let that one go.

jeanservaas (approved):

theiliad (changes requested):
@shaziajk looks good, I think we're good to go except for adding some final typescript definitions for the latest configurables that you've added

theiliad:
> @shaziajk looks good, I think we're good to go except for adding some final typescript definitions for the latest configurables that you've added

https://github.com/carbon-design-system/carbon-charts/tree/master/packages/core/src/interfaces

theiliad (dismissed):
Good to go actually 💯

theiliad (approved):

cal-smith (approved):

PR #845, opened 2020-10-20
fix(core): pass all data to customHTML for scatter-stacked charts

obambrough:
Provide all data to the customHTML tooltip callback for scatter-stacked, rather than just the group
name, domain and range values.

Before:
image

After (with additional properties from data series):
image

- [ ] Demos all features
- [ ] Documented/annotated
- [ ] Matches UI/UX specs
- [ ] Accessible

### DIFF
--- a/packages/core/src/components/graphs/scatter-stacked.ts
+++ b/packages/core/src/components/graphs/scatter-stacked.ts
@@ -121,6 +121,17 @@ export class StackedScatter extends Scatter {
 				}
 			});
 		});
-		return tooltipData;
+
+		return this.model.getDisplayData().filter((datapoint) => {
+			return (
+				tooltipData.find((tooltipDatapoint) => {
+					return (
+						tooltipDatapoint[groupMapsTo] == datapoint[groupMapsTo] &&
+						tooltipDatapoint[domainIdentifier] == datapoint[domainIdentifier] &&
+						tooltipDatapoint[rangeIdentifier] == datapoint[rangeIdentifier]
+					);
+				}) !== undefined
+			);
+		});
 	}
 }


packages/core/src/components/graphs/scatter-stacked.ts
```diff
@@ -121,6 +121,17 @@ export class StackedScatter extends Scatter {
 				}
 			});
 		});
-		return tooltipData;
+
+		return this.model.getDisplayData().filter((d) => {
+			return (
+				tooltipData.find((ttd) => {
```
theiliad, line 127:
```suggestion
				tooltipData.find((tooltipDatapoint) => {
```

packages/core/src/components/graphs/scatter-stacked.ts
```diff
@@ -121,6 +121,17 @@ export class StackedScatter extends Scatter {
 				}
 			});
 		});
-		return tooltipData;
+
+		return this.model.getDisplayData().filter((d) => {
```
theiliad, line 125:
```suggestion
		return this.model.getDisplayData().filter((datapoint) => {
```

packages/core/src/components/graphs/scatter-stacked.ts
```diff
@@ -121,6 +121,17 @@ export class StackedScatter extends Scatter {
...
+			return (
+				tooltipData.find((ttd) => {
+					return (
+						ttd[groupMapsTo] == d[groupMapsTo] &&
+						ttd[domainIdentifier] == d[domainIdentifier] &&
+						ttd[rangeIdentifier] == d[rangeIdentifier]
```
theiliad, line 131: and updating the same names here. I think this'll make it a bit more readable

theiliad (changes requested):

packages/core/src/components/graphs/scatter-stacked.ts
obambrough: I've pushed the changes with the updated names

theiliad (approved):

sophiiae (approved):

natashadecoste (approved):

PR #855, opened 2020-10-30
fix(zoombar): cache zoombar data and zoombar default domain

scottdickerson:
### Updates
- list
new demo data, new high-scale testcase with 2000 data points in zoom bar
- out
- updates
zoom-bar.ts, pass the calculated zoom bar data to the domain calculator on render
zoom.ts, add the custom zoomBarData to the model if it exists and pull it from there, remove the sanitization logic as that will be done by @theiliad and that was causing the slowdown at render time

- [ ] Demos all features
- [ ] Documented/annotated
- [ ] Matches UI/UX specs
- [ ] Accessible


theiliad (commented):
Looks ok mostly, just a couple of concerns:
1. What are the demo data additions? Do we need those or are you testing using those?
2. How much have you tested this? Is there a chance that these changes will break some other demos or functionalities?

scottdickerson:
The demo data additions are just for testing

scottdickerson:
I tested all of the existing demos and they’re looking fine for zoombar

hlyang397:
@scottdickerson How can we update zoom bar data if it's necessary ?

scottdickerson:
@hlyang397 thanks I added a new method to model called setZoomBarData to address that, please take another look.  @theiliad I think this fix is ready to merge when you have a chance to review it.

packages/core/stories/utils.ts
```diff
@@ -48,3 +48,48 @@ export const addRadioButtonEventListeners = (container) => {
...
+				parseInt(e.currentTarget[0].value) || 2000;
+			console.log(recordsToGenerate);
+			chart.model.setData(
+				generateRandomData(recordsToGenerate, 100, 500)
+			);
+			console.log(`updating chart`, chart);
+			chart.update();
+		});
+	}
+};
```
theiliad, line 95: this form is a bit buggy in storybook. When you go to 1000 then 100, the zoombar will still show the incorrect data until it's been interacted with

packages/core/stories/utils.ts
```diff
@@ -48,3 +48,48 @@ export const addRadioButtonEventListeners = (container) => {
...
+	`<form id="demo-data"><label for="demo-data-name">Records to generate: </label><input type="number" id="demo-data-number" name="number" required
+	 size="5" value="100"><input type="submit"></label></form>`;
+export const addDemoDataFormListeners = (container, demo, chart) => {
+	// Add event listeners for form
+	const form = container.querySelector("form#demo-data");
+	if (form) {
+		form.addEventListener("submit", (e: any) => {
+			e.stopPropagation();
+			e.preventDefault();
+			console.log("submit event", e);
```
theiliad, line 84: do we need this?

theiliad (changes requested):
Could you please delete this demo set entirely as well
image

packages/core/stories/utils.ts
```diff
@@ -48,3 +48,48 @@ export const addRadioButtonEventListeners = (container) => {
...
+				group: "group",
+				value: Math.floor(Math.random() * (max - min + 1) + min),
+				date: new Date(now.valueOf() + (index - quantity) * 60000) // go forward a minute for every value
+			};
+		});
+};
+/**
+ * Adds a generate demo data form to the story
+ */
+export const generateDemoDataForm = () =>
```
theiliad, line 74:
```suggestion
export const generateHighScaleDemoDataForm = () =>
```

theiliad (changes requested):

packages/core/stories/utils.ts
scottdickerson: I wasn't able to reproduce that, can you try again?

theiliad:
sanitization logic is available via #856

scottdickerson:
ok I've merged your stuff in now @theiliad and I'm ready for a merge

packages/core/src/model.ts
```diff
@@ -139,9 +139,54 @@ export class ChartModel {
 			dataGroups
 		});
 
+		if (this.getZoomBarData()) {
```
theiliad, line 142: shouldn't there also be a check for whether the zoombar is enabled?

theiliad: Also, it seems like we need a new model class for axis charts. We can call it `AxisChartsModel` and it can extend `Model` & add all this stuff. There's no zoombar in a pie chart for instance so it doesn't make sense to include all this for it

theiliad (changes requested):

packages/core/src/components/axes/zoom-bar.ts
```diff
@@ -526,27 +541,31 @@ export class ZoomBar extends Component {
...
+		const zoomBarData = Tools.clone(data);
 
 		const domainIdentifier = this.services.cartesianScales.getDomainIdentifier();
 		const rangeIdentifier = this.services.cartesianScales.getRangeIdentifier();
 
 		// if min domain is extended
-		if (Number(defaultDomain[0]) < Number(data[0][domainIdentifier])) {
+		if (
+			Number(defaultDomain[0]) < Number(zoomBarData[0][domainIdentifier])
+		) {
```
theiliad, line 552: will the domain always be in numeric values? Are we concerned with date values here?

packages/core/src/model.ts
```diff
@@ -139,9 +139,54 @@ export class ChartModel {
...
+		const zoomBarNormalizedValues = allDates.map((date) => {
+			let sum = 0;
+			const datum = {};
+
+			sanitizedData.forEach((data) => {
+				if (data[domainIdentifier].getTime() === date) {
+					sum += data[rangeIdentifier];
+				}
+			});
+			datum[domainIdentifier] = new Date(date);
```
theiliad, line 177: do we still need to wrap date values into JS-dates? The sanitize PR should've addressed this and now you should be getting JS-dates already

theiliad (changes requested):

packages/core/src/model.ts
scottdickerson: unfortunately this has to live here for now I think because I need to update the zoomBarData whenever the models setData method is called (to sync it), the only way that this.getZoomBarData() returns with values if it a zoom-bar.ts has already set it up

packages/core/src/components/axes/zoom-bar.ts
scottdickerson: this one I didn't change

packages/core/src/model.ts
scottdickerson: I do, because in this case the date is actually the UTC time (see line 162)

scottdickerson:
I don't think I can make those suggested changes @theiliad but let me know if I'm incorrect, I did remove the __DEV__ stories as part of my latest push

theiliad (approved):

sophiiae:
Screen Shot 2020-11-05 at 2 31 05 PM

seems something is broken here

natashadecoste (approved):

natashadecoste (changes requested):
you get a blank in the `Angular` storybook too in addition to @sophiiae comment on the react

natashadecoste:
@scottdickerson 
it seems on the initial load there isnt any data but the model still tries to convert it to tabular data format. 
the model tries to use the `sanitize` method to change the data but i think data isnt empty [] its actually `undefined`.

scottdickerson:
thanks @natashadecoste @sophiiae I've fixed, I'm just going to hide this story from the downstream examples for now (only show in Core) since I don't have access to the storyUtils to generate the data, it's ready for re-review

theiliad (approved):

natashadecoste (approved):

PR #857, opened 2020-11-01
Render zoom bar skeleton whenever loading is true

hlyang397:
### Updates
- render zoom bar skeleton whenever loading is true
  -- needed if charts is changing from loaded to loading
- dispatch Events.ZoomDomain.CHANGE when zoom bar selection is completed
  -- useful for user to listen to Events.ZoomDomain.CHANGE instead of Events.ZoomBar.SELECTION_END

- No UI change

- [ ] Demos all features
- [ ] Documented/annotated
- [ ] Matches UI/UX specs
- [ ] Accessible

### DIFF
--- a/packages/core/src/components/axes/zoom-bar.ts
+++ b/packages/core/src/components/axes/zoom-bar.ts
@@ -110,14 +110,16 @@ export class ZoomBar extends Component {
 				.attr("x", axesLeftMargin)
 				.attr("y", 0)
 				.attr("width", width - axesLeftMargin)
-				.attr("height", "100%");
+				.attr("height", "100%")
+				.classed("zoom-bg-skeleton", isTopZoomBarLoading);
 		} else if (zoombarType === ZoomBarTypes.SLIDER_VIEW) {
 			// Draw zoombar background line
 			DOMUtils.appendOrSelect(container, "rect.zoom-slider-bg")
 				.attr("x", axesLeftMargin)
 				.attr("y", zoombarHeight / 2 - 1)
 				.attr("width", width - axesLeftMargin)
-				.attr("height", 2);
+				.attr("height", 2)
+				.classed("zoom-slider-bg-skeleton", isTopZoomBarLoading);
 		}
 
 		if (isTopZoomBarLoading) {
@@ -574,7 +576,7 @@ export class ZoomBar extends Component {
 		return zoomBarData;
 	}
 
-	renderZoomBarBaseline(container, startX, endX) {
+	renderZoomBarBaseline(container, startX, endX, skeletonClass = false) {
 		const zoombarType = Tools.getProperty(
 			this.model.getOptions(),
 			"zoomBar",
@@ -586,10 +588,9 @@ export class ZoomBar extends Component {
 			[startX, zoombarHeight],
 			[endX, zoombarHeight]
 		]);
-		DOMUtils.appendOrSelect(container, "path.zoom-bg-baseline").attr(
-			"d",
-			baselineGenerator
-		);
+		DOMUtils.appendOrSelect(container, "path.zoom-bg-baseline")
+			.attr("d", baselineGenerator)
+			.classed("zoom-bg-baseline-skeleton", skeletonClass);
 	}
 
 	renderSkeleton(container, startX, endX) {
@@ -613,8 +614,17 @@ export class ZoomBar extends Component {
 			this.getContainerSVG(),
 			this.brushSelector
 		).html(null);
+
 		// re-render baseline because no axis labels in skeleton so the baseline length needs to change
-		this.renderZoomBarBaseline(container, startX, endX);
+		const zoombarType = Tools.getProperty(
+			this.getOptions(),
+			"zoomBar",
+			AxisPositions.TOP,
+			"type"
+		);
+		if (zoombarType === ZoomBarTypes.GRAPH_VIEW) {
+			this.renderZoomBarBaseline(container, startX, endX, true);
+		}
 	}
 
 	destroy() {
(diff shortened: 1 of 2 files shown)


theiliad (dismissed):

hlyang397:
@natashadecoste @zvonimirfras Could you help to review this PR?  Thanks.

packages/core/src/components/axes/zoom-bar.ts
```diff
@@ -43,8 +48,12 @@ export class ZoomBar extends Component {
...
 		const svg = this.getContainerSVG();
 
-		const isTopZoomBarLoading = this.services.zoom.isZoomBarLoading(AxisPositions.TOP);
-		const isTopZoomBarLocked = this.services.zoom.isZoomBarLocked(AxisPositions.TOP);
+		const isTopZoomBarLoading = this.services.zoom.isZoomBarLoading(
+			AxisPositions.TOP
+		);
+		const isTopZoomBarLocked = this.services.zoom.isZoomBarLocked(
+			AxisPositions.TOP
+		);
```
zvonimirfras, line 56: Why can't these stay on their own lines?
```suggestion
		const isTopZoomBarLoading = this.services.zoom.isZoomBarLoading(AxisPositions.TOP);
		const isTopZoomBarLocked = this.services.zoom.isZoomBarLocked(AxisPositions.TOP);
```

packages/core/src/components/axes/zoom-bar.ts
```diff
@@ -317,6 +321,10 @@ export class ZoomBar extends Component {
 				zoomBarEventType = Events.ZoomBar.SELECTION_IN_PROGRESS;
 			} else if (event.type === "end") {
 				zoomBarEventType = Events.ZoomBar.SELECTION_END;
+				// only dispatch zoom domain change event for triggering api call when event type equals to end
+				this.services.events.dispatchEvent(Events.ZoomDomain.CHANGE, {
```
zvonimirfras, line 325: Should/Could `SELECTION_START` and `SELECTION_IN_PROGRESS` also be considered change events?

packages/core/src/components/axes/zoom-bar.ts
theiliad: Prettier will format these anyways. Looks like it decided to do that

packages/core/src/components/axes/zoom-bar.ts
hlyang397: @zvonimirfras 
I don't think so. `ZoomDomain.CHANGE` is used to notify users that the zoom domain change is completed.
When users start to drag the zoom bar handle, there will be one `SELECTION_START` and many `SELECTION_IN_PROGRESS` events. However the zoom domain is still changing, and not the users want at this moment.
After users stop the dragging, that means users want the zoom domain, then we could/should dispatch the `CHANGE` event to notify the zoom domain change.

If users really like to know the all details of the zoom bar handle, they can just list to `ZoomBar.SELECTION_XXXX` events.

packages/core/src/components/axes/zoom-bar.ts
hlyang397: That's right. If we like to keep them in singe line, we need to change the prettier configuration.

zvonimirfras (dismissed):

theiliad:
@jeanservaas is there a shimmer effect needed for the zoombar skeleton?

jeanservaas:
@theiliad 

yeah, my assumption is, that it would get the shimmer like the other containers, otherwise it's nearly invisible. What do you think?

theiliad:
> @theiliad
> 
> yeah, my assumption is, that it would get the shimmer like the other containers, otherwise it's nearly invisible. What do you think?

agreed, @hlyang397 are we able to add the shimmer effect here?

hlyang397:
@theiliad @jeanservaas @zvonimirfras 
Shimmer effect is added to zoom bar skeleton in latest commit https://github.com/carbon-design-system/carbon-charts/pull/857/commits/7bee806fa9515f0884e51a928b6229d670dda043.

Please help to review, thanks.

theiliad:
> @theiliad @jeanservaas @zvonimirfras
> Shimmer effect is added to zoom bar skeleton in latest commit [7bee806](https://github.com/carbon-design-system/carbon-charts/commit/7bee806fa9515f0884e51a928b6229d670dda043).
> 
> Please help to review it, thanks.

@hlyang397 I forget, could you please remind me, which demo would show what you've implemented?

hlyang397:
@theiliad  It should be something like this in your local storybook.
http://localhost:9006/?path=/story/zoom-bar-alpha--zoom-bar-skeleton

theiliad:
> @theiliad It should be something like this in your local storybook.
> http://localhost:9006/?path=/story/zoom-bar-alpha--zoom-bar-skeleton

so `data.loading` can be true and we still won't see a shimmer effect on the zoombar until `zoomBar.top.loading` is true?

hlyang397:
> so `data.loading` can be true and we still won't see a shimmer effect on the zoombar until `zoomBar.top.loading` is true?

Yes, the shimmer effect is just for skeleton.
Zoom bar shows skeleton only if `zoomBar.top.loading` is true.
Chart skeleton (data.loading) and zoom bar skeleton (zoomBar.top.loading) are in different options.
In our scenario, we need to display zoom bar while chart data is still loading.
For example, user changes zoom domain by changing handles in zoom bar. Chart needs to load new data while zoom bar stays with the same data.

It's not hard for user to sync these two options if necessary.

hlyang397:
@theiliad @jeanservaas @zvonimirfras
Could you help to review this PR? Thanks.

hlyang397:
@natashadecoste Could you help to review this PR? Thanks.

theiliad (approved):

sophiiae (approved):

PR #858, opened 2020-11-03
Set toolbar control background to transparent

hlyang397:
### Updates
- use fill:none to allow transparent toolbar control background. It fixes the problem when user changes the chart background color.

No UI change in demo.

- [ ] Demos all features
- [ ] Documented/annotated
- [ ] Matches UI/UX specs
- [ ] Accessible

### DIFF
--- a/packages/core/src/styles/components/_toolbar.scss
+++ b/packages/core/src/styles/components/_toolbar.scss
@@ -6,14 +6,14 @@ g.#{$prefix}--#{$charts-prefix}--toolbar {
 
 	.toolbar-container {
 		.toolbar-button {
-			fill: $ui-background;
+			fill: none;
 			cursor: pointer;
 			&:hover {
 				.toolbar-button-background {
 					fill: $hover-ui;
 				}
 				.toolbar-button-background--disabled {
-					fill: $ui-background;
+					fill: none;
 				}
 			}
 


theiliad (changes requested):
Why would the user change the background of the chart in any of the 4 themes? That'd probably result in a lack of compliance with the design language...?

hlyang397:
> Why would the user change the background of the chart in any of the 4 themes? That'd probably result in a lack of compliance with the design language...?
When we set the control background to transparent (fill:none), it could work no matter if user like to follow design language or not.

Our application does use the g90 theme, but we also use a little different background color for chart area.

joshkimmell:
If you don't feel comfortable allowing developers to set the background to transparent, why not give them more theme choices? In our case, the sparkline provides a visual element to our query cards, which are IDL compliant and accessible. We need the background to be transparent because of our card states, which include hover and disabled. Unless you're going to include states in the chart, transparency is the only thing that would work here.

theiliad:
> If you don't feel comfortable allowing developers to set the background to transparent, why not give them more theme choices? In our case, the sparkline provides a visual element to our query cards, which are IDL compliant and accessible. We need the background to be transparent because of our card states, which include hover and disabled. Unless you're going to include states in the chart, transparency is the only thing that would work here.

I think that we definitely need to discuss this on the guild call https://www.carbondesignsystem.com/whats-happening/meetups#data-visualization-weekly-playbackwednesdays-11am12pm-central

theiliad (approved):

hlyang397:
@natashadecoste @cal-smith
Could you help to review this PR ? Thanks.

cal-smith (approved):
Makes sense to me 👍

PR #862, opened 2020-11-04
Fix chart colors

sophiiae:


packages/core/src/components/graphs/gauge.ts
```diff
@@ -87,7 +87,7 @@ export class Gauge extends Component {
 		}
 	}
 
-	render(animate = true) {
+	render(animate: boolean = true) {
```
theiliad, line 90: the fact that `animate` is a boolean is inferred **automatically** by typescript from the fact that you're setting it to `true`

theiliad (changes requested):

theiliad (approved):

theiliad (approved):

natashadecoste (changes requested):
using a custom color for guage is weird - if it has `delta` the user is no longer able to make use of applying a status color to `delta`, the user would have to provide override colors for **both** value and delta in order to see the what they want for the value (arc). But then it doesnt apply the color to the delta chevron meaning it isnt possible for users to handle this on their own.

we need to keep the functionality options for the delta arrow (optional carbon statuses) but also allow both colors to be handled by the user.

theiliad:
are we good here for a final review @sophiiae?

sophiiae:
> are we good here for a final review @sophiiae?

yes, it's ready for final review.

theiliad (approved):

natashadecoste (approved):

PR #863, opened 2020-11-04
feat: provide the option to let user set sparkline tick value

JennChao:
### Updates
- list
- out
- updates

tick value set to 16
圖片

tick value set to 100
圖片

- [ ] Demos all features
- [ ] Documented/annotated
- [ ] Matches UI/UX specs
- [ ] Accessible


JennChao:
@theiliad 
This PR is created for letting the user set the tick value as wanted.

theiliad (commented):
this doesn't add up? If your chart is loading, that means that you don't yet have access to the data that the chart is going to show. At that point, how will you know the value you'd need for your ticks?

JennChao:
> this doesn't add up? If your chart is loading, that means that you don't yet have access to the data that the chart is going to show. At that point, how will you know the value you'd need for your ticks?

That's true that user doesn't know the value while data is loading. However, tick value in here is for deciding where we are going to render horizontal loading line. It is for UI display only. Tick value in here doesn't have meaning related to value of data. 

With tick value set to `100`, the loading sparkline chart is too tall to fit in query card in DE. Without adjusting height of query card, we can't see the loading sparkline at all...
圖片


With tick value set to `16`, the loading sparkline chart can fit in query card in DE just right.
圖片

I am thinking that we should provide users a flexible option to adjust tick value for UI display only so the chart can fit in their other component, for example query card in DE.

JennChao:
@theiliad @natashadecoste @zvonimirfras 
Can any one help to review this PR? Thank you very much!

packages/core/demo/data/line.ts
```diff
@@ -52,7 +52,8 @@ export const sparklineLoadingOptions = {
 			visible: false,
 			mapsTo: "value",
 			title: "Conversion rate",
-			scaleType: "linear"
+			scaleType: "linear",
+			ticks: {values: [16]}
```
natashadecoste, line 56: I think that using the configuration for `values` is confusing here. That configuration was to provide values within the domain that would be visually rendered to the chart, and the sparkline does the opposite - rendering no text ticks. I know its called `ticksValues` **within** sparkline rendering code but it feels confusing to use it in the configurations when it doesn't work the same way as it does for the charts that _do render axis tick text_

packages/core/src/components/graphs/skeleton-lines.ts
```diff
@@ -64,30 +64,37 @@ export class SkeletonLines extends Skeleton {
...
 		} else if (isDataLoading && isSparkline) {
-			this.renderSparklineSkeleton(isDataLoading);
+			const tickValue = Tools.getProperty(
+				this.model.getOptions(),
+				"axes",
+				"left",
+				"ticks",
+				"values"
+			);
+			this.renderSparklineSkeleton(isDataLoading, tickValue);
```
natashadecoste, line 74: when you remove the 
```
ticks: {
    values: [16] 
}
```
you get an error because this passes in null which doesnt get overridden with the `tickValue: number[] = [100]`

this or something more elegant 
```
this.renderSparklineSkeleton(isDataLoading, tickValue ? tickValue : [100]);
```

makes sure theres a value passed in, otherwise `renderSparklineSkeleton` will also pass a null to `drawSparkline`

natashadecoste (changes requested):

packages/core/demo/data/line.ts
JennChao: @natashadecoste 
Thanks for the review comment.

I am not quite sure about what you refer as _render axis tick text_?  `ticks: {values: [16]}` in here renders a horizontal line instead of _axis tick text_. 

And also, according to @theiliad 's previous suggestion, `loading Sparkline` is one kind of `Skeleton` so we are reusing methods from `Skeleton` to make a `loading Sparkline`.
https://github.com/carbon-design-system/carbon-charts/pull/837#pullrequestreview-513799536

Hope this could explain why I use` ticksValues` to render a horizontal loading sparkline.

PR #867, opened 2020-11-06
feat(react): port React chart wrappers to TypeScript

metonym:
### Updates

- Port React chart wrappers to TypeScript (`.jsx` --> `.tsx`)
- Install TypeScript and React type definitions as development dependencies
- Update the build script (`react/build.sh`) to use TypeScript to build library in CJS format and emit Type declarations
- Update TypeScript entry in package.json to be `"./index.d.ts"`
- Update `data` type in core Chart interfaces to reflect the new tabular data format (`Record<string, any>[]`)
- Update React storybook configuration to use transpiled TS when running in development mode or building for production

### Notes

- Some fields in chart `options` will require the consumer to use the ENUM type (e.g. [ScaleTypes](https://github.com/metonym/carbon-charts/blob/e91808cc7432838826484bed763475034d586287/packages/react/carbon__charts-react-tests.tsx#L34)) from the core `@carbon/charts` package

- [ ] Demos all features
- [ ] Documented/annotated
- [ ] Matches UI/UX specs
- [ ] Accessible


cal-smith (commented):
It may be easier to maintain if we just switch the react codebase over to typescript ... That way all the type declarations can be co-located with the code, and we won't have to worry about syncing them. I'm not sure what we're using for builds right now, but if it's babel we should be able to enable Typescript parsing there and just use tsc to generate the `.d.ts`

theiliad:
> It may be easier to maintain if we just switch the react codebase over to typescript ... That way all the type declarations can be co-located with the code, and we won't have to worry about syncing them. I'm not sure what we're using for builds right now, but if it's babel we should be able to enable Typescript parsing there and just use tsc to generate the `.d.ts`

That makes sense, any comments @metonym?

metonym:
@theiliad I concur. Porting to TypeScript would be easier to maintain and less error prone.

packages/react/src/line-chart.tsx
```diff
@@ -0,0 +1,36 @@
...
+import { LineChart as LC } from "@carbon/charts";
+import {
+  ChartData,
+  LineChartOptions,
+} from "@carbon/charts/interfaces";
+import BaseChart, { TabularData } from "./base-chart";
+
+type LineChartData =
+  | TabularData
+  | ChartData
```
theiliad, line 11: we actually need to remove this type or update it to match the tabular format. Not sure why it's still in our types

theiliad: all mentions of `datasets` & `labels` in the charting data should also disappear from the react types or tests

theiliad: https://carbon-design-system.github.io/carbon-charts/?path=/story/tutorials--tabular-data-format

metonym: Added a `ChartTabularData` type to the core model interfaces, which is pretty loose: `Record<string, any>[]`  

Haven't yet removed mentions of `datasets/labels` etc..

packages/react/package.json
```diff
@@ -76,6 +79,7 @@
 		"sass": "1.25.0",
 		"sass-loader": "8.0.0",
 		"style-loader": "0.19.0",
+		"typescript": "^4.1.2",
```
theiliad, line 82:
```suggestion
		"typescript": "4.1.2",
```

packages/react/package.json
```diff
@@ -56,6 +57,8 @@
 		"@storybook/addon-knobs": "5.3.12",
 		"@storybook/addon-options": "5.3.12",
 		"@storybook/react": "5.3.12",
+		"@types/react": "^17.0.0",
+		"@types/react-dom": "^17.0.0",
```
theiliad, line 61:
```suggestion
		"@types/react-dom": "17.0.0",
```

packages/react/package.json
```diff
@@ -56,6 +57,8 @@
 		"@storybook/addon-knobs": "5.3.12",
 		"@storybook/addon-options": "5.3.12",
 		"@storybook/react": "5.3.12",
+		"@types/react": "^17.0.0",
```
theiliad, line 60:
```suggestion
		"@types/react": "17.0.0",
```

packages/react/carbon__charts-react-tests.tsx
```diff
@@ -0,0 +1,55 @@
...
+              "Dataset 3": "#725699",
+              "Dataset 4": "#ccc",
+            },
+          },
+          height: "400px",
+        }}
+      />
+    );
+  }
+}
```
theiliad, line 55:
```suggestion
}

```

packages/react/carbon__charts-react-tests.tsx
```diff
@@ -0,0 +1,55 @@
...
+import { LineChart } from "./dist";
+
+class App extends React.Component {
+  ref: React.RefObject<LineChart> = React.createRef();
+
+  state = {
+    data: {
+      labels: [''],
+      datasets: []
+    }
```
theiliad, line 12: this is the older data format that we deprecated many versions ago https://carbon-design-system.github.io/carbon-charts/?path=/story/tutorials--tabular-data-format

packages/react/src/base-chart.tsx
```diff
@@ -0,0 +1,49 @@
...
+
+  shouldComponentUpdate(nextProps: Props<Options, Data>) {
+    return (
+      this.props.data !== nextProps.data ||
+      this.props.options !== nextProps.options
+    );
+  }
+
+  componentDidUpdate() {
+    // @ts-ignore
```
theiliad, line 44: curios as to why these are needed? do we need to fix something in the `setData` method? add some typing maybe?

theiliad: not necessarily requesting a change here, more so looking for the problem

theiliad (changes requested):

packages/react/build.sh
```diff
@@ -9,12 +9,12 @@ echo "linking local packages (non-lerna)"
...
+echo "compiling ts to js"
+tsc
 
 echo "bundling..."
 rollup -c
 
 echo "copying metadata"
 cp *.md dist/
-cp package.json dist/
+cp package.json dist/
```
theiliad, line 20:
```suggestion
cp package.json dist/

```

theiliad (approved):

theiliad (changes requested):
meant to request the change

packages/react/carbon__charts-react-tests.tsx
theiliad: are you able to update this to the tabular format?

metonym: Updated in https://github.com/carbon-design-system/carbon-charts/pull/867/commits/e9a53f4336a7af42a6b9caf555185df5624fc4de#diff-02ba3e7463b36e80fe42c20ff1e8d59c9575c7dbbd322bb6ce24630a2e240272

packages/react/src/base-chart.tsx
metonym: Fixed in https://github.com/carbon-design-system/carbon-charts/pull/867/commits/e9a53f4336a7af42a6b9caf555185df5624fc4de#diff-827cd8b812b357745e517c845fa7d5ca9f9d0cb61fdb0e50d285f809ff9d631b

theiliad: Nice 🎉

theiliad (changes requested):
I'm good to merge, except that the react storybook is fully breaking. Are you able to fix @metonym? 🙂

theiliad (approved):
💯

cal-smith (approved):
Sweet 🎉 Nice job getting this switched over @metonym

PR #877, opened 2020-11-09
feat(core): add alignment options for gauge charts

### LINKED ISSUE #870: No alignment option in GaugeChart
#### I am submitting a...

- [x] Feature request
- [ ] Design defect
- [ ] Source code defect
- [x] Demo/documentation defect
- [ ] Other

#### charts version:

v0.40.13

### Issue description

There is no option for `alignment` of [gauge chart](https://carbon-design-system.github.io/carbon-charts/documentation/interfaces/_interfaces_charts_.gaugechartoptions.html#gauge). I'm expecting the following code to center align the gauge in the same way as we can do this with a [donut chart](https://carbon-design-system.github.io/carbon-charts/documentation/interfaces/_interfaces_charts_.donutchartoptions.html#donut)

```javascript
<GaugeChart
    data={data}
    options={{
        "title": "",
        "height": "250px",
        "gauge": {
            "type": "full",
            "alignment": "center"
        },
        "color": {
            "scale": {
                "value": "#891EE8"
            }
        }
    }
}>
```
Checkout [this sandbox](https://codesandbox.io/s/carbon-react-gauge-v04013-not-aligned-82d9t?file=/src/index.js) which compares the `alignment` property of donut and gauge chart component

gauge

From the [donut chart source code](https://github.com/carbon-design-system/carbon-charts/blob/master/packages/core/src/components/graphs/donut.ts#L10) it seems it extends `Pie`; this is why it works for donut and not for [gauge chart](https://github.com/carbon-design-system/carbon-charts/blob/master/packages/core/src/components/graphs/gauge.ts#L21) which doesn't extend `Pie`. 

Also in the [carbon design chart presentation page](https://www.carbondesignsystem.com/data-visualization/basic-charts/) there is a demo for a lot of component with a "centered" version for each one and there is no "centered" example for the gauge component thus creating this feature request

In addition to this feature request, do you see an alternate way / workaround to center the gauge while keeping responsivity ?

theiliad:
`center | left | right` alignment for gauge charts

Closes #870

### DIFF
--- a/packages/core/src/components/graphs/gauge.ts
+++ b/packages/core/src/components/graphs/gauge.ts
@@ -6,7 +6,8 @@ import {
 	Events,
 	GaugeTypes,
 	ArrowDirections,
-	ColorClassNameTypes
+	ColorClassNameTypes,
+	Alignments
 } from "../../interfaces";
 import { Tools } from "../../tools";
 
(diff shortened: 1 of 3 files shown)


cal-smith (commented):
I'm having a hard time visualizing what this actually does ... Maybe it would make sense to have a story for it?

cal-smith:
Or I suppose I could play with the config ... Interesting 🤔

cal-smith:
This seems like a bit of an odd fix for the situation where the graph frame is taking up too much space ... I would expect this to be a lot easier to deal with in CSS ... 

I'll give this an approve, since it basically seems to just bring it inline with Donut/Pie, but I'm not sure it's the _best_ solution for Gauge alignment.

cal-smith (approved):

natashadecoste (approved):

PR #879, opened 2020-11-10
fix(core): make color map a global scss variable

### LINKED ISSUE #866: Colors are missing when importing charts/styles/styles.scss
- [ ] Feature request
- [ ] Design defect
- [x] Source code defect
- [ ] Other

#### charts version:

`0.40.11` Vanilla

### Issue description
Colors are missing when importing `charts/styles/styles.scss` SCSS file directly.

Using `node-sass@4.14.1`.

### Screenshot or recording

image

https://github.com/carbon-design-system/carbon-charts/blob/27e1725e6989796a3066ba73321ff77755b5187e/packages/core/src/styles/colors.scss#L275-L280

The above block should use the `!global` flag to update the global scope `$color-map`.

https://github.com/carbon-design-system/carbon-charts/blob/27e1725e6989796a3066ba73321ff77755b5187e/packages/core/src/styles/colors.scss#L250

```
$color-map: map-merge(
    $color-map,
    (
        "#{$name}-#{$category}": $value
    )
) !global;
```

Example: https://codesandbox.io/s/carbon-chart-missing-colors-2jgh2

theiliad:
Closes #866

### DIFF
--- a/packages/core/src/styles/colors.scss
+++ b/packages/core/src/styles/colors.scss
@@ -277,7 +277,7 @@ $color-map: ();
 					(
 						"#{$name}-#{$category}": $value
 					)
-				);
+				) !global;
 			}
 		}
 	} @else {


packages/core/src/styles/colors.scss
```diff
@@ -277,7 +277,7 @@ $color-map: ();
 					(
 						"#{$name}-#{$category}": $value
 					)
-				);
+				) !global;
```
sophiiae, line 280: this makes the `color-map` a local variable, isn't it?
I think you may need to make `color-map` both global and local. 
when you make it only a local variable, our demo brokes.

theiliad: @sophiiae is there an approach that'll make both work? I haven't looked into it much, are you able to pick this up from here?

sophiiae: > @sophiiae is there an approach that'll make both work? I haven't looked into it much, are you able to pick this up from here?

As I know so far, one global and one local `color-map` can make both work. 
But I'm still trying to find a better solution for this.

theiliad: Are you able to push what you're talking about into a PR? Not sure why demos are breaking, they're still just importing our SCSS styles...

sophiiae: sure, I can make a PR for it.

theiliad: Thanks

sophiiae:
#880 fixes this issue

PR #893, opened 2020-11-19
feat(core): smoother transitions in grouped and stacked bar charts

theiliad:
See demos. Bars now morph rather than fading away and fading back in


natashadecoste (approved):
looks good 👍

sophiiae (approved):

PR #906, opened 2020-12-09
fix(combo chart): cleanup demos and bugs

natashadecoste:
### Updates
todo:
- [x] axis on dual axis discrete bubble
- [x] color on area charts
- [x] custom domains
- [x] overriding the title orientation on vertical axes 

#### default direction
Default orientations:
left axis: left 
right axis: right 
image

#### overriding both left and right axes

```
axes: {
		left: {
			title: "left",
			stacked: true,
			mapsTo: 'value',
			titleOrientation: "right" // here
		},
		bottom: {
			scaleType: 'time',
			mapsTo: 'date',
		},
		right: {
			title: "right",
			scaleType: 'linear',
			mapsTo: 'temp',
			correspondingDatasets: ['Temperature'],
			titleOrientation: "left" // here
		},
	},
```
image

- [ ] Demos all features
- [ ] Documented/annotated
- [ ] Matches UI/UX specs
- [ ] Accessible


Donisius (dismissed):

sophiiae (dismissed):

packages/core/src/components/axes/axis.ts
```diff
@@ -395,10 +395,10 @@ export class Axis extends Component {
...
 				case AxisPositions.RIGHT:
 					axisTitleRef
-						.attr("transform", "rotate(90)")
-						.attr("y", -width)
-						.attr("x", scale.range()[0] / 2)
-						.attr("dy", "1em")
+						.attr("transform", "rotate(-90)")
+						.attr("y", width)
+						.attr("x", -(scale.range()[0] / 2))
+						.attr("dy", "-0.5em")
```
theiliad, line 401: does this break any other charts?

packages/core/src/model.ts
```diff
@@ -80,11 +80,11 @@ export class ChartModel {
 						);
 					} else {
 						const [start, end] = axesOptions[axis].domain;
-
-						// Filter out data outside domain
+						// Filter out data outside domain if that datapoint is using that axis (has mapsTo property)
 						allData = allData.filter(
 							(datum) =>
-								datum[mapsTo] >= start && datum[mapsTo] <= end
+								!(mapsTo in datum) ||
```
theiliad, line 86: what's this do?

theiliad (changes requested):

packages/core/src/components/axes/axis.ts
natashadecoste: it affects but doesnt break the other axis charts, but im now wondering if this is ideal:
image

The conversation to have the vertical titles oriented in same direction (and that being to align with the left axis title orientation) didnt go as far as to cover if they should still be facing the other direction when there is just a right axis. I don't see why this (above) is illogical but from the globalization standpoint, maybe we would want all labels facing the other direction in some cases? Would we want to provide a configuration to override the title orientation on each axis ?? 

this way a user would get the chart above **but they can modify it to get this:**
image

but again, not sure about the criticality of the * direction in which the titles face** so long as its consistent

packages/core/src/model.ts
natashadecoste: your model might contain data for different axes (ex: dual axes line chart)
image

when the function going through **all** the axes and "discarding"/filtering out the points that are outside of the custom domain of each axis, this check makes sure that it isnt discarding datapoints that are irrelevant to the current axis.

packages/core/src/components/axes/axis.ts
theiliad: cc @jeanservaas

packages/core/src/components/axes/axis.ts
```diff
@@ -401,12 +415,20 @@ export class Axis extends Component {
 						.style('text-anchor', 'middle');
 					break;
 				case AxisPositions.RIGHT:
-					axisTitleRef
-						.attr('transform', 'rotate(90)')
-						.attr('y', -width)
-						.attr('x', scale.range()[0] / 2)
-						.attr('dy', '1em')
-						.style('text-anchor', 'middle');
+					if (titleOrientation === AxisPositions.LEFT) {
```
theiliad, line 418:
```suggestion
					if (titleOrientation === AxisTitleOrientations.LEFT) {
```

theiliad (approved):

theiliad (changes requested):
the feedback above

theiliad (approved):

sophiiae (approved):

PR #911, opened 2020-12-15
fix(core): use new object to avoid always true in object comparison

hlyang397:
### Updates
- when options.zoomBar.top.initialZoomDomain is updated again, and it's different from the old initialZoomDomain (keep in model, we should update zoomDomain based new initialZoomDomain.
- before this fix, the comparison is based on object reference, so condition is always false.

No change in UI demos.

- [ ] Demos all features
- [ ] Documented/annotated
- [ ] Matches UI/UX specs
- [ ] Accessible


theiliad (commented):
Hi @hlyang397 👋🏻 , the original code doesn't seem like it would break, could you please specify where this is breaking and how?

hlyang397:
> Hi @hlyang397 👋🏻 , the original code doesn't seem like it would break, could you please specify where this is breaking and how?

Good question, it's a little tricky here.
Let me explain in current code:
1. line 150, initialZoomDomain is referred to initialZoomDomain in `model` ( let's call it objA)
2. line 152,  newInitialZoomDomain is referred to the initialZoomDomain in` configuration` ( let's call it objB)
3. line 171, these two objects (objA != objB) are different, so 
4. line initialZoomDomain in model is also referred to newInitialZoomDomain which is objB.
It's correct so far, however, both initialZoomDomain in `model` and initialZoomDomain in `configuration` are **referred to the same object (objB)**

So when initialZoomDomain in `configuration` (objB) value is changed by user, but
1. line 150, initialZoomDomain is referred to initialZoomDomain in `model` (objB)
2. line 152,  newInitialZoomDomain is referred to the initialZoomDomain in` configuration` (objB)
3.  line 171, these two object references are the same, so no change made. (incorrect)

That's why the code fix is about assigning initialZoomDomain in model to a new object instead of an existing object (objB in this case)

Hope these lines could help you to clarify the situation.

- BTW, we need to change initialZoomDomain in configuration based on user events in our project.

packages/core/src/components/axes/zoom-bar.ts
```diff
@@ -167,21 +167,27 @@ export class ZoomBar extends Component {
...
 				)
 			) {
 				this.model.set(
 					{
-						initialZoomDomain: newInitialZoomDomain,
+						// use a new object instead of newInitialZoomDomain
+						initialZoomDomain: [
+							new Date(newInitialZoomDomain[0].getTime()),
+							new Date(newInitialZoomDomain[1].getTime())
+						],
```
theiliad, line 185: so the changes to the if-statement above makes sense, but what are we achieving here? if we really need a new object should we use `Tools.merge({}, initialZoomDomain)`?

packages/core/src/components/axes/zoom-bar.ts
```diff
@@ -167,21 +167,27 @@ export class ZoomBar extends Component {
...
 						zoomDomain: newInitialZoomDomain
-							? newInitialZoomDomain
+							? [
+									new Date(newInitialZoomDomain[0].getTime()),
+									new Date(newInitialZoomDomain[1].getTime())
+							  ]
```
theiliad, line 190: so the changes to the if-statement above makes sense, but what are we achieving here? if we really need a new object should we use `Tools.merge({}, newInitialZoomDomain)`?

theiliad (changes requested):

packages/core/src/components/axes/zoom-bar.ts
hlyang397: `Tools.merge` makes sense to me. Updated in https://github.com/carbon-design-system/carbon-charts/pull/911/commits/9c959c413b7a7df918609ecca286fe9f1069d004.

packages/core/src/components/axes/zoom-bar.ts
hlyang397: `Tools.merge` makes sense to me. Updated in https://github.com/carbon-design-system/carbon-charts/pull/911/commits/9c959c413b7a7df918609ecca286fe9f1069d004.

hlyang397: If there is a new initialZoomDomain set by user, we should update zoomDomain according to the new initialZoomDomain.

packages/core/src/components/axes/zoom-bar.ts
```diff
@@ -172,21 +172,24 @@ export class ZoomBar extends Component {
...
 					{
-						initialZoomDomain: newInitialZoomDomain,
+						// use a new object instead of newInitialZoomDomain
+						initialZoomDomain: Tools.merge(
+							[],
+							newInitialZoomDomain
+						),
 						zoomDomain: newInitialZoomDomain
-							? newInitialZoomDomain
+							? Tools.merge([], newInitialZoomDomain)
```
theiliad, line 192: sorry this might be my mistake but does lodash's merge support array merges?

theiliad: could you please verify that this works?

hlyang397: Yes, it works as expected.

console.log("newInitialZoomDomain");
console.log(newInitialZoomDomain);
console.log(Tools.merge([], newInitialZoomDomain));
				
image

natashadecoste (approved):

theiliad (approved):

packages/core/src/components/axes/zoom-bar.ts
theiliad: 👍🏻

PR #918, opened 2021-01-04
feat(core): add custom chart - standalone components renders

### LINKED ISSUE #903: Request: Standalone Legend component
#### I am submitting a...

- [x] Feature request
- [ ] Design defect
- [ ] Source code defect
- [ ] Other

### Issue description
There are times when we need to build visualizations that are unique to our offering and do not exist in the Carbon Charts ecosystem. For example, a global map visualizing network connections.

Even though the visualization might be custom for the application, it would be great to follow the conventions provided by these charting libraries. One of those conventions is the Legend and its interactions. 

#### Current behavior
As a producer of custom visualizations, I cannot leverage any of the underlying conventions offered by Carbon Charts without duplicating code that already exists in this project.

#### Expected behavior
As a producer of custom visualizations, I want a way to leverage the Legend as a standalone component to use in my visualization. Having access to this building block, I can ensure that the charts that I produce adhere to and follow the same UX principles as all other charts.

theiliad:
👋 

This is a thought/proposal...
Closes #903 

I've added a `Standalone components` group to the **vanilla** storybook demos. It is rendering a legend component as a standalone.

image
https://deploy-preview-918--carbon-charts-core.netlify.app/?path=/story/standalone-components--demo

This is the quickest method that I could think of... However, it does have a few downsides of its own

To name a few downsides:
1. This works as if you're initializing an entire chart with just a legend visible (you'd need extra options added, e.g. rather than saying I need a legend rendered with `these` items, you're rather providing some data, providing a `groupMapsTo` field and getting the legend to render the correct datagroups...)
2. The above makes it easier to combine components through the `LayoutComponent`, however may introduce many breaking changes as we change the library

### DIFF
--- a/packages/core/stories/custom-charts.stories.ts
+++ b/packages/core/stories/custom-charts.stories.ts
@@ -1,0 +1,69 @@
+import { storiesOf } from "@storybook/html";
+import { withKnobs, object } from "@storybook/addon-knobs";
+
+import { storybookDemoGroups } from "../demo/data";
+import * as ChartComponents from "../src/charts";
+import * as storyUtils from "./utils";
+
+import "../demo/styles.scss";
+import { Legend } from "../src/components";
+
+const groupStories = storiesOf("Standalone components", module).addDecorator(
+	withKnobs
+);
+
+// Loop through the demos for the group
+groupStories.add("Demo", () => {
+	// container creation
+	const container = document.createElement("div");
+	container.setAttribute("class", "container theme--g100");
+
+	container.innerHTML = `
+<h3>
+<b class="component">Component</b>
+<span class="bx--tag bx--tag--green component-name">CustomChart</span>
+</h3>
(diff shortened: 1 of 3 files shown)


packages/core/stories/custom-charts.stories.ts
```diff
@@ -0,0 +1,69 @@
...
+					groupMapsTo: "group"
+				},
+				components: [
+					{
+						component: Legend,
+						configs: {
+							one: true
+						}
+					}
+				]
```
cal-smith, line 60: This is a neat concept 🤔 I gather the `CustomChart` is needed to setup the base layout stuff and data handling, but other than that, this is a pretty tidy way of exposing all the internals.

Definitly agree with the concern around breaking changes ... Though, once the foundational components are solidified it should be easier to justify a v1 and stabilize a bit.

Issue #917, opened 2020-12-31
Donut / Pie chart: Hide labels? Hide labels based on % of the pie/donut filled?


tony:
Code sandbox:

- Donut: https://codesandbox.io/s/carbon-components-hide-labels-donut-0mngj?file=/src/index.js
- Pie: https://codesandbox.io/s/carbon-components-hide-labels-pie-9o3h6?file=/src/index.js

#### I am submitting a...

- [x] Feature request
- [ ] Design defect
- [ ] Source code defect
- [ ] Demo/documentation defect
- [ ] Other

#### Current behavior
![image](https://user-images.githubusercontent.com/26336/103430484-2bf55300-4b8a-11eb-87fe-15db6640c1d6.png)
![image](https://user-images.githubusercontent.com/26336/103430525-a920c800-4b8a-11eb-961d-3e464c14adb2.png)

#### Attempting to hide:

```
      pie: {
        labels: {
          formatter: () => void 0
        }
      },
```

`PieChart` with `pie.legends` zero'd out: https://codesandbox.io/s/carbon-components-hide-labels-pie-void-7yxzs?file=/src/index.js

![image](https://user-images.githubusercontent.com/26336/103430540-d3728580-4b8a-11eb-86c9-a838e8bf925c.png)

`DonutChart` With `.pie.legends` set to `void`: https://codesandbox.io/s/carbon-components-hide-labels-donut-void-j2tc6

![image](https://user-images.githubusercontent.com/26336/103430513-8393be80-4b8a-11eb-96a7-58183ba309a0.png)

#### Expected behavior

Would hope for some way to hide labels outright

![image](https://user-images.githubusercontent.com/26336/103430556-fdc44300-4b8a-11eb-9ecc-bbd48e7065a3.png)

Or a way to hide conditionally based on % of the graph/circle filled

tony:
Edited above:

I think this needs design advice since this is how to handle the layout of data when there's not enough room for it.

natashadecoste:
@jeanservaas 
I can add a configuration to enable/disable the labels (and then not draw any callouts either). 
Is there more design explorations that need to be done for conditional rendering of callouts? There are some problems with what to do with slices that are too small for labels to not collide

jeanservaas:
@natashadecoste Sorry, I know this issue has been kicking around for a while. 

I would definitely add a config to enable/disable the labels in the meantime.

[...]

PR #927, opened 2021-01-11
pie/donut: disable labels (and callouts)

natashadecoste:
### Updates
- there is no respacing of the pie/donut so that the labels can be toggled on/off in different UIs/dashboards

labels off
image
labels on
image

### configs
```
pie: {
    labels: {
        enabled: false | true (defaults to true)
    }
}
```

related to #917 

- [ ] Demos all features
- [ ] Documented/annotated
- [ ] Matches UI/UX specs
- [ ] Accessible


theiliad (changes requested):
Does the pie not translate anymore and not accomodate for rendered callouts? and is @jeanservaas ok with this?

natashadecoste:
> Does the pie not translate anymore and not accomodate for rendered callouts? and is @jeanservaas ok with this?

 

> Does the pie not translate anymore and not accomodate for rendered callouts? and is @jeanservaas ok with this?

```		
if (calloutData.length > 0) {
	pieTranslateY += Configuration.pie.yOffsetCallout;
}
```

if **we do add** the padding when the labels are disabled - and then the labels are toggled back on enabled: the pie might move if you have **no calloutData** (slices are all a bigger size)

if **we don't add** padding when labels are disabled - then they are toggled back on enabled: the pie might still move around if there **is calloutData** but wont move if there isnt

the only way to have the pie stay in the same place would be if we wanted to add the callout padding regardless of whether there is callouts being rendered. this also would make sure when the legend is being used to toggle on/off datagroups that there isnt movement of the pie graph

theiliad:
> > Does the pie not translate anymore and not accomodate for rendered callouts? and is @jeanservaas ok with this?
> 
> ```
> if (calloutData.length > 0) {
> 	pieTranslateY += Configuration.pie.yOffsetCallout;
> }
> ```
> 
> if **we do add** the padding when the labels are disabled - and then the labels are toggled back on enabled: the pie might move if you have **no calloutData** (slices are all a bigger size)
> 
> if **we don't add** padding when labels are disabled - then they are toggled back on enabled: the pie might still move around if there **is calloutData** but wont move if there isnt
> 
> the only way to have the pie stay in the same place would be if we wanted to add the callout padding regardless of whether there is callouts being rendered. this also would make sure when the legend is being used to toggle on/off datagroups that there isnt movement of the pie graph

if you disable then enable callouts I'd assume that the pie moving wouldn't bee too much of an issue

theiliad:
has @jeanservaas looked at this?

tony:
This looks really good!

Any chance that disabling labels/callouts would also re-orient the box so the rest of the space is used? (If that's extra additional effort - I think it could be - I can also make a separate issue to address whether that's possible)

sophiiae (approved):

theiliad (dismissed):

theiliad:
@natashadecoste has Jeannie seen this? good to merge?

natashadecoste:
> This looks really good!
> 
> Any chance that disabling labels/callouts would also re-orient the box so the rest of the space is used? (If that's extra additional effort - I think it could be - I can also make a separate issue to address whether that's possible)

updated it 👍

tony:
@natashadecoste 

As for the extra spacing left over when stuff is hidden, will that be handled by this PR?  (I can also make a separate issue for that, if it's easier to look at after). What do you think?

Some context:

image

As a workaround until then, I manually move the container in itself:

``` scss
.chartRow--donut {
  margin-left: -29px;
  margin-top: -25px;
  z-index: -10;
}
```

https://cv-react-v2.git-pull.com/ with `^0.41.28`

image

natashadecoste:
@tony 
posted below you can see that without the labels, the pie expands and is slightly larger. it will never go right under the title because there is space there **intentionally** to allow the title to breathe. 

#### with labels
image

#### without 
image

--- 
you can see here that the pie is taking all of its space.
image


The better workaround for you would be to not use a title (implement the title separately) and then the donut gets all the space
image

tony:
@natashadecoste Thank you for the explanation!  This has been super helpful!

And to clarify, are the screenshots/explanations the present behavior in the stable release, or are we referring to new adjustments in this PR?

> implement the title separately

That's interesting to me - could you steer me to where in the documentation I can find more about this, or potentially [fork my sandbox](https://codesandbox.io/s/carbon-components-hide-labels-donut-0mngj?file=/src/index.js) to show me what that'd look like?

natashadecoste:
> And to clarify, are the screenshots/explanations the present behavior in the stable release, or are we referring to new adjustments in this PR?

Its in this PR

for the title,if you just remove the `title` option you will get the chart without the title & spacer rendered. 
Then you can implement the title separately in your app (in whichever framework) 
Specs for the titles in carbon charts are:
image

tony:
Thank you once again @natashadecoste!

sophiiae (approved):

theiliad (approved):

PR #933, opened 2021-01-13
fix(@carbon/charts): fix issues with the filled flag being undefined in bubble charts

theiliad:
image


natashadecoste (approved):
looks good

sophiiae (approved):

PR #942, opened 2021-01-28
feat(gauge): added hideSymbol property to gauge to prevent display of…

### LINKED ISSUE #941: Gauge chart should have option to hide percent symbol
#### I am submitting a...

- [x ] Feature request
- [ ] Design defect
- [ ] Source code defect
- [ ] Other

### Issue description
We are using the gauge semi chart to signify risk on a scale of 1-150. The gauge only seems to allow 1-100 and always displays the percent symbol. It is possible to override the scale via the numberFormatter option, but it is not possible to disable the symbol from displaying. Hiding the .gauge-value-symbol via CSS leaves the 10px shift for the value.

I am creating a PR to address this.

### Screenshot or recording
image

theoshu:
… symbol and x shift in value

fix #941

### Updates
- add hideSymbol boolean to hide the display of the gauge-value-symbol and prevent the horizontal shifting of the gauge-value-number

- [ ] Demos all features
- [ ] Documented/annotated
- [ ] Matches UI/UX specs
- [ ] Accessible

---
This change is [Reviewable](https://reviewable.io/reviews/carbon-design-system/carbon-charts/942)


natashadecoste (changes requested):
Thanks for creating the issue and this PR to fix it.
Only notes i have is that I think it would be better if you made the configuration key a bit more verbose like "showPercentageSymbol". This would make it more clear that its not the delta symbol toggle but the percentage token. 

And also, if you would extend it so that if there is **no percentage token** on the number value, there would also be **none on the delta value either** (see below). So basically the one configuration would turn on/off the percentage token for both.

image

theoshu:
@natashadecoste Done! I wasn't able to see the changes in the react storyboard. Is there a step I missed?

packages/core/src/components/graphs/gauge.ts
```diff
@@ -253,10 +262,12 @@ export class Gauge extends Component {
...
-		DOMUtils.appendOrSelect(valueNumberGroup, 'text.gauge-value-symbol')
-			.style('font-size', `${valueFontSize(radius) / 2}px`)
-			.attr('x', valueNumberWidth / 2)
-			.text('%');
+		if (showPercentageSymbol) {
+			DOMUtils.appendOrSelect(valueNumberGroup, 'text.gauge-value-symbol')
+				.style('font-size', `${valueFontSize(radius) / 2}px`)
+				.attr('x', valueNumberWidth / 2)
+				.text('%');
+		}
```
theiliad, line 270: this won't work if the user initially has percentages on & then turns them off

packages/core/src/components/graphs/gauge.ts
```diff
@@ -222,10 +228,13 @@ export class Gauge extends Component {
...
 		// Add the big number
-		const valueNumberGroup = DOMUtils.appendOrSelect(
+		const valueNumberGroup = showPercentageSymbol ? DOMUtils.appendOrSelect(
+			numbersGroup,
+			'g.gauge-value-number'
+		).attr('transform', 'translate(-10, 0)').attr('transform', 'translate(-10, 0)') : DOMUtils.appendOrSelect(
 			numbersGroup,
 			'g.gauge-value-number'
-		).attr('transform', 'translate(-10, 0)'); // Optical centering for the presence of the smaller % symbol
+		); // Optical centering for the presence of the smaller % symbol
```
theiliad, line 237: this won't work if the user initially has percentages on & then turns them off

theiliad (changes requested):

theoshu:
@natashadecoste @theiliad 👋

natashadecoste (approved):

sophiiae (approved):

theiliad (approved):

PR #943, opened 2021-01-29
chore(angular): whitelist carbon telemetry dependency for Angular CLI

theiliad:


natashadecoste (approved):

cal-smith (changes requested):
There's several moving parts across these two PRs that I'm concerned about:
- where is the script to add the `postinstall` back to angular?
- https://github.com/carbon-design-system/carbon-charts/blob/master/scripts/clean-package-jsons.js from what I can tell this is still a factor? It's used in `deploy.sh` which is still in use.
- https://github.com/carbon-design-system/carbon-charts/blob/40d1d5982f58bdfcbf2e61fe5a12e5a3321731f4/packages/core/package.json#L19 this isn't going to work - there's several teams on Windows where this won't work. I'd suggest removing _all_ scripts from core after the build and only adding back the telemetry script.

I appreciate the eagerness to get this integrated, but it's clear this hasn't been tested at all.

theiliad:
> There's several moving parts across these two PRs that I'm concerned about:
> 
> * where is the script to add the `postinstall` back to angular?
> * https://github.com/carbon-design-system/carbon-charts/blob/master/scripts/clean-package-jsons.js from what I can tell this is still a factor? It's used in `deploy.sh` which is still in use.
> * https://github.com/carbon-design-system/carbon-charts/blob/40d1d5982f58bdfcbf2e61fe5a12e5a3321731f4/packages/core/package.json#L19
>    this isn't going to work - there's several teams on Windows where this won't work. I'd suggest removing _all_ scripts from core after the build and only adding back the telemetry script.
> 
> I appreciate the eagerness to get this integrated, but it's clear this hasn't been tested at all.

it's been tested locally, however the package.json cleaning script only runs in CI which is a good catch. Angular CLI doesn't remove the postinstall, I checked the dist

scripts/add-telemetry-to-packages.js
```diff
@@ -0,0 +1,34 @@
+const fs = require('fs');
+const path = require('path');
+
+// Define paths to package.json files
+const TELEMETRY_ENABLED_PACKAGES = [
```
cal-smith, line 5: This is excellent! Thanks 👍

cal-smith (approved):
Looks great now :+1:

natashadecoste (approved):

PR #960, opened 2021-02-22
Reset to default domain by setting InitialZoomDomain to null

hlyang397:
### Updates
- User can set InitialZoomDomain to null to remove the initialZoomDomain (or reset to default domain)

- no UI change for this PR

- [ ] Demos all features
- [ ] Documented/annotated
- [ ] Matches UI/UX specs
- [ ] Accessible


packages/core/src/components/axes/zoom-bar.ts
```diff
@@ -194,6 +194,17 @@ export class ZoomBar extends Component {
...
 				);
+			} else if (newInitialZoomDomain === null && oldInitialZoomDomain != null){
+				// if newInitialZoomDomain is set to null (when oldInitialZoomDomain is not null)
+				// save initialZoomDomain and reset zoom domain to default domain
+				this.model.set(
+					{
+						initialZoomDomain: null,
+						zoomDomain: Tools.merge([], defaultDomain)
+					},
+					{ skipUpdate: true }
```
theiliad, line 205: @hlyang397 have you tested that the `skipUpdate: true` doesn't cause issues for any other elements or chart types? `skipUpdate` is a flag that we should be using **very** carefully

theiliad (approved):

packages/core/src/components/axes/zoom-bar.ts
hlyang397: > @hlyang397 have you tested that the `skipUpdate: true` doesn't cause issues for any other elements or chart types? `skipUpdate` is a flag that we should be using **very** carefully

@theiliad
Thanks for the reminder.
I went through all demos with zoom bar, and I didn't find any issue.
This code change is limited to InitialZoomDomain change (from not null to null), and it only changes `initialZoomDomain` and `zoomDomain` in model, so it shouldn't have impact to other elements or charts.

`skipUpdate` is set to avoid triggering another render.
I just like to update `zoomDomain` here, and then ZoomBar can get the correct `zoomDomain` value later at line 219.
The rendering of zoom bar and chart will be completed in the following code as usual.

hlyang397: > @hlyang397 have you tested that the `skipUpdate: true` doesn't cause issues for any other elements or chart types? `skipUpdate` is a flag that we should be using **very** carefully

@theiliad
Thanks for the reminder.
I went through all demos with zoom bar, and I didn't find any issue.
This code change is limited to InitialZoomDomain change (from not null to null), and it only changes `initialZoomDomain` and `zoomDomain` in model, so it shouldn't have impact to other elements or charts.

`skipUpdate` is set to avoid triggering another render.
I just like to update `zoomDomain` here, and then ZoomBar can get the correct `zoomDomain` value later at line 219.
The rendering of zoom bar and chart will be completed in the following code as usual.

natashadecoste (approved):

PR #969, opened 2021-03-10
feat(core): add bullet graph

theiliad:
image


DianaStanciulescu:
1. i think we should limit to 3 unless we get push against it (we run into some accessibility issues if we don;t)

2. there is an option without any area - that space is shown like a simple rectange
Screenshot 2021-03-12 at 21 46 57

3. no, they can be called whatever people need to call them.

sophiiae (dismissed):

packages/core/demo/data/bullet.ts
```diff
@@ -0,0 +1,56 @@
+import { mouse } from 'd3';
```
natashadecoste, line 1:
```suggestion
```

natashadecoste (changes requested):
the tooltip _values_ are correct, but the color flag seems to only ever use the first default purple
image

theiliad:
> the tooltip _values_ are correct, but the color flag seems to only ever use the first default purple

fixed

DianaStanciulescu:
hi @theiliad , question: why are the X axis values slanted, and if the need to be, why is 0 slanted the other way?

DianaStanciulescu:
also, the quartile marks on top of the colored bar should be black in the dark theme, and white when they are outside the colored bars.

cal-smith (dismissed):

cal-smith:
Just a code review on my part 👍

theiliad:
> hi @theiliad , question: why are the X axis values slanted, and if the need to be, why is 0 slanted the other way?

the slanting is a bug in the axis charts where in some cases the labels (that do fit) are still slanted, have to fix that

sophiiae (approved):

theiliad:
Have to convert to draft again since we're missing:
1. correct legend behaviour #725 
2. wrappers

mjabbink:
@theiliad I think the inside chart background should be white when on G10 background. 

Screen Shot 2021-03-22 at 3 52 30 PM

mjabbink:
Currently all G10.

Screen Shot 2021-03-22 at 3 52 54 PM

mjabbink:
Same for dark G90 dark theme

Screen Shot 2021-03-22 at 3 57 37 PM


Screen Shot 2021-03-22 at 3 57 52 PM

mjabbink:
@DianaStanciulescu  :point_up:

jeanservaas:
Agree about live chart background. I'm pretty sure this was just an oversight and not intentional as far as I can remember. 

So I'm assuming that there is still tone in those first sections, like this:
image

mjabbink:
I wonder in the case of this chart it’s better as it is? Seems almost to get more busy following the suggestion?

DianaStanciulescu:
yes, the bg was not intentional, it should be indeed white. i worked on completely white BG so I missed pointing this out. 

i guess they all look busier with white bg vs g10 general bg,  but at the same time maybe more legible, better contrast, etc?

mjabbink:
And probably more typically on white or G100 but totally guessing. Also best to stay consistent. We can change if we think it will hinder readability later on down the line.

sophiiae (dismissed):

sophiiae (approved):

theiliad:
> @DianaStanciulescu ☝️ FYI @jeanservaas

I just added the BGs on g10 & g90

natashadecoste (dismissed):

sophiiae (approved):

natashadecoste (approved):

PR #971, opened 2021-03-16
chore(react): update readme

grommett:
### Updates
This PR updates the install instructions for React and makes them consistent with the [step-by-step install instructions](https://carbon-design-system.github.io/carbon-charts/?path=/story/tutorials-getting-started--react).

Previously, the install instructions had the user install `@carbon/charts` as part of the install process, but this library brings it in without the _need_ to install it separately. From a conversation with @theiliad the reason it was written this way is because some users had issues where their environment wouldn't resolve `@carbon/charts/styles.css`, for which I added a section addressing this issue.

### Reason
In my case another team member had installed this library. I wanted to use the `GaugeChart` component and had to update to the latest `@carbon/charts-react`. After updating chart styles were only partially applied. After reaching out and digging around in `package.json` I realized that I also needed to update `@carbon/charts`, which could have been prevented with just simply installing `@carbon/charts-react` initially.

- [ ] Demos all features
- [ ] Documented/annotated
- [ ] Matches UI/UX specs
- [ ] Accessible

### DIFF
--- a/packages/react/README.md
+++ b/packages/react/README.md
@@ -7,18 +7,17 @@
 **[Storybook demo sources](https://github.com/carbon-design-system/carbon-charts/tree/master/packages/core/demo/data)**
 
 ## Getting started
-
 Run the following command using [npm](https://www.npmjs.com/):
 
 ```bash
-npm install -S @carbon/charts @carbon/charts-react d3@5.x
+npm install -S @carbon/charts-react d3@5.x
 ```
 
 If you prefer [Yarn](https://yarnpkg.com/en/), use the following command
 instead:
 
 ```bash
-yarn add @carbon/charts @carbon/charts-react d3@5.x
+yarn add @carbon/charts-react d3@5.x
 ```
 
 **Note:** you'd also need to install `carbon-components` if you're not using a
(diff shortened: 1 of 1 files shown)


cal-smith (commented):
Relying on npm/yarn to place `@carbon/charts` in the right place so that the styles will resolve correctly seems pretty brittle to me ... It's not an edge case for users to install styles or use types from `@carbon/charts` so making it an explicit install seems sensible to me.

The other option I see would be _explicitly_ re-exporting the styles and code from `@carbon/charts` in `@carbon/charts-react`, making it truly redundant to also install `@carbon/charts`

theiliad:
> Relying on npm/yarn to place `@carbon/charts` in the right place so that the styles will resolve correctly seems pretty brittle to me ... It's not an edge case for users to install styles or use types from `@carbon/charts` so making it an explicit install seems sensible to me.
> 
> The other option I see would be _explicitly_ re-exporting the styles and code from `@carbon/charts` in `@carbon/charts-react`, making it truly redundant to also install `@carbon/charts`

agreed, the times that I see NPM, bundlers & CI mess this stuff up actually makes me not want to remove the `@carbon/charts` install instruction from our tutorials & READMEs.

If it automatically works, great! If it doesn't, then we're slowing the developer down by trying to either debug the issue or try to figure out that they need to install `@carbon/charts`.

I'll close for now, we can discuss again if need be

grommett:
@theiliad @cal-smith then I would suggest updating your install information [here](https://carbon-design-system.github.io/carbon-charts/?path=/story/docs-getting-started--react) to match.

cal-smith:
> @theiliad @cal-smith then I would suggest updating your install information [here](https://carbon-design-system.github.io/carbon-charts/?path=/story/docs-getting-started--react) to match.

For sure! It's easy for the docs to get out of sync ... If you'd like to open a PR for that we'll gladly take it!

PR #977, opened 2021-03-22
improvement(configuration for gridline/tick alignment)

natashadecoste:
### Updates
override x or y grid to correspond with tick values
it defaults to carbon specs for x and y gridlines (5 on y and 15 on x)

## current behaviour
currently if you customize the tick values for an axis, there is no way to get the gridlines to line up to the tick values
**below has a custom Y axis** 
image

image
codesandbox: https://codesandbox.io/s/zen-water-jocd6?file=/src/index.js:772-805

## new behaviour
if there are custom tick values, the grid can use that:
image

##### options
```
export const simpleBarOptionsCustomTicks = {
	title: 'Simple bar (custom ticks)',
	axes: {
		left: {
			mapsTo: 'value',
			ticks: {
				values: [0, 1.2, 1.3, 2],
			},
		},
		bottom: {
			mapsTo: 'group',
			scaleType: 'labels',
		},
	},
	grid: {
		y: {
			useTickValues: true, // defaults to carbon specs for number of ticks
		}
	}
};
```

- [ ] Demos all features
- [ ] Documented/annotated
- [ ] Matches UI/UX specs
- [ ] Accessible


packages/core/src/configuration.ts
```diff
@@ -73,11 +73,13 @@ export const grid: GridOptions = {
 		// set enable to false will not draw grid and stroke of grid backdrop
 		enabled: true,
 		numberOfTicks: 15,
+		useTickValues: false
```
theiliad, line 76: would `alignWithAxisTicks` be a more intuitive name?

theiliad (changes requested):

theiliad (approved):

sophiiae (approved):

PR #983, opened 2021-03-29
feat(core): add word cloud chart

theiliad:
image


jeanservaas (commented):
Now that all the words are horizontal and assuming that all type dims out as intended on hover the only other comment that I have is that any type larger than 32px should be pulling in Plex Condensed Light (300) weight, not Regular (400) weight.

natashadecoste (dismissed):
its nice on the resize ✨  and i like how easy it is to re-label the tooltip

natashadecoste (approved):

PR #985, opened 2021-03-30
fix(bounded area chart): update bounded area chart so bounds don't sh…

### LINKED ISSUE #980: Min and max bounds with null values are drawn as zero values in bounded area charts
#### I am submitting a...

- [ ] Feature request
- [ ] Design defect
- [x] Source code defect
- [ ] Other

#### charts version:
0.41.44

### Issue description
Null values are drawn as zero values for bounds, while they are ignored for the line.

Set the min or max value in https://carbon-design-system.github.io/carbon-charts/?path=/story/area--bounded-area-time-series-natural-curve to null.

#### Current behavior
The null value of the bound is drawn as a zero value.

#### Expected behavior
Similarly to the line the bounds should be interrupted.

### Screenshot or recording
bounded-area-chart-null

shaziajk:
…ow for null values

Behaviour of bounded area chart now match that of the line chart where null values are shown as a
break in the chart

fix #980

### Updates
- Fix for issue https://github.com/carbon-design-system/carbon-charts/issues/980

Not sure if this is ok, but this is how it looks with a null value in
Screenshot 2021-03-30 at 10 18 41

- [ ] Demos all features
- [ ] Documented/annotated
- [ ] Matches UI/UX specs
- [ ] Accessible

### DIFF
--- a/packages/core/src/components/graphs/area.ts
+++ b/packages/core/src/components/graphs/area.ts
@@ -41,17 +41,16 @@ export class Area extends Component {
 		const { cartesianScales } = this.services;
 
 		const orientation = cartesianScales.getOrientation();
-		const areaGenerator = area().curve(this.services.curves.getD3Curve())
-		.defined((datum: any, i) => {
-			if (boundsEnabled) {
+		const areaGenerator = area()
+			.curve(this.services.curves.getD3Curve())
+			.defined((datum: any, i) => {
 				const rangeIdentifier = cartesianScales.getRangeIdentifier();
 				const value = datum[rangeIdentifier];
 				if (value === null || value === undefined) {
 					return false;
 				}
-			}
-			return true;
-		});
+				return true;
+			});
 
 		// Update the bound data on area groups
 		const groupedData = this.model.getGroupedData(this.configs.groups);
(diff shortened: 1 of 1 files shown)


packages/core/src/components/graphs/area.ts
```diff
@@ -41,7 +41,17 @@ export class Area extends Component {
...
+		.defined((datum: any, i) => {
+			if (boundsEnabled) {
+				const rangeIdentifier = cartesianScales.getRangeIdentifier();
+				const value = datum[rangeIdentifier];
+				if (value === null || value === undefined) {
+					return false;
+				}
+			}
+			return true;
+		});
```
theiliad, line 54: seems like the use of `defined()` is missing in the area chart, can we remove the `if (boundsEnabled)` check to make sure that this also works for regular area charts? Thanks

theiliad (changes requested):

packages/core/src/components/graphs/area.ts
shaziajk: Yep sure should be done now! 
Screenshot 2021-03-30 at 16 43 56

theiliad (approved):
Great work thank you @shaziajk

sophiiae (approved):

PR #1009, opened 2021-04-29
fix(core): fix bullet demo issue

sophiiae:


natashadecoste (approved):
🎉

packages/core/src/interfaces/enums.ts
```diff
@@ -214,7 +214,7 @@ export enum AxisTitleOrientations {
 /**
  * enum of legend item type
  */
-export enum LegendItemType {
+export enum LegendItemTypes {
```
cal-smith, line 217: Just want to make sure the `LegendItemType` -> `LegendItemTypes` change is something that we want, and isn't something that's going to trip up consumers? I might be missing something, but it just seems like an aesthetic change?

packages/core/src/tools.ts
```diff
@@ -295,25 +297,28 @@ export namespace Tools {
 		);
 
 		// Retain default legend additional items
-		if (defaultAdditionalItems && userProvidedAdditionalItems) {
+		if (defaultAdditionalItems && userProvidedAdditionalItems) {
```
cal-smith, line 300: Weird extra space here, not sure why prettier didn't catch that ...
```suggestion
		if (defaultAdditionalItems && userProvidedAdditionalItems) {
```

cal-smith (approved):
Looks good overall, just a couple questions

PR #1013, opened 2021-05-04
HTML Layout components + additional toolbar controls

theiliad:
image
image

UPDATES:
- Changes the layout engine to use HTML instead of SVG (flexbox layouts)
- Optimized axis label rotation logic
- Legends & chart titles are now HTML
- Legend & layout using flexbox now
- Export as CSV/JPG/PNG

TODO:

- [x] Fix skeletons
- [x] Fix toolbar
- [x] Fix scatter tooltip issues
- [x] Cross-browser fullscreen
- [x] Choose between `dom-to-image` & `html-to-image`
- [x] Plex support in image exports
- [x] Moving file downloads into a file service
- [x] Add imports for modal & data-table
- [x] Un-break meter
- [x] Un-break circle pack zooming

Future (separate PRs):
- CSV exports for non-cartesian charts


jeanservaas:
Looking good! Make sure the modal button only spans 50% of the modal... right now it looks like more than that.

And put the 'x' on Keto to shed some pounds!

image

theiliad:
> Looking good! Make sure the modal button only spans 50% of the modal... right now it looks like more than that.
> 
> And put the 'x' on Keto to shed some pounds!
> 
> image

Will do haha!

cal-smith (commented):
Lots of good stuff in here, I'm sure I'll have to do a second pass .... It might be good to get Plex added in a separate PR though, over 2,000 files really bogs down the Github UI.

theiliad:
> Lots of good stuff in here, I'm sure I'll have to do a second pass .... It might be good to get Plex added in a separate PR though, over 2,000 files really bogs down the Github UI.

Found a different way of using plex so we're good

packages/core/src/model-cartesian-charts.ts
```diff
@@ -11,6 +14,50 @@ export class ChartModelCartesian extends ChartModel {
...
+					: datum[rangeIdentifier].toLocaleString(), // range value
+			])
+		);
+
+		let csvString = '',
+			csvData = '';
+		data.forEach(function (d, i) {
+			csvData = d.join(';');
+			csvString += i < data.length ? csvData + '\n' : csvData;
+		});
```
Akshat55, line 56: CSV files should be comma-separated, however from what I’ve read this can vary based on region. So semi-colon will not work in NA by default but may work in other regions? I think it might be best to use commas by default or hardcode the delimiter?

To hardcode the delimiter, change line 51 to:
```js
let csvString = 'sep=;\n',
```

packages/core/src/interfaces/enums.ts
```diff
@@ -1,6 +1,11 @@
 import * as EventEnums from './events';
 export const Events = EventEnums;
 
+export enum RenderTypes {
+	HTML = 'html',
+	SVG = 'svg',
+}
+
```
Akshat55, line 8: Just a heads up - two of the same exports of RenderTypes
1. packages/core/src/interfaces/enums.ts
2. packages/core/src/components/component.ts (line 12 - 16)

Akshat55 (approved):

PR #1041, opened 2021-06-04
fix(core): radar animation issues, axis updates and zooming

theiliad:
1. Stops axes from updating when using the radar legend
2. Optimizes rendering logic and animations in radar
3. Zooming animation is minimized and more controlled


cal-smith (approved):

cameroncalder (commented):
Everything is looking great but I'm seeing the chart resize _just a bit_ when a single source is selected from the legend the first time. It doesn't seem to happen after, though.


https://user-images.githubusercontent.com/50155706/121236218-521acf80-c85b-11eb-96c0-88ad4beb3617.mp4

cameroncalder (approved):

Issue #949, opened 2021-02-04
Avoid data sorting in Donut chart


matias-quezada:
#### I am submitting a...

- [ ] Feature request
- [x] Design defect
- [ ] Source code defect
- [ ] Demo/documentation defect
- [ ] Other

#### charts version:

```json
    "@carbon/charts": "^0.41.30",
    "@carbon/charts-react": "^0.41.30",
    "carbon-components": "^10.22.0",
    "carbon-components-react": "^7.22.0",
```

### Issue description
Donut chart is sorting my data, I want to avoid that

#### Steps to produce the issue
Create a Donut chart where options have a specific order.
Click [here](https://codesandbox.io/s/wonderful-robinson-d3tx2?file=/src/index.js) to see an example.

#### Current behavior
It sorts the data by decreasing order.

#### Expected behavior
It should render the data in the order I give them to it. Or at least provide an option to behave like that.

### Screenshot or recording
![Screenshot 2021-02-04 at 12 48 22](https://user-images.githubusercontent.com/44360321/106889328-25f21500-66e8-11eb-9e9d-0e548f86a21f.png)

matias-quezada:
Related but not duplicate of https://github.com/carbon-design-system/carbon-charts/issues/449

theiliad:
I think that this was a design choice at the time of the donut implementation. @jeanservaas could you please chime in?

vgb1993:
Hy, I have the same problem.

I'm showing multiple Donuts that have the same groups. It's very misleading to have the order altered.

I tried using:

```
"legend": {
 "order": ["JQAI 2M4L1", "YEL48 Q6XK YEL48"]
}
```

But it doesn't work.

Also having to specify the order of the items would make it harder to use in our use case. It would be nice to have a On/Off flag to disable sort.

joan-pascual:
I think I'm facing the same issue. Starting from 12 o'clock the the size of the pieces should be specified, or at least implement a coherent default order, for exemple, descending in size, as i show in the picture
![MicrosoftTeams-image](https://user-images.githubusercontent.com/11899282/111813916-1fa3ab00-88da-11eb-927a-3d2b06934c05.png)

theiliad:
> I think I'm facing the same issue. Starting from 12 o'clock the the order of the pieces should be specifiable, or at least implement a coherent default order, for exemple, in the picture case I always want to see the same order in every Donut to keep it coherent

I'll make sure that design responds to this soon

[...]

PR #1043, opened 2021-06-06
feat(pie): add sortFunction option for pie chart

ntsd:
Relate to #949 because we need the fixed data sorting for pie and doughnut chart
so it would be good if we can add sort function to the pie chart option

### Updates
- Add sortFunction option for pie chart and doughnut chart

### Examples

sort by reversed value 

``` Javascript
option: {
	pie: {
      sortFunction: (a: any, b: any) => {
		return b.value - a.value; // sort by reversed value
      },
    },
}
```

sort by name

``` Javascript
option: {
	pie: {
      sortFunction: (a: any, b: any) => {
		return a.group.localeCompare(b.group); // sort by name
      },
    },
}
```

sort by group order

``` Javascript
const groupOrders = ['group1', 'group2'];

.
.
.

option: {
	pie: {
      sortFunction: (a: any, b: any) => {
		return groupOrders.indexOf(a.group) - groupOrders.indexOf(b.group);  // sort by group orders
      },
    },
}
```

image

- [ ] Demos all features
- [ ] Documented/annotated
- [ ] Matches UI/UX specs
- [ ] Accessible


packages/core/src/components/graphs/pie.ts
```diff
@@ -82,6 +82,12 @@ export class Pie extends Component {
 			.sort(null)
 			.padAngle(Configuration.pie.padAngle);
 
+		// set sort function from options
+		const sortFunction = options?.pie?.sortFunction;
+		if (sortFunction) {
+			pieLayout.sort(sortFunction);
+		}
+
```
theiliad, line 90: We  should add the default sort function here https://github.com/carbon-design-system/carbon-charts/blob/8f3f8fb112277fb3fa609273dcbd27fc39a39a7d/packages/core/src/configuration.ts#L375-L383

That way there should be no need to have the if-statement

packages/core/src/components/graphs/pie.ts
```diff
@@ -82,6 +82,12 @@ export class Pie extends Component {
...
 		// Sort pie layout data based off of the indecies the layout creates
 		const pieLayoutData = pieLayout(displayData).sort(
```
theiliad, line 92: should remove the `.sort()` here

packages/core/src/components/graphs/pie.ts
```diff
@@ -82,6 +82,12 @@ export class Pie extends Component {
 			.sort(null)
```
theiliad, line 82:
```suggestion
			.sort(Tools.getProperty(options, "pie", "sortFunction"))
```

theiliad: `Tools.getProperty()` returns null for any null or undefined values

theiliad (changes requested):

packages/core/src/configuration.ts
```diff
@@ -379,6 +379,7 @@ const pieChart: PieChartOptions = Tools.merge({}, chart, {
 			enabled: true,
 		},
 		alignment: Alignments.LEFT,
+		sortFunction: null,
```
theiliad, line 382:
```suggestion
		sortFunction: (a: any, b: any) => a.index - b.index
```

theiliad (changes requested):
Pls also remove this piece 
image

packages/core/src/configuration.ts
ntsd: this doesn't work because data inside `pieLayout` only have `{group: "YEL48 Q6XK YEL48", value: 10000}` which is data that put into `const pieLayoutData = pieLayout(displayData);` and I think the sort below at line 86 doesn't sort anything because when the data change d3 pie will using the sort function from the layout. I think the best way is to use null for default sort.

theiliad (approved):

PR #1048, opened 2021-06-22
feat(highlight): Add highlights

shaziajk:
### Updates
Add in Highlight feature for graphs!!! 

Screenshot 2021-07-01 at 17 30 43

- [ ] Demos all features
- [ ] Documented/annotated
- [ ] Matches UI/UX specs
- [ ] Accessible


jeanservaas:
Hi @shaziajk thanks for pairing with me on this one and tweaking it live!

So this is where we landed. 

- Added a Magenta 10 fill highlight*
- 1px dashed stroke boarder for accessibility

* Don't use red unless you want the highlight to have a negative connotation 

image

Also, try adding the exact same treatment to the zoombar and quick share that with me too in case we need to tweak... But I hope works the same way on the zoom bar.

jeanservaas:
@shaziajk 

All looks good apart from the fact that the top dashed lines are getting cut off 

image

image

We talked about this on Slack, it'd be great if we can find out how to get them even, but apart from that everything looks good!

shaziajk:
@jeanservaas That should all be done now!

packages/core/demo/data/area.ts
```diff
@@ -1,3 +1,6 @@
+import { timeFormat } from 'd3';
```
theiliad, line 1: We can't import stuff directly from `d3`, that'll ruin tree-shaking. Pls switch to the correct package, which I think is `d3-time-format`

theiliad (changes requested):

packages/core/demo/data/area.ts
shaziajk: @theiliad have removed this, it wasn't being used and was left over when I was adding stuff in

theiliad:
@shaziajk sorry forgot about this PR...

are there any concerns here? or anything else we need to look at?

mjabbink:
I think this looks good and is what I remember discussing but @jeanservaas will be better to approve on those details.

shaziajk:
> @shaziajk sorry forgot about this PR...
> 
> are there any concerns here? or anything else we need to look at?

No problem! There are no concerns or anything else that needs to be looked at from my end

jeanservaas (approved):
Looks good to me!

theiliad:
> Looks good to me!

@jeanservaas @shaziajk do we only have a red highlight option? do we need other colors?

does this support horizontal highlights as well? should it?

shaziajk:
@theiliad the default colour is one of the carbon magenta colours but the user can specify the colour they want in a similar way to how they can define the colour for some of the other chart components i.e. using:
```
"color": {
          "scale": {
            "Custom formatter": "orange"
          }
        },
```

It also supports horizontal highlights as well, however this does not show on the zoom bar as I am not sure if that was needed as it nots as effective as the vertical version but can be added in if necessary?

theiliad:
> It also supports horizontal highlights as well, however this does not show on the zoom bar as I am not sure if that was needed as it nots as effective as the vertical version but can be added in if necessary?

I think that makes sense

jeanservaas:
@theiliad what Shazia said makes sense to me. There weren't extensive explorations done with horizontal highlights or various palette options. But we can go out with this for now and if we find we need a deeper design dive we can resource it.

theiliad:
@shaziajk do you know why the angular demos have an error?

image

shaziajk:
Hmm im not sure why that errors showing 🤔 I didnt think I made any changes which could cause that
But ill take a look and see if i can spot anything that I might have done which have caused it

theiliad:
> Hmm im not sure why that errors showing 🤔 I didnt think I made any changes which could cause that
> But ill take a look and see if i can spot anything that I might have done which have caused it

no I think it's just randomly happening. going to merge and check

PR #1116, opened 2021-08-11
feat(core): add tree chart

theiliad:
TODO:
- [x] Clean up code
- [x] add enum for `treeType`
- [x] Wrappers


packages/core/src/components/graphs/tree.ts
```diff
@@ -0,0 +1,270 @@
...
+	// 				element: hoveredElement,
+	// 				datum,
+	// 			});
+
+	// 			// Hide tooltip
+	// 			self.services.events.dispatchEvent(Events.Tooltip.HIDE, {
+	// 				hoveredElement,
+	// 			});
+	// 		});
+	// }
```
Akshat55, line 269: Commented out events for pie

theiliad: yeah adding those events right now

PR #1132, opened 2021-09-03
feat(core): create alluvial chart

### LINKED ISSUE #871: Alluvial diagram - Explorations
Working [Box Note](https://ibm.box.com/s/yba0ie4522utyyfain342fxukn4lzbn1)

Akshat55:
### Updates
- Create alluvial chart component
- Created alluvial demos
- Exported alluvial in all packages

closes #871

https://user-images.githubusercontent.com/38994122/131957958-a8cabed5-3ed5-40c9-b816-82d460cdb864.mov

- [ ] Demos all features
- [ ] Documented/annotated
- [ ] Matches UI/UX specs
- [ ] Accessible


packages/core/src/interfaces/charts.ts
```diff
@@ -486,3 +486,25 @@ export interface CirclePackChartOptions extends BaseChartOptions {
...
+	}>;
+	/**
+	 * Set the node padding
+	 */
+	nodePadding?: number;
+	/**
+	 * Enable single color usage for lines
+	 */
+	monochrome?: boolean;
+}
```
theiliad, line 510: can we scope all this under `{ alluvial: {...} }`?

similar to other chart options

packages/core/src/components/graphs/alluvial.ts
```diff
@@ -0,0 +1,596 @@
...
+		<defs>
+			<style>
+			.cls-1 {
+				fill: none;
+			}
+			</style>
+		</defs>
+		<polygon points="18 6 16.57 7.393 24.15 15 4 15 4 17 24.15 17 16.57 24.573 18 26 28 16 18 6"/>
+		<rect id="_Transparent_Rectangle_" data-name="&lt;Transparent Rectangle&gt;" class="cls-1" width="32" height="32"/>
+		</svg>`;
```
theiliad, line 584: could you pls clean up this svg? we can remove the classname & just add the style as an attribute to the `rect`, and the `id` seems redundant bc. if we have multiple, then it's a problem

packages/core/src/components/graphs/alluvial.ts
```diff
@@ -0,0 +1,596 @@
...
+		this.parent
+			.selectAll('path.link')
+			.on('mouseover', function (event, datum) {
+				const hoveredElement = select(this);
+				debouncedLineHighlight(this, 'mouseover');
+				hoveredElement.classed('hovered', true);
+
+				const strokeColor = getComputedStyle(
+					this,
+					null
```
theiliad, line 276: I think you can drop the null for `getComputedStyle`

packages/core/src/components/graphs/alluvial.ts
```diff
@@ -0,0 +1,596 @@
...
+			.attr('height', 18)
+			.attr('stroke-width', 2)
+			.lower();
+
+		// Position group based on text width
+		textNode.attr('transform', (d, i) => {
+			const { width } = this.parent
+				.select(`text#node-text-${i}`)
+				.node()
+				.getBBox();
```
theiliad, line 208: getBBox() is known to have issues in firefox, depending on the version I think & the conditions of the element. We should probably use

```
const { width } = DOMUtils.getSVGElementSize(..., {
	useBBox: true,
});
```

packages/core/src/components/graphs/alluvial.ts
```diff
@@ -0,0 +1,596 @@
...
+		// Text background
+		textNode
+			.append('rect')
+			.classed('node-text-bg', true)
+			.attr('width', (d, i) => {
+				// Determine rectangle width based on text width
+				const { width } = this.parent
+					.select(`text#node-text-${i}`)
+					.node()
+					.getBBox();
```
theiliad, line 196: same note as below re. `getBBox()`

packages/core/src/components/graphs/alluvial.ts
```diff
@@ -0,0 +1,596 @@
...
+					return nodeCoordinates[d];
+				}
+				return '';
+			})
+			.attr('y', 20)
+			.attr('x', (d, i) => {
+				const { width } = this.parent
+					.select(`text#category-${i}`)
+					.node()
+					.getBBox();
```
theiliad, line 102: same note as below for `getBBox()`

theiliad (changes requested):

packages/core/src/components/graphs/alluvial.ts
```diff
@@ -0,0 +1,596 @@
...
+			]);
+
+		// Construct a graph with the provided user data
+		this.graph = sankey({
+			nodes: options.nodes.map((d) => Object.assign({}, d)),
+			links: data.map((d) => Object.assign({}, d)),
+		});
+
+		// Filter out unused nodes so they are not rendered
+		this.graph.nodes = this.graph.nodes.filter((node) => node.value !== 0);
```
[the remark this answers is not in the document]
Akshat55: That would require additional computation. I'm filtering the nodes after the graph has been created so I wouldn't have to construct a set and iterate through the data then remove nodes that aren't being used. When the graph is created, the nodes all have values equal to the number of edges they have so, if a node has a value of 0 it should be removed.

packages/core/src/components/graphs/alluvial.ts
```diff
@@ -0,0 +1,596 @@
...
+// D3 imports
+import { select } from 'd3-selection';
+import { sankey as d3Sankey, sankeyLinkHorizontal } from 'd3-sankey';
+
+export class Alluvial extends Component {
+	type = 'alluvial';
+	renderType = RenderTypes.SVG;
+
+	private graph: any;
+	private width;
```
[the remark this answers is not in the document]
Akshat55: No, we do not

packages/core/src/components/graphs/alluvial.ts
```diff
@@ -0,0 +1,596 @@
...
+		}
+
+		const sankey = d3Sankey()
+			.nodeId((d) => d.name)
+			.nodeWidth(Configuration.alluvial.nodeWidth)
+			// Distance nodes are apart from each other
+			.nodePadding(nodePadding)
+			// size of the chart and its padding
+			.extent([
+				[2, 30],
```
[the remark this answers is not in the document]
Akshat55: I've added the following comments to clarify, for some reason, it isn't being shown

```js
// Size of the chart and its padding
// Chart starts at 2 and ends at width - 2 so the outer nodes can expand from center
// Chart starts from 30 so node categories can be displayed
```

theiliad (approved):

PR #1158, opened 2021-09-20
fix(core): allow users to define custom buttons in toolbar

### LINKED ISSUE #1129: Toolbar - Allow custom toolbar controls
#### I am submitting a...

- [x] Feature request
- [ ] Design defect
- [ ] Source code defect
- [ ] Other

#### charts version:

v0.46.1

### Issue description

We would like to extend the use cases of the toolbar. This would allow a neat and integrated way for us develop our ui. 
After trying and looking at the source code we found that the toolbar can not be configured with custom toolbar controls. 

Passing in this will fail:
`"controls": [
		{
				"type": "Custom",
				"id": "PIN_CHART",
				"label": "Pin chart",
		},
		{
				"type": "Custom",
				"id": "SHOW_HELP",
				"label": "Show help",
				"icon": "....",
		}
]`

#### Current behavior

All the non predefined controls throw an exception in [toolbat.ts](https://github.com/carbon-design-system/carbon-charts/blob/master/packages/core/src/components/axes/toolbar.ts) ln. 339

`
switch (controlType) {
	.....
	default:
		throw Error('Not supported toolbar control type: ' + controlType);
}
`

#### Expected behavior

Custom toolbar controls should not throw.

Within the toolbar Options a icon could or could not be specified for a control, in that case a default icon should be assigned.

A click handler should be exposed, emmiting the id of the toolbar control.

### Screenshot or recording

I edited using devtools to showcase what we want:

image

### LINKED ISSUE #1237: [Toolbar]:  Adding custom label/html and custom icons on toolbar options
### Contact Details

adyasha.das1@ibm.com

### Summary

It is found that the toolbar can not be configured with custom toolbar controls. We expect some more customization properties like adding html, custom label on the grid view and custom tool-control options. This will allow us to integrate the chart into the UI in a more cleaner way. 

### Justification

image

- we expect custom dropdown options here. It duplicates the values if specific control types are added

image

- Instead of default grid header title we expect custom title and custom time format column.

Allowing developer to add the custom feature would help us for a cleaner integration of chart and it's details in the UI. It will improve the user experience.

### "Must have" functionality

- Custom dropdown options in the download tool.
- Custom grid header title.
- Custom html option for the Control-toolbar.

It would be very helpful if requested features would be taken care asap. We really have a requirement in our current application.

### Available extra resources

--

Akshat55:
### Updates
- Fixed typo (‘hight’)
- Allow users to add ‘custom’ icons along with height & width
- Improve keyboard events (Arrow keys & selection with both spacebar and enter)
- Emit events on button click including preconfigured buttons
- Allow users to pass in a clickFunction if they don’t want to
- Assign a custom ID to the button if the user doesn't pass it in
- **If button.length is less than number of icons and user doesn't pass in the svg element content, the icon will be placed in the overflow menu including all of the succeeding buttons.**
- Improve toolbar options buttons accessibility 

fix #1129 

Closes #1237

image

- [ ] Demos all features
- [ ] Documented/annotated
- [ ] Matches UI/UX specs
- [ ] Accessible


packages/core/src/components/axes/toolbar.ts
```diff
@@ -180,9 +208,11 @@ export class Toolbar extends Component {
 		enteringOverflowMenuControls
 			.merge(overflowMenuControls)
 			.classed('bx--overflow-menu-options__option--disabled', (d) =>
-				d.shouldBeDisabled()
+				d.shouldBeDisabled ? d.shouldBeDisabled() : false
+			)
+			.attr('aria-disabled', (d) =>
+				d.shouldBeDisabled ? d.shouldBeDisabled() : false
```
theiliad, line 214: Looks like lower in the code we're setting a default for `d.shouldBeDisabled`, right? Then do we need the check?

packages/core/src/components/axes/toolbar.ts
```diff
@@ -340,17 +370,39 @@ export class Toolbar extends Component {
...
+							{
+								event,
+								element: menuItem,
+							}
+						);
+
+						// call the specified function if it exists
+						if (typeof menuItem.clickFunction === 'function') {
+							menuItem.clickFunction(event);
+						}
```
theiliad, line 385: I think I saw this same piece of logic 4 times, can we re-use it?

packages/core/src/components/axes/toolbar.ts
```diff
@@ -379,47 +431,81 @@ export class Toolbar extends Component {
...
+
 		controls.forEach((control) => {
-			const controlConfig = this.getControlConfigByType(control.type);
+			let controlConfig = null;
+			// check if button is custom or default control
+			if (control.type === ToolbarControlTypes.CUSTOM) {
+				// add generic id if missing
+				if (Tools.getProperty(control, 'id') === null) {
+					// add id directly to the data passed so that id isn't reassigned on rerender
+					control.id = `toolbar-button-${Toolbar.buttonID}`;
```
theiliad, line 456:
```suggestion
					control.id = `toolbar-button-${Toolbar.buttonID++}`;
```

packages/core/src/components/axes/toolbar.ts
```diff
@@ -379,47 +431,81 @@ export class Toolbar extends Component {
...
+					Toolbar.buttonID++;
```
theiliad, line 457:
```suggestion
```

theiliad: why not do the above?

packages/core/src/components/axes/toolbar.ts
```diff
@@ -379,47 +431,81 @@ export class Toolbar extends Component {
...
+					// if button does not exist, all of the preceding items will be added to overflow menu
+					// this will ensure users use correct order & fill in any missing data
+					if (
+						Tools.getProperty(
+							controlConfig,
+							'iconSVG',
+							'content'
+						) === null
+					) {
+						displayIcons = false;
```
theiliad, line 486: hmmm... why do this? why wouldn't we just move anything we find with no icons into the overflow menu?

theiliad: this seems like an undocumented or unexpected side-effect

packages/core/src/components/axes/toolbar.ts
```diff
@@ -379,47 +431,81 @@ export class Toolbar extends Component {
...
+					overflowList.push(controlConfig);
 				}
 			}
 		});
 
-		if (
-			controlList.length <= numberOfIcons &&
-			overflowSpecificControls.length === 0
-		) {
+		// Ensures exports are always at the bottom
```
theiliad, line 497:
```suggestion
		// Ensures the `export` controls are always at the bottom
```

theiliad (changes requested):

theiliad (approved):

Likith-Chintala:
what did you code bro @theiliad it never works

PR #1159, opened 2021-09-20
fix(core): correctly draw the threshold from multiple axes

### LINKED ISSUE #1153: [Bug]: Threshold is being applied to the wrong axes in combochart
### Contact Details

_No response_

### What happened?

Threshold is being drawn only on the main y-axis, threshold cannot be applied to the opposite y-axis. If the y-axis ticks do not match, the data is misrepresented.

image

(Threshold should be on y1 axis at 18 instead of the main axis which is set as y2).

### Version

Latest - 0.48.4

### Data & options used

_No response_

### Relevant log output

_No response_

### Codesandbox example

_No response_

Akshat55:
### Updates
- Correctly draw the threshold from all axes regardless the main axes.

fix #1153 

image
image

- [ ] Demos all features
- [ ] Documented/annotated
- [ ] Matches UI/UX specs
- [ ] Accessible

### DIFF
--- a/packages/core/src/components/essentials/threshold.ts
+++ b/packages/core/src/components/essentials/threshold.ts
@@ -91,8 +91,7 @@ export class Threshold extends Component {
 				d.thresholds.map((threshold) => {
 					// Merge relevant keys into the threshold object
 					threshold.axisPosition = d.axisPosition;
-					threshold.correspondingDatasets = d?.correspondingDatasets;
-					threshold.mapsTo = d?.mapsTo;
+					threshold.datum = this.constructDatumObj(d, threshold);
 					return threshold;
 				})
 			);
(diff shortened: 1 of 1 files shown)


packages/core/src/components/essentials/threshold.ts
```diff
@@ -86,7 +89,10 @@ export class Threshold extends Component {
 			.selectAll('g.threshold-group')
 			.data((d) =>
 				d.thresholds.map((threshold) => {
+					// Merge relevant keys into the threshold object
 					threshold.axisPosition = d.axisPosition;
+					threshold.correspondingDatasets = d?.correspondingDatasets;
+					threshold.mapsTo = d?.mapsTo;
 					return threshold;
```
theiliad, line 96: seems like `self.constructDatumObj(d)` should just run here right? edit the data right in the beginning instead of doing it multiple times

theiliad (changes requested):

theiliad (approved):
LGTM

PR #1200, opened 2021-10-25
feat(core): create heatmap chart 

Akshat55:
### Updates
- Create heat map
- Create custom color scale (Color scale is WIP)
- Export to all charting libraries

image

- [ ] Demos all features
- [ ] Documented/annotated
- [ ] Matches UI/UX specs
- [ ] Accessible


packages/core/src/styles/graphs/_heatmap.scss
```diff
@@ -0,0 +1,96 @@
...
+		}
+
+		rect.null-state {
+			fill: #f4f4f4;
+		}
+	}
+
+	@if $carbon--theme == $carbon--theme--g10 {
+		rect.heat {
+			stroke: #f4f4f4;
```
theiliad, line 69: are there carbon tokens that have these values?

theiliad: you can probably ask @DianaStanciulescu or @jeanservaas

packages/core/src/services/scales-cartesian.ts
```diff
@@ -214,6 +214,15 @@ export class CartesianScales extends Service {
...
 	}
 
+	getCustomDomainValuesByposition(axisPosition: AxisPositions) {
+		return Tools.getProperty(
+			this.model.getOptions(),
+			'axes',
+			axisPosition,
+			'domain'
+		);
+	}
```
theiliad, line 224: let's add that validation here... isArray && length === 2, and we can probably send a console.error if the wrong format is provided

packages/core/src/model/heatmap.ts
```diff
@@ -0,0 +1,326 @@
...
+			this.getDisplayData().forEach((d, i) => {
+				this._matrix[d[domainIdentifier]][d[rangeIdentifier]] = {
+					value: d['value'],
+					index: i,
+				};
+			});
+		}
+
+		return this._matrix;
+	}
```
theiliad, line 195: Do we absolutely need all the processing?

The answer doesn't have to be a no, but if it is then we should simplify

packages/core/src/interfaces/enums.ts
```diff
@@ -250,3 +250,20 @@ export enum LegendItemType {
...
+/**
+ * enum of color legend types
+ */
+export enum ColorLegendType {
+	LINEAR = 'linear',
+	QUANTIZE = 'quantize',
+}
+
+/**
+ * enum of axis ticks rotation
```
theiliad, line 263: is this the right comment?

packages/core/src/interfaces/components.ts
```diff
@@ -44,6 +45,18 @@ export interface LegendOptions {
...
+	 * enabled by default on select charts
+	 */
+	colorLegend?: {
+		/**
+		 * Text to display beside or on top of the legend
+		 * Position is determined by text length
+		 */
+		title?: string;
+		type: ColorLegendType;
+	};
```
theiliad, line 59: would this add `colorLegend` to the types for the rest of the charts? is that what we want?

packages/core/src/interfaces/charts.ts
```diff
@@ -137,7 +138,14 @@ export interface BaseChartOptions {
...
 		 * e.g. { enabled: true }
 		 */
-		gradient?: object;
+		gradient?: {
+			enabled?: boolean;
+			/**
+			 * hex color array
+			 * e.g. ['#fff', '#000', ...]
+			 */
+			colors?: Array<String>;
```
theiliad, line 147: do we want this applied to other chart types?

packages/core/src/charts/heatmap.ts
```diff
@@ -0,0 +1,188 @@
...
+			)
+		);
+
+		// Initialize data, services, components etc.
+		this.init(holder, chartConfigs);
+	}
+
+	// Custom getChartComponents - Implements getChartComponents
+	// Removes zoombar support and additional `features` that are not supported in heatmap
+	private getChartComponentsList(graphFrameComponents: any[], configs?: any) {
```
theiliad, line 54: this should just go into `getChartComponents`...?

packages/core/src/components/axes/hover-axis.ts
```diff
@@ -0,0 +1,264 @@
...
+import { AxisPositions, Events, ScaleTypes } from '../../interfaces';
+import { ChartModel } from '../../model/model';
+import { DOMUtils } from '../../services';
+import { Tools } from '../../tools';
+import * as Configuration from '../../configuration';
+
+// D3 Imports
+import { select } from 'd3-selection';
+
+export class HoverAxis extends Axis {
```
theiliad, line 12:
```suggestion
export class HoverableAxis extends Axis {
```

packages/core/src/components/axes/hover-axis.ts
```diff
@@ -0,0 +1,264 @@
...
+	constructor(model: ChartModel, services: any, configs?: any) {
+		super(model, services, configs);
+	}
+
+	render(animate = true) {
+		super.render(animate);
+		// Remove existing event listeners to avoid flashing behavior
+		super.destroy();
+		const { position: axisPosition } = this.configs;
```
theiliad, line 21:
```suggestion

		const { position: axisPosition } = this.configs;
```

packages/core/src/components/axes/hover-axis.ts
```diff
@@ -0,0 +1,264 @@
...
+import { select } from 'd3-selection';
+
+export class HoverAxis extends Axis {
+	constructor(model: ChartModel, services: any, configs?: any) {
+		super(model, services, configs);
+	}
+
+	render(animate = true) {
+		super.render(animate);
+		// Remove existing event listeners to avoid flashing behavior
```
theiliad, line 19:
```suggestion

		// Remove existing event listeners to avoid flashing behavior
```

packages/core/src/components/axes/two-dimensional-hover-axes.ts
```diff
@@ -0,0 +1,122 @@
+// Internal Imports
+import { TwoDimensionalAxes } from './two-dimensional-axes';
+import { AxisPositions } from '../../interfaces';
+import { Tools } from '../../tools';
+import { DOMUtils } from '../../services';
+import { Events } from './../../interfaces';
+import { HoverAxis } from './hover-axis';
+
+export class TwoDimensionalHoverAxes extends TwoDimensionalAxes {
```
theiliad, line 9:
```suggestion
export class TwoDimensionalHoverableAxes extends TwoDimensionalAxes {
```

packages/core/src/components/axes/two-dimensional-hover-axes.ts
```diff
@@ -0,0 +1,122 @@
...
+
+			Object.keys(this.children).forEach((childKey) => {
+				const child = this.children[childKey];
+				child.margins = this.margins;
+			});
+
+			this.render(true);
+		}
+	}
+}
```
theiliad, line 122: is this an almost identical copy of `TwoDimensionalAxis`?

If yes, then we should avoid that...

For instance, if you'd like to do `new HoverAxis()` you can just set that as an attribute in the class and only extend that...

e.g.

```ts
private AxisComponent = Axis;
```


```ts
private AxisComponent = HoverableAxis;
```

packages/core/src/components/essentials/color-scale-legend.ts
```diff
@@ -0,0 +1,327 @@
...
+			'legend',
+			'colorLegend',
+			'title'
+		);
+
+		const customColorsEnabled = !Tools.isEmpty(customColors);
+		const domain = this.model.getValueDomain();
+
+		const svg = this.getComponentContainer();
+		svg.html('').attr('role', Roles.GROUP);
```
theiliad, line 128: so we're setting the SVG innerHTML to '' everytime? is there not a better way to render using d3 here?

packages/core/src/components/essentials/color-scale-legend.ts
```diff
@@ -0,0 +1,327 @@
...
+		const group = svg.append('g').classed('legend-rectangle', true);
```
theiliad, line 129:
```suggestion

		const group = svg.append('g').classed('legend-rectangle', true);
```

packages/core/src/interfaces/events.ts
```diff
@@ -65,6 +65,9 @@ export enum Axis {
 	LABEL_MOUSEMOVE = 'axis-label-mousemove',
 	LABEL_CLICK = 'axis-label-click',
 	LABEL_MOUSEOUT = 'axis-label-mouseout',
+	LABEL_FOCUS = 'axis-label-focus',
+	LABEL_BLUR = 'axis-label-blur',
+	RENDER_COMPLETE = 'axis-render-complete',
```
theiliad, line 70: seems like these are added only for the new HoverableAxis components, we should also add them to the existing axis components

theiliad (changes requested):
Overall I think it looks ok, just a couple of considerations around efficiency and duplication

packages/core/src/services/scales-cartesian.ts
Akshat55: Fixed:

```ts
	getCustomDomainValuesByposition(axisPosition: AxisPositions) {
		const domain = Tools.getProperty(
			this.model.getOptions(),
			'axes',
			axisPosition,
			'domain'
		);

		// Check if domain is an array
		if (domain && !Array.isArray(domain)) {
			throw new Error(
				`Domain in ${axisPosition} axis is not a valid array`
			);
		}

		// Determine number of elements passed in domain depending on scale types
		if (Array.isArray(domain)) {
			if (
				(this.scaleTypes[axisPosition] === ScaleTypes.LINEAR ||
					this.scaleTypes[axisPosition] === ScaleTypes.TIME) &&
				domain.length !== 2
			) {
				throw new Error(
					`There can only be 2 elements in domain for scale type: ${this.scaleTypes[axisPosition]}`
				);
			}
		}

		return domain;
	}
```

packages/core/src/interfaces/charts.ts
Akshat55: Yes, I think it makes sense to define what object attribute it accepts, setting it to `object` is not documentation friendly.

[https://charts.carbondesignsystem.com/documentation/interfaces/_interfaces_charts_.basechartoptions.html#color](https://charts.carbondesignsystem.com/documentation/interfaces/_interfaces_charts_.basechartoptions.html#color)

That being said, not all chart types support gradient.

packages/core/src/interfaces/components.ts
Akshat55: The option currently is **ONLY** used by Heatmap - if we are going to have future charts that use color legend, it makes sense to put it under the legend option, since it is a legend. 

Setting the ``type`` or ``title`` will not do enable or do anything in other chart types.

packages/core/src/components/essentials/color-scale-legend.ts
Akshat55: Fixed, switched over to AppendOrSelect for commonly used elements like title & axis and wrapper (<g>)

packages/core/src/interfaces/events.ts
Akshat55: I have added `RENDER_COMPLETE` event to all charts, but the `LABEL_FOCUS` && `LABEL_BLUR` events I cannot since we do not provide axis focus via tab-index on non-hoverable cartesian charts.

packages/core/src/components/axes/two-dimensional-hover-axes.ts
Akshat55: Removed `two-dimensional-hover-axes.ts` file. Now cartesian models define what type of axes to create via `protected axisFlavor = AxisFlavor.DEFAULT;` variable. Defined it in main `ChartModelCartesian` set to default and over rid the variable in heatmap model to `AxisFlavor.HOVER`.

packages/core/src/components/axes/two-dimensional-hover-axes.ts
Akshat55: No longer applicable

packages/core/src/interfaces/enums.ts
Akshat55: Fixed

packages/core/src/styles/graphs/_heatmap.scss
Akshat55: Fixed

packages/core/src/interfaces/enums.ts
```diff
@@ -250,3 +250,28 @@ export enum LegendItemType {
...
+	AUTO = 'auto',
+	OFF = 'off',
+}
+
+/**
+ * enum of axis flavor
+ */
+export enum AxisFlavor {
+	DEFAULT = 'default',
+	HOVER = 'hover',
```
theiliad, line 276:
```suggestion
	HOVERABLE = 'hoverable',
```

theiliad (changes requested):

packages/core/src/components/axes/hover-axis.ts
theiliad: Also the name of the file

packages/core/src/model/heatmap.ts
Akshat55: I think it is good to have this processing, it only happens on chart creation but it speeds up our axes hover. Instead of iterating through entire data length (n), we can access the row/column by simply accessing the object key (constant).

Additionally, it allows us to find any cells that are empty in the matrix.

Usage can be found here - [https://github.com/carbon-design-system/carbon-charts/blob/b57e14cdf3db0d02fd454ad356aa40a1e1996213/packages/core/src/components/graphs/heatmap.ts#L352](https://github.com/carbon-design-system/carbon-charts/blob/b57e14cdf3db0d02fd454ad356aa40a1e1996213/packages/core/src/components/graphs/heatmap.ts#L352)

Akshat55: Updated efficiency from n^2 to 2n

packages/core/src/interfaces/enums.ts
Akshat55: Fixed

packages/core/src/interfaces/components.ts
theiliad: Let's (for now) extend this to a `HeatmapLegendOptions` and only include there, as we haven't figured out how this new legend type will work with other chart types

theiliad (changes requested):
Just the 2 remaining changes for `hover-axis.ts` & the legend types requested above

Also @Akshat55 could you pls add the `experimental` flag to the demos?

Akshat55:
@theiliad, made them experimental and moved color legend under heatmap options

theiliad (approved):
LGTM, tried to look at everything 👍🏻

PR #1207, opened 2021-11-01
fix(core): render chart axis labels correctly when data loads async

### LINKED ISSUE #1135: [Bug]: Radar chart axis labels do not update when data loads async
### Contact Details

alecbarrett@newschool.edu

### What happened?

When radar chart data is loaded asynchronously, the Y axis labels end up at (0,0) with `NaN` positions:

image

The reason this happens is that if `data = []`, then the Y axis labels get `.data([undefined, undefined])`. When the [update function is called](https://github.com/carbon-design-system/carbon-charts/blob/master/packages/core/src/components/graphs/radar.ts#L663-L679), `.attr('x')` was previously `NaN` so the transition cannot transition to a new valid numerical value.

I believe the `NaN` is introduced in `polarToCartesianCoords` ([code](https://github.com/carbon-design-system/carbon-charts/blob/master/packages/core/src/services/angle-utils.ts#L78-L86)): if `r` is undefined in that function, then `x` and `y` are `NaN`. Perhaps that function could return `{ x: 0, y: 0 }` instead if `r` is undefined.

Here's a minimal reproducible example illustrating this property of d3: https://observablehq.com/d/84cc622860cab954

### Version

0.46.4

### Data & options used

_No response_

### Relevant log output

_No response_

### Codesandbox example

https://codesandbox.io/s/solitary-shadow-09gxx?file=/src/components/chart.vue

Akshat55:
fix #1135

### Updates
- NaN coordinates are positioned at (0, 0)
  - If value equals NaN, then we reassign it value 0 to allow further operations.

https://user-images.githubusercontent.com/38994122/139632321-8b1de3fe-af4f-463c-87ce-fe6963de5423.mov

- [ ] Demos all features
- [ ] Documented/annotated
- [ ] Matches UI/UX specs
- [ ] Accessible

### DIFF
--- a/packages/core/src/services/angle-utils.ts
+++ b/packages/core/src/services/angle-utils.ts
@@ -80,8 +80,8 @@ export function polarToCartesianCoords(
 	r: number,
 	t: Point = { x: 0, y: 0 }
 ): Point {
-	let x = r * Math.cos(a) + t.x;
-	let y = r * Math.sin(a) + t.y;
+	const x = r * Math.cos(a) + t.x;
+	const y = r * Math.sin(a) + t.y;
 
 	// NaN is rendered at coordinate 0 in browsers
 	// By setting it to 0, further operations can be performed
--- a/packages/core/src/services/angle-utils.ts
+++ b/packages/core/src/services/angle-utils.ts
@@ -80,16 +80,12 @@ export function polarToCartesianCoords(
 	r: number,
 	t: Point = { x: 0, y: 0 }
 ): Point {
-	const x = r * Math.cos(a) + t.x;
-	const y = r * Math.sin(a) + t.y;
+	let x = r * Math.cos(a) + t.x;
+	let y = r * Math.sin(a) + t.y;
 
 	// NaN is rendered at coordinate 0 in browsers
 	// By setting it to 0, further operations can be performed
-	if (isNaN(x) || isNaN(y)) {
-		return { x: 0, y: 0 };
-	}
-
-	return { x, y };
+	return { x: isNaN(x) ? 0 : x, y: isNaN(y) ? 0 : y };
 }
 
 // Return the distance between a point (described with polar coordinates)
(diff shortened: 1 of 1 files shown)


packages/core/src/services/angle-utils.ts
```diff
@@ -82,6 +82,13 @@ export function polarToCartesianCoords(
 ): Point {
 	const x = r * Math.cos(a) + t.x;
 	const y = r * Math.sin(a) + t.y;
+
+	// NaN is rendered at coordinate 0 in browsers
+	// By setting it to 0, further operations can be performed
+	if (isNaN(x) || isNaN(y)) {
```
theiliad, line 88: shouldn't this just set x to 0 if it's NaN, and same for y?

are they always both NaN?

theiliad (changes requested):

packages/core/src/services/angle-utils.ts
Akshat55: I don't think it really matters, if either x or y equals NaN it will render at position 0. But I'll add separate conditions for both x and y.

packages/core/src/services/angle-utils.ts
```diff
@@ -80,16 +80,12 @@ export function polarToCartesianCoords(
 	r: number,
 	t: Point = { x: 0, y: 0 }
 ): Point {
-	const x = r * Math.cos(a) + t.x;
-	const y = r * Math.sin(a) + t.y;
+	let x = r * Math.cos(a) + t.x;
+	let y = r * Math.sin(a) + t.y;
```
theiliad, line 84: why would x & y not be `const`s?

theiliad (changes requested):

packages/core/src/services/angle-utils.ts
Akshat55: Missed that during the previous fix, I have just changed the variable declarations to `const`.

theiliad (approved):

PR #1222, opened 2021-11-20
feat(react diagrams): add support for applying a custom class to the outer node

### LINKED ISSUE #1221: [Enhancement]: Add support for specifying a custom class name on CardNode
### Contact Details

_No response_

### Summary

Add support for specifying a custom class name on the CardNode react component which can then be used to control the colour of the border or other styles as needed.

### Justification

The CardNode react component has a `color` prop which can be used to specify the colour of the border applied on the outer node. However, this requires the consuming code to be aware of the active theme, handle dark mode, etc. to select an appropriate colour in some cases.

For component libraries this can be problematic as now both the components and their consumers need to be aware of theme rather than relying on this being handled entirely in (S)CSS.

Consumers of the CardNode component can add a custom class to the outer node and use this to style the card and its children.

### "Must have" functionality

n/a

n/a

### Available extra resources

I can implement this if it's an acceptable change.

AlanGreene:
### Updates
The CardNode component has a `color` prop which can be used to specify the
colour of the border applied on the outer node. However, this requires the
code to be aware of the active theme, handle dark mode, etc. to select an
appropriate colour in some cases.

For component libraries this can be problematic as now both the components
and their consumers need to be aware of theme rather than relying on this
being handled entirely in (S)CSS.

Add support for applying a custom class to the CardNode component which will
be applied to the outer node and can be used to apply styles to the card and
its children, including specifying the border colour for the card.

resolves #1221

Add similar `className` prop to all diagram components for consistency.

- [ ] Demos all features
- [ ] Documented/annotated
- [ ] Matches UI/UX specs
- [ ] Accessible

### DIFF
--- a/packages/react/src/diagrams/CardNode/CardNode.tsx
+++ b/packages/react/src/diagrams/CardNode/CardNode.tsx
@@ -13,6 +13,7 @@ const namespace = `${prefix}--cc--card-node`;
 const CardNode = ({
 	as = 'div',
 	children,
+	className,
 	color,
 	href = null,
 	onMouseEnter = null,
@@ -37,6 +38,7 @@ const CardNode = ({
 	const cardClasses = classnames(namespace, {
 		[`${namespace}--stacked`]: stacked,
 		[`${namespace}--${Component}`]: Component,
+		[className]: className,
 	});
 
 	return (
@@ -66,6 +68,11 @@ CardNode.propTypes = {
 	 * Pass in the children that will be rendered within the CardNode
 	 */
 	children: PropTypes.node,
+
+	/**
+	 * Provide an optional class to be applied on the outer element
+	 */
+	className: PropTypes.string,
 
 	/**
 	 * Specify the node's border color
--- a/packages/react/src/diagrams/CardNode/CardNodeColumn.tsx
+++ b/packages/react/src/diagrams/CardNode/CardNodeColumn.tsx
@@ -2,19 +2,22 @@ import React from 'react';
 import PropTypes from 'prop-types';
 
 // @ts-ignore
+import classnames from 'classnames';
+
+// @ts-ignore
 import settings from 'carbon-components/src/globals/js/settings';
 
 const { prefix } = settings;
 const namespace = `${prefix}--cc--card-node`;
 
-const CardNodeColumn = ({ children, farsideColumn }: any) => (
-	<div
-		className={`${namespace}__column ${
-			farsideColumn && `${namespace}__column--farside`
-		}`}>
-		{children}
-	</div>
-);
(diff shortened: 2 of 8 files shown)


packages/react/src/diagrams/CardNode/CardNode.tsx
```diff
@@ -67,6 +69,11 @@ CardNode.propTypes = {
 	 */
 	children: PropTypes.node,
 
+	/**
+	 * Provide an optional class to be applied on the outer element
+	 */
+	className: PropTypes.string,
```
theiliad, line 75: this could be misleading. Seems like it should be called `additionalClassNames`? and why just 1 additional className? maybe a user would like to add multiple

theiliad (changes requested):

packages/react/src/diagrams/CardNode/CardNode.tsx
AlanGreene: The `className` attribute on a DOM element accepts multiple classes and this naming/use is consistent with carbon-components-react.

For example I could have `<CardNode className="my-custom-class another-class" … />` and it works just the same as a single class.

AlanGreene: @theiliad what do you think? Is it OK as-is?

I would like to get this wrapped up soon so we can start using it.

theiliad: @AlanGreene I guess so. My only concern now is, what about the rest of the diagram components?

This'll kind of be an inconsistency if we add the classname to just 1 component.

What do you think?

AlanGreene: I don't have a particular use case in mind for it on any of the other diagram components but would be happy to add it if you think it would be useful.

theiliad (changes requested):
Thank you @AlanGreene looks good, however I think in 1 of the instances you've added the prop but just not applying it.

packages/react/src/diagrams/CardNode/CardNodeColumn.tsx
```diff
@@ -1,20 +1,22 @@
...
-		className={`${namespace}__column ${
-			farsideColumn && `${namespace}__column--farside`
-		}`}>
-		{children}
-	</div>
-);
+const CardNodeColumn = ({ children, className, farsideColumn }: any) => {
+	const classes = classnames(`${namespace}__column`, {
+		[`${namespace}__column--farside`]: farsideColumn,
+	});
```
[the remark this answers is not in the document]
AlanGreene: good catch, updated it now. Thanks @theiliad

theiliad (approved):
SGTM 💯

Akshat55 (approved):

PR #1233, opened 2021-12-01
fix(core): display the correct group when hovering over scatter point

### LINKED ISSUE #1188: [Bug]: Stack Area chart tooltip doesn't display the correct group as selected in the legend
### Contact Details

bikash.pattnayak@gmail.com

### What happened?

Tooltip does not show the correct group. This issue i can reproduce in the storybook demo as well.
https://charts.carbondesignsystem.com/angular/?path=/story/simple-charts-area-stacked--stacked-area-time-series-with-uneven-data

Just select any group from the legend and hover over the chart to see the tooltip. The tooltip group displayed doesn't match the legend selected.
image

### Version

0.50.7. 
This  can reproduced in any older version as well.

### Data & options used

_No response_

### Relevant log output

_No response_

### Codesandbox example

_No response_

Akshat55:
### Updates
- Render the scatter points only if the status of the data group is true.

fix #1188

image

- [ ] Demos all features
- [ ] Documented/annotated
- [ ] Matches UI/UX specs
- [ ] Accessible

### DIFF
--- a/packages/core/src/components/graphs/scatter-stacked.ts
+++ b/packages/core/src/components/graphs/scatter-stacked.ts
@@ -30,6 +30,9 @@ export class StackedScatter extends Scatter {
 			percentage,
 		});
 
+		// Organize the data based on the active data groups
+		const dGroups = this.model.getDataGroups();
+
 		// Update data on dot groups
 		const circleGroups = svg
 			.selectAll('g.dots')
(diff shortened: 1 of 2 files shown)


packages/core/src/components/graphs/scatter-stacked.ts
```diff
@@ -30,6 +30,9 @@ export class StackedScatter extends Scatter {
 			percentage,
 		});
 
+		// Organize the data based on the active data groups
+		const dGroups = this.model.getDataGroups();
```
theiliad, line 34:
```suggestion
		const dataGroups = this.model.getDataGroups();
```

theiliad: let's just use full names

theiliad (changes requested):

theiliad (approved):

theiliad (approved):

Issue #1232, opened 2021-11-30
[Bug]: StackedBarChart bar width depends on number of data series


piotr-kotara-ibm:
### Contact Details

piotr.kotara@ibm.com

### What happened?

The default width of the bar in the chart depends on the length of the data:
https://github.com/carbon-design-system/carbon-charts/blob/b11873b24c3ca19ef07381b1647be70be33b4baf/packages/core/src/components/graphs/bar.ts#L13
https://github.com/carbon-design-system/carbon-charts/blob/b11873b24c3ca19ef07381b1647be70be33b4baf/packages/core/src/components/graphs/bar.ts#L19-L24

However, when there are multiple data groups defined, the width of the bars is smaller the greater the number of data series is, eg:
One data serie:
<img width="1100" alt="obraz" src="https://user-images.githubusercontent.com/92912677/144094933-e62b760e-99d1-4158-a855-c28181162e84.png">

Same chart, but 3 data series:
<img width="1100" alt="obraz" src="https://user-images.githubusercontent.com/92912677/144095219-21822bca-9678-4d26-866d-50776cfb1cd8.png">

Example in https://codesandbox.io/s/sparkling-microservice-l2w7t?file=/src/index.js:455-3626

<img width="682" alt="obraz" src="https://user-images.githubusercontent.com/92912677/144095923-4301b300-a761-4168-9c77-d789cf326154.png">
<img width="682" alt="obraz" src="https://user-images.githubusercontent.com/92912677/144095970-a5bf05a0-bef3-4a7c-a737-d5a78a31063a.png">

### Version

`@carbon/charts@0.52.3`

### Codesandbox example

https://codesandbox.io/s/sparkling-microservice-l2w7t?file=/src/index.js:455-3626

piotr-kotara-ibm:
I have came up with the solution for the problem which I've included in the #1235 pull request. Hope it helps. Please feel free to contact me if you will have any questions.

piotr-kotara-ibm:
Fixed by https://github.com/carbon-design-system/carbon-charts/pull/1235

PR #1235, opened 2021-12-02
fix(stacked-bar): set bar width based on number of domain values

piotr-kotara-ibm:
This fixes the #1232 issue:

The solution is to override the default method for getting width of the bar width for the  `StackedBar` :
https://github.com/carbon-design-system/carbon-charts/blob/b11873b24c3ca19ef07381b1647be70be33b4baf/packages/core/src/components/graphs/bar.ts#L7-L27

with the analogous method which uses `this.model.getStackKeys().length;` as the count of data points on domain we divide in line 22 of default method: https://github.com/carbon-design-system/carbon-charts/blob/b11873b24c3ca19ef07381b1647be70be33b4baf/packages/core/src/components/graphs/bar.ts#L22
instead of using `this.model.getDisplayData().length;` which just gets length of whole visible data, thus for charts with many series, the greater the number of series is, the thinner the bars are.

before fix:
obraz

after fix:
obraz

for the single serie chart we see lack of difference:

before fix:
obraz

after fix:
obraz

- [ ] Demos all features
- [ ] Documented/annotated
- [ ] Matches UI/UX specs
- [ ] Accessible

### DIFF
--- a/packages/core/src/components/graphs/bar-stacked.ts
+++ b/packages/core/src/components/graphs/bar-stacked.ts
@@ -304,7 +304,7 @@ export class StackedBar extends Bar {
 			useAttrs: true,
 		}).width;
 	
-		const numberOfDomainValues = this.model.getStackKeys({}).length;
+		const numberOfDomainValues = this.model.getStackKeys().length;
 	
 		if (!mainXScale.step) {
 			return Math.min(
--- a/packages/core/src/components/graphs/bar-stacked.ts
+++ b/packages/core/src/components/graphs/bar-stacked.ts
@@ -8,6 +8,8 @@ import {
 	ColorClassNameTypes,
 	RenderTypes,
 } from '../../interfaces';
+import { DOMUtils } from '../../services';
+
 
 // D3 Imports
 import { select } from 'd3-selection';
@@ -292,6 +294,27 @@ export class StackedBar extends Bar {
 			});
 	}
 
+	protected getBarWidth() {
+		const options = this.getOptions();
+		if (options.bars.width) {
+			return options.bars.width;
+		}
+		const mainXScale = this.services.cartesianScales.getMainXScale();
+		const chartWidth = DOMUtils.getSVGElementSize(this.parent, {
+			useAttrs: true,
+		}).width;
+	
+		const numberOfDomainValues = this.model.getStackKeys().length;
+	
+		if (!mainXScale.step) {
+			return Math.min(
+				options.bars.maxWidth,
+				(chartWidth * 0.25) / numberOfDomainValues
+			);
+		}
+		return Math.min(options.bars.maxWidth, mainXScale.step() / 2);
+	}
+	
 	destroy() {
 		// Remove event listeners
 		this.parent
--- a/packages/core/src/components/graphs/bar-stacked.ts
+++ b/packages/core/src/components/graphs/bar-stacked.ts
@@ -8,6 +8,8 @@ import {
 	ColorClassNameTypes,
 	RenderTypes,
 } from '../../interfaces';
+import { DOMUtils } from '../../services';
+
 
 // D3 Imports
 import { select } from 'd3-selection';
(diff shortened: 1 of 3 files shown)


packages/core/src/components/graphs/bar-stacked.ts
```diff
@@ -292,6 +294,27 @@ export class StackedBar extends Bar {
 			});
 	}
 
+	protected getBarWidth() {
+		const options = this.getOptions();
+		if (options.bars.width) {
```
theiliad, line 299:
```suggestion
		if (Tools.getProperty(options, "bars", "width")) {
```

theiliad (changes requested):
Hi @piotr-kotara-ibm 👋🏻 

Thanks for making this PR. Seems like the stacked bar demos are broken now in the deploy preview 
image

piotr-kotara-ibm:
@theiliad sorry I have missclicked te re-request review now - the error I see now after applying suggested changes is as follows:
obraz

I will try to address it. Do you have any idea what might have caused it? Thank you for help in advance

piotr-kotara-ibm:
@theiliad I have addressed error - it was caused by wrong use of ` this.model.getStackKeys({})` function. I didn't pass any arg for it but it requires one argument even if one wants to pass the null in both  `bins` and  `groups` params, so the solution was to change `this.model.getStackKeys()` to `this.model.getStackKeys({})`.

Deploy previews are now working - I would be grateful if you could review this PR once again.

packages/core/src/components/graphs/bar-stacked.ts
```diff
@@ -292,6 +294,27 @@ export class StackedBar extends Bar {
...
+		const options = this.getOptions();
+		if (Tools.getProperty(options, "bars", "width")) {
+			return options.bars.width;
+		}
+		const mainXScale = this.services.cartesianScales.getMainXScale();
+		const chartWidth = DOMUtils.getSVGElementSize(this.parent, {
+			useAttrs: true,
+		}).width;
+	
+		const numberOfDomainValues = this.model.getStackKeys({}).length;
```
theiliad, line 307: I'm not a big fan of this...

We can change `getStackKeys()` a bit to fix this issue

image


```ts
	getStackKeys(
		{ bins = null, groups = null } = { bins: null, groups: null }
	) 
```

theiliad (changes requested):

piotr-kotara-ibm:
Hi @theiliad , Thank you for your proposition - I have applied it to the PR. Could you look at it one more time?

packages/core/src/components/graphs/bar-stacked.ts
```diff
@@ -292,6 +294,27 @@ export class StackedBar extends Bar {
...
+	
+		if (!mainXScale.step) {
+			return Math.min(
+				options.bars.maxWidth,
+				(chartWidth * 0.25) / numberOfDomainValues
+			);
+		}
+		return Math.min(options.bars.maxWidth, mainXScale.step() / 2);
+	}
+
```
theiliad, line 317: this seems to exist inside `bar.ts`
https://github.com/carbon-design-system/carbon-charts/blob/c600eafe6996eb1678f56850abdfbb907edb75d5/packages/core/src/components/graphs/bar.ts#L7-L27

We probably need this bit removed and just need the addition to `model.ts`

theiliad (changes requested):

packages/core/src/components/graphs/bar-stacked.ts
piotr-kotara-ibm: Hi @theiliad, thanks for feedback. I have looked more closely into the `getStackKeys()` function.

https://github.com/carbon-design-system/carbon-charts/blob/fe4d592991e9cbf68c4b9efab0be4aa46dfcd416/packages/core/src/model/model.ts#L302-L344

As we see it bases on `getDisplayData()` function and just filters out the duplicates and sorts the data effectively giving us the number of bars, and that's the data we want in general in this case, not the number of data entries.

Since we use in `bar.ts` `getDisplayData()` data count, the use of `getStackKeys()` by default in `bar.ts` will only cause to filter out the duplicates, thus not impacting simple bar chart and fixing the issue in stacked bar charts.

We do not have a redundancy this way and we are just passing right data (so the number of bars instead of number of data entries) to determine the bar width

packages/core/src/components/graphs/bar.ts
```diff
@@ -10,7 +10,7 @@ export class Bar extends Component {
 			return options.bars.width;
 		}
 
-		const numberOfDatapoints = this.model.getDisplayData().length;
```
theiliad, line 13: sorry...!

I actually made a mistake here.

We should keep the override rather than just updating the `bar.ts` code. I think once we revert that piece we're good to merge.

theiliad (changes requested):

packages/core/src/components/graphs/bar.ts
piotr-kotara-ibm: hi @theiliad, I have reverted last commit

theiliad (approved):

PR #1243, opened 2021-12-13
fix(core): use domain in zoom bar if defined in axes

### LINKED ISSUE #1228: [Bug]: Enabling zoombar ignores the set time domain in scatter chart
### Contact Details

nsandhu@us.ibm.com

### What happened?

When I set the x-axis time domain and enable zoombar, the time domain is ignored. The x-axis range is set based on min/max values in data instead.

This started happening after I upgraded the carbon-charts version from `0.41.98-d3v5-latest-0.1` to `0.52.3`.

### Version

`@carbon/charts@0.52.3`

### Data & options used

_No response_

### Relevant log output

_No response_

### Codesandbox example

https://codesandbox.io/s/funny-lehmann-ezu8u

Akshat55:
### Updates
- Set zoom bar domain to the initial domain defined in axes
- User can use initialDomain in zoom bar options to zoom in into min/max value. T
  - User must calculate extent

fix #1228 

image

- [ ] Demos all features
- [ ] Documented/annotated
- [ ] Matches UI/UX specs
- [ ] Accessible

### DIFF
--- a/packages/core/src/services/zoom.ts
+++ b/packages/core/src/services/zoom.ts
@@ -67,11 +67,21 @@ export class Zoom extends Service {
 		const mainXAxisPosition = cartesianScales.getMainXAxisPosition();
 		const domainIdentifier = cartesianScales.getDomainIdentifier();
 
-		// default to full range with extended domain
-		return cartesianScales.extendsDomain(
+		const customDomain = Tools.getProperty(
+			this.model.getOptions(),
+			'axes',
 			mainXAxisPosition,
-			extent(allZoomBarData, (d: any) => d[domainIdentifier])
+			'domain'
 		);
+
+		// Return custom domain if defined
+		// default to full range with extended domain
+		return customDomain
+			? customDomain
+			: cartesianScales.extendsDomain(
+					mainXAxisPosition,
+					extent(allZoomBarData, (d: any) => d[domainIdentifier])
+			  );
 	}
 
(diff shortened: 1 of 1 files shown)


packages/core/src/services/zoom.ts
```diff
@@ -67,11 +67,21 @@ export class Zoom extends Service {
...
 		);
+
+		// Return custom domain if defined
+		// default to full range with extended domain
+		return customDomain
+			? customDomain
+			: cartesianScales.extendsDomain(
+					mainXAxisPosition,
+					extent(allZoomBarData, (d: any) => d[domainIdentifier])
+			  );
```
theiliad, line 84: we should probably do some checks here and make sure the custom domain is correctly formatted... e.g. the provided custom domain isn't a string, or an int, but an array of size 2...?

theiliad: this should probably just be a function somewhere that we can re-use since we pull in the custom domains in multiple places

theiliad (changes requested):

packages/core/src/services/zoom.ts
Akshat55: > this should probably just be a function somewhere that we can re-use since we pull in the custom domains in multiple places

From my findings, custom domain is only retrieved twice, one in `zoom.ts` & second in `axis.ts`
```ts
Tools.getProperty(
       options,
       'axes',
       axisPosition,
       'domain'
);
```

theiliad (approved):

PR #1249, opened 2021-12-17
fix(core): keep radar chart alignment after fullscreen exit

### LINKED ISSUE #1247: [Bug]: Radar chart is no longer aligned after exiting full screen
### Contact Details

_No response_

### What happened?

Radar alignment is incorrect after exiting full screen. This behavior is noticeable when initial alignment is set to "center" or "right". 

image

### Version

Latest

### Data & options used

```js
...
  "radar": {
    "axes": {
      "angle": "feature",
      "value": "score"
    },
    "alignment": "center" || "right"
  },
...
```

### Relevant log output

_No response_

### Codesandbox example

_No response_

Akshat55:
### Updates
- Align the chart after the transitions are complete, previously alignment was triggered async which caused it to be misaligned after fullscreen exit

fix #1247

- [ ] Demos all features
- [ ] Documented/annotated
- [ ] Matches UI/UX specs
- [ ] Accessible

### DIFF
--- a/packages/core/src/components/graphs/radar.ts
+++ b/packages/core/src/components/graphs/radar.ts
@@ -444,6 +444,22 @@ export class Radar extends Component {
 									c
 								).y
 						)
+						.end()
+						.finally(() => {
+							// Align chart horizontally after x-axies has finished rendering
+							const alignment = Tools.getProperty(
+								options,
+								'radar',
+								'alignment'
+							);
+
+							const alignmentXOffset = this.getAlignmentXOffset(
+								alignment,
+								svg,
+								this.getParent()
+							);
+							svg.attr('x', alignmentXOffset);
+						})
 				),
 			(exit) =>
 				exit.call((selection) =>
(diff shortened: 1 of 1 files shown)


packages/core/src/components/graphs/radar.ts
```diff
@@ -444,6 +444,22 @@ export class Radar extends Component {
...
+								'alignment'
+							);
+
+							const alignmentXOffset = this.getAlignmentXOffset(
+								alignment,
+								svg,
+								this.getParent()
+							);
+							svg.attr('x', alignmentXOffset);
+						})
```
theiliad, line 462: this is running after .update(), what about .enter()?

theiliad (changes requested):

packages/core/src/components/graphs/radar.ts
Akshat55: .enter() will be called only once - when the radar chart is being created. Hence, it makes more sense to add it to update since the element exists and the position can be realigned when any changes are made.

I cannot chain the promise to the .call function. If I chain it inside the call function, then the alignment will be incorrect on fullscreen exit. This is the same case if I were to add the above code to **both** enter and update.

theiliad (approved):

PR #1259, opened 2022-01-04
fix(core): add background to chart in fullscreen mode

theiliad:


Akshat55:
Is this PR also updating carbon styles to a newer version? 

Seems like adding
```
&.filled, &.fullscreen {
```
 in `packages/core/src/styles/_chart-holder.scss` file fixes the full screen background issue.

PR #1263, opened 2022-01-07
fix(core): prevent area chart from bleeding out of cartesian plane when includeZero is enabled

### LINKED ISSUE #1248: [Bug]: Area chart bleeds out of 'area container' when includeZero is set to false
### Contact Details

_No response_

### What happened?

Include Zero set to false results in area (simple) charts to cover bottom axes as shown in image below.

image

### Version

Latest

### Data & options used

```js
http://localhost:9006/?path=/story/simple-charts-area-simple--area-discrete-domain

...
 "left": {
      "mapsTo": "value",
      "title": "Conversion rate",
      "scaleType": "linear",
      "includeZero": false
    }
...
```

### Relevant log output

_No response_

### Codesandbox example

_No response_

Akshat55:
### Updates
- Added a function that updates the upperBound range value if includeZero is enabled for a specific axes
   - Does not affect percentage or stacked
- Removed unused import (d3-array sum) in Cartesian service

fix #1248

image

- [ ] Demos all features
- [ ] Documented/annotated
- [ ] Matches UI/UX specs
- [ ] Accessible

### DIFF
--- a/packages/core/src/components/graphs/area.ts
+++ b/packages/core/src/components/graphs/area.ts
@@ -68,7 +68,7 @@ export class Area extends Component {
 		}
 
 		let upperBoundRangeValue = 0;
-		// If includeZero is enabled, we want to replace upperBooundRange from 0 to domain value
+		// If includeZero is enabled, we want to replace upperBoundRange from 0 to domain value
 		const includeZeroInRangeValue = (position, domain) => {
 			if (
 				Tools.getProperty(options, 'axes', position, 'includeZero') ===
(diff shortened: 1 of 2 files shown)


packages/core/src/components/graphs/area.ts
```diff
@@ -66,24 +67,48 @@ export class Area extends Component {
 			); // eslint-disable-line no-console
 		}
 
+		let upperBoundRangeValue = 0;
+		// If includeZero is enabled, we want to replace upperBooundRange from 0 to domain value
```
theiliad, line 71:
```suggestion
		// If includeZero is enabled, we want to replace upperBoundRange from 0 to domain value
```

theiliad (changes requested):

theiliad (approved):

PR #1273, opened 2022-01-17
fix(core): render only the components that are being used in proportional meter

### LINKED ISSUE #1262: [Bug]: Meter Chart - Height options behave unpredictable
### Contact Details

steffen.baumann@de.ibm.com

### What happened?

**Description:**
Setting a Meters height is unpredictable due to the height applying to the overall size, but the actual meter (band) is just a portion of the overall size. The complete height seems divided into:
- Chart Title + Toolbar (bx--cc--title)
- Spacer (bx--cc--spacer)
- MeterTitle + Breakdown + Total (meter-title)  // actually including
- Spacer (bx--cc--spacer)
- Chart (bx--cc--meter)
- Spacer (bx--cc--spacer)
- Legend (bx--cc--legend)

Interestingly, removing the title from the options and disabling the toolbar, removes the DOM element and its trailing Spacer. It is similar for the legend, by removing the leading Spacer.
However, avoiding a MeterTitle (not set in proportional Meters at all but still in the DOM) and hiding total as well as breakdown values (through the Formatters), still keeps the container and Spacer as seen in the below screen capture:
image

Even worse, the `options.meter.height` value is a number, applied to the SVG components in `bx--cc--meter`. However, `options.height` is divided by the available components. That said, the following configuration:
```
height: 200px
toolbar: {
	enabled: false
},
meter: {
	height: 200,
	proportional: {
		breakdownFormatter: (value) => "",
		totalFormatter: (value) => ""
	}
},
legend:{
	enabled: false,
}
```
results in this UI:
image
in which the actual meter (band) is only 66.66px and remaining content is overflowing:
image

**Expectation(s):**
If neither `options.meter.title`, nor `breakdown` and `total` (formatter return a falsify value) for proportional Meters are defined, the DOM element with CSS class `meter-title` should not be rendered/added at all. Additionally, the trailing Spacer needs to be removed.
Additionally, the Spacer trailing the `meter-title` should not be growing proportional. This leads to weird designs as shown by the below screen capture:
image
Finally, it might be reasonable to provide some sample chart with customized meter/band height. This allows developers to better understand the impacts and how that values interacts with the chart's height option. 

**Justification:**
There might have been a decision, Total and Breakdown values are mandatory for accessibility of the Chart. Nonetheless, the formatter allow plain text usage only. Any dedicated (HTML) styling isn't possible with those details. As a result, removing them and going custom is required. But exactly that isn't much feasible, as empty contents with certain heights are overlaying details. Additionally, the calculation of the band height is depending of the values.
This said, as long as those Formatters do not allow custom HTML, the behavior needs to be adjusted. Additionally, 

### Version

reproduced with `@carbon/charts@0.54.0`

### Data & options used

```js
const data = [{
	group: "emails",
	value: 202
},
{
	group: "photos",
	value: 654
},
{
	group: "text messages",
	value: 723
},
{
	group: "other",
	value: 120
}];

const options = {
	// title: "test",
	height: "200px",
	toolbar: {
		enabled: false
	},
	meter: {
		height: 200,
		proportional: {
			breakdownFormatter: (value) => "",
			totalFormatter: (value) => ""
		}
	},
	legend:{
		enabled: false,
	}
};
```

### Relevant log output

_No response_

### Codesandbox example

https://codesandbox.io/s/relaxed-heyrovsky-ih5pj

Akshat55:
### Updates
- Add condition to render only the components that are used in proportional meter

**Might be considered a breaking change for the users using formatters?**

fix #1262

image

*Can not be tested in storybook since we are working with functions here. You can see in the image above we are not rendering the total,breakdown or the spacer since they are not being used.

- [ ] Demos all features
- [ ] Documented/annotated
- [ ] Matches UI/UX specs
- [ ] Accessible

### DIFF
--- a/packages/core/src/charts/meter.ts
+++ b/packages/core/src/charts/meter.ts
@@ -49,39 +49,61 @@ export class MeterChart extends Chart {
 	}
 
 	getComponents() {
-		// Specify what to render inside the graph only
-		const graph = {
-			id: 'meter-graph',
-			components: [new Meter(this.model, this.services)],
-			growth: LayoutGrowth.STRETCH,
-			renderType: RenderTypes.SVG,
-		};
+		const options = this.model.getOptions();
+		const meterComponents = [
+			// Meter has a unique dataset title within the graph
+			{
+				id: 'meter-title',
+				components: [new MeterTitle(this.model, this.services)],
+				growth: LayoutGrowth.STRETCH,
+				renderType: RenderTypes.SVG,
+			},
+			// Create the title spacer
+			{
+				id: 'spacer',
+				components: [
+					new Spacer(this.model, this.services, { size: 8 }),
+				],
+				growth: LayoutGrowth.STRETCH,
+			},
+			// Specify what to render inside the graph only
+			{
+				id: 'meter-graph',
+				components: [new Meter(this.model, this.services)],
+				growth: LayoutGrowth.STRETCH,
+				renderType: RenderTypes.SVG,
+			},
+		];
 
-		// Meter has an unique dataset title within the graph
-		const titleComponent = {
-			id: 'meter-title',
-			components: [new MeterTitle(this.model, this.services)],
-			growth: LayoutGrowth.STRETCH,
-			renderType: RenderTypes.SVG,
-		};
+		//breakdownFormatter
+		const breakdownFormatter = Tools.getProperty(
+			options,
+			'meter',
+			'proportional',
+			'breakdownFormatter'
+		);
 
-		// create the title spacer
-		const titleSpacerComponent = {
-			id: 'spacer',
-			components: [new Spacer(this.model, this.services, { size: 8 })],
-			growth: LayoutGrowth.STRETCH,
-		};
+		// totalFormatter function
+		const totalFormatter = Tools.getProperty(
+			options,
+			'meter',
+			'proportional',
+			'totalFormatter'
+		);
+
+		// Check to see if the formatter functions exist
+		if (!!breakdownFormatter && !!totalFormatter) {
+			// Check if both formatters return null or empty string, do not render them
+			if (!breakdownFormatter(null) && !totalFormatter(null)) {
+				meterComponents.splice(0, 2);
+			}
+		}
 
 		// the graph frame for meter includes the custom title (and spacer)
 		const graphFrame = [
-			new LayoutComponent(
-				this.model,
-				this.services,
-				[titleComponent, titleSpacerComponent, graph],
-				{
-					direction: LayoutDirection.COLUMN,
-				}
-			),
+			new LayoutComponent(this.model, this.services, meterComponents, {
+				direction: LayoutDirection.COLUMN,
+			}),
 		];
 
 		// add the meter title as a top level component


packages/core/src/charts/meter.ts
```diff
@@ -49,39 +49,61 @@ export class MeterChart extends Chart {
...
-		// create the title spacer
-		const titleSpacerComponent = {
-			id: 'spacer',
-			components: [new Spacer(this.model, this.services, { size: 8 })],
-			growth: LayoutGrowth.STRETCH,
-		};
+		// Check to see if the formatter functions exist
+		if (!!breakdownFormatter && !!totalFormatter) {
+			// Check if both formatters return null or empty string, do not render them
+			if (!breakdownFormatter(null) && !totalFormatter(null)) {
```
theiliad, line 97: should we pass `null` in here? sounds like that could cause errors on the user's side, e.g. if they assume they'll get a string and try to do something like `str.replace()`...

theiliad (changes requested):

packages/core/src/charts/meter.ts
Akshat55: You're right, for that reason I have decided that it would make more sense to simply add an option to proportional meter chart. This would do two things:

1. Would not cause a breaking change for users that are returning null value from total and breakdown formatters (Previously, it would require them to perhaps make css adjustments since the padding would be gone)
2. Less computation & no chance of errors like previously

```js
proportional?: {
    ...
    renderTitle?: boolean;
};
```

theiliad: does `renderTitle` just remove the chart title? If so, we should add it to all the charts, correct?

theiliad: e.g. `{ title: { visible: false } }` or something like that

Akshat55:
An updated image in storybook:
image

We probably may not want to create a demo since it may incentivize teams to do this. The total and breakdown is useful in terms of accessibility but this PR will make this chart more consistent with others in terms of not rendering things that are not being used.

packages/core/src/charts/meter.ts
Akshat55: > does `renderTitle` just remove the chart title? If so, we should add it to all the charts, correct?

Yes, it removes the Meter Title (Total and breakdown) specifically not the chart title.

> does `renderTitle` just remove the chart title? If so, we should add it to all the charts, correct?
> e.g. `{ title: { visible: false } }` or something like that

Currently, there is a condition in code that renders the title only if it exists in `chart.ts`:

```ts
		if (titleAvailable || toolbarEnabled) {

// Line below adds the header (title) to the layout only if the title value exists. If you don't pass in a truthy title, 
// it will not render
			topLevelLayoutComponents.push(headerComponent);

			const titleSpacerComponent = {
				id: 'spacer',
				components: [
					new Spacer(
						this.model,
						this.services,
						toolbarEnabled ? { size: 15 } : undefined
					),
				],
				growth: LayoutGrowth.PREFERRED,
			};

			topLevelLayoutComponents.push(titleSpacerComponent);
		}

```

theiliad (changes requested):
image

When I disable `showHeader`, there's a gap between the legend & the meter graph. What are we expecting to happen here?

Also what about calling the flag `showLabels` or something of that effect?

Akshat55:
> image
> 
> When I disable `showHeader`, there's a gap between the legend & the meter graph. What are we expecting to happen here?
> 
> Also what about calling the flag `showLabels` or something of that effect?



This behavior is consistent with the existing meter behavior, for example increasing height of proportional meter from 130px to 250px will increase the spacing between the chart because the element height scales off height.

image

The height is divided into three, so the spacers including the chart meter chart itself scales off the height. In the image I just sent for height of 200px, Each element has a height of 47.667px. So when we remove the labels (previously called header), it will add extra spacing between the legend and the chart.

theiliad (approved):

theiliad:
Makes sense. Let's follow up with the height discussion in future PRs

PR #1324, opened 2022-03-09
docs(carbon-platform): index charts with schema

mattrosno:
### Updates
- Indexes these libraries and assets using the schema (https://github.com/carbon-design-system/carbon-platform/blob/main/docs/resource-schemas.md)
  - [x] Vanilla
  - [x] Angular
  - [x] React
  - [x] Svelte
  - [x] Vue 

N/A

### To do
- [x] Need thumbnails for
  - [x] Combo
  - [x] Sparkline
  - [x] Step
- [x] Need tags for (https://github.com/carbon-design-system/carbon-platform/blob/main/docs/resource-schemas.md#data-visualization-tags)
  - [x] ~~Combo~~ (we don't need an additional tag for every asset if there's not a fit)
  - [x] Sparkline
  - [x] Step
- [x] Should `Meter (proportional)` be indexed? I see in as top-level in Storybook, but not listed in legacy website index. If so, we need a thumbnail for it.

### To do some time later

- [ ] Add descriptions to every component


packages/core/carbon.yml
```diff
@@ -0,0 +1,527 @@
...
+      - geographic-overlay
+    framework: design-only
+    platform: web
+    thumbnailPath: /src/thumbnails/geo-choropleth.svg
+    externalDocsUrl: https://www.carbondesignsystem.com/data-visualization/complex-charts#proportional-symbol-map
+  heatmap:
+    name: Heat map
+    status:
+      key: experimental
+      note:
```
mattrosno, line 274: @theiliad is Heat map experimental? I saw this note in Storybook.

packages/core/carbon.yml
```diff
@@ -0,0 +1,527 @@
...
+    name: Stream
+    status: draft
+    type: component
+    tags:
+      - data-visualization
+      - trend
+    framework: design-only
+    platform: web
+    thumbnailPath: /src/thumbnails/stream.svg
+    noIndex: true
```
mattrosno, line 479: I put Steam as `noIndex` so it doesn't show up in the catalogs. It looks grayed out in the legacy site with no click target. This right or wrong?

packages/core/carbon.yml
theiliad: Yes we're still making adjustments to it

packages/core/carbon.yml
theiliad: yeah better to exclude for now

packages/core/carbon.yml
mattrosno: Is this experimental for all? Or just vanilla and stable for others?

packages/angular/carbon.yml
```diff
@@ -0,0 +1,275 @@
...
+    externalDocsUrl: https://www.carbondesignsystem.com/data-visualization/complex-charts#circle-packs
+    demoLinks:
+      - type: storybook
+        name: Storybook
+        action: link
+        url: https://charts.carbondesignsystem.com/angular/?path=/story/complex-charts-circle-pack
+  combo:
+    status: stable
+    framework: angular
+    externalDocsUrl: https://www.carbondesignsystem.com/data-visualization/getting-started
```
theiliad, line 116: oh wow that's a gap, but yeah probably the best link for now

theiliad (approved):
looks good at a quick glance, mainly checked the Angular configs, assuming the rest are almost duplicates

theiliad:
> * [ ]  Do we have descriptions for each chart?

no not yet

> * [ ]  Need tags for (https://github.com/carbon-design-system/carbon-platform/blob/main/docs/resource-schemas.md#data-visualization-tags)
>   * [ ]  Combo

Probably need a new tag

>   * [ ]  Sparkline

trend

>   * [ ]  Step

I think trends

> * [ ]  Should `Meter (proportional)` be indexed? I see in as top-level in Storybook, but not listed in legacy website index. If so, we need a thumbnail for it.

I think so.

theiliad (changes requested):
could we move `thumbnails` to the root pls?

jeanservaas:
Here are thumbs for:
– step
- combo
- prop. meter
- sparkline

[new_charts.zip](https://github.com/carbon-design-system/carbon-charts/files/8216971/new_charts.zip)

packages/angular/carbon.yml
mattrosno: Ya, I don't see any cds.com content for Combo, Sparkline, Step. In Carbon Next private beta, all assets need `externalDocsUrl`. So for now, we can point people to the cds.com data viz getting started. In a later platform release, we'll remove the `externalDocsUrl`s when we have detail pages populated.

theiliad (approved):
good at a quick glance. didn't check the thumbnails, assuming Jeannie did that

PR #1325, opened 2022-03-09
fix(core): enable heatmap value domain to support small values

### LINKED ISSUE #1321: [Bug]: Heatmap domain does not support small values (Including legend)
### What happened?

While implementing a heatmap chart that has very small values (0 to 1) I've noticed that the legend/color scaling seems  to be locked to 1-10 at a minimum. From what I can tell there is no way to override this behavior via the options object and it results in a heatmap where a majority of the cells are the lowest legend color:

image
 
I am hoping I'm missing a configuration or workaround as this unfortunately makes the heatmap un-usable for datasets with small values.

### Version

"@carbon/charts-svelte": "0.54.12"

### Data & options used

```js
const data = [
	  {
	    letter: "A",
	    month: "January",
	    value: 41
	  },
	  {
	    letter: "B",
	    month: "January",
	    value: 7
	  },
	  {
	    letter: "C",
	    month: "January",
	    value: 66
	  },
	  {
	    letter: "D",
	    month: "January",
	    value: 85
	  },
	 // continued in code sandbox
	].map(entry => ({
	  ...entry,
	  value: entry.value / 100
	}));

const options = {
	"title": "Heatmap (Quantize legend)",
	"axes": {
		"bottom": {
			"title": "Letters",
			"mapsTo": "letter",
			"scaleType": "labels"
		},
		"left": {
			"title": "Months",
			"mapsTo": "month",
			"scaleType": "labels"
		}
	},
	"heatmap": {
		"colorLegend": {
			"title": "Legend title",
			"type": "quantize"
		}
	},
	"experimental": true,
	"height": "400px"
}
```

### Relevant log output

_No response_

### Codesandbox example

https://codesandbox.io/s/epic-franklin-cv1nz0?file=/App.svelte

Akshat55:
### Updates
- Heatmap domain now supports small values (No user change required)

image

fix #1321

- [ ] Demos all features
- [ ] Documented/annotated
- [ ] Matches UI/UX specs
- [ ] Accessible

### DIFF
--- a/packages/core/src/model/heatmap.ts
+++ b/packages/core/src/model/heatmap.ts
@@ -5,7 +5,7 @@ import { Tools } from '../tools';
 
 // d3 imports
 import { extent } from 'd3-array';
-import { scaleQuantize } from 'd3-scale';
+import { scaleQuantize, scaleLinear } from 'd3-scale';
 
 /** The gauge chart model layer */
 export class HeatmapModel extends ChartModelCartesian {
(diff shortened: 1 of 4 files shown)


packages/core/src/model/heatmap.ts
```diff
@@ -48,29 +48,18 @@ export class HeatmapModel extends ChartModelCartesian {
...
-			domain.push(value);
-		});
+		const limits = extent(this.getDisplayData(), (d: any) => d.value);
+		const domain = scaleLinear()
+			.domain(limits as [number, number])
+			.nice()
+			.domain();
+
+		// Ensuring limits start at 0 to make scale look more `nicer`
+		if (domain[0] > 0) {
```
theiliad, line 58: is this correct? what if limits or domain include negative values?

theiliad (changes requested):
@Akshat55 do we have a demo to show this functionality?

Akshat55:
Added demo where we are using small numbers (0 - 10) and divergent color scheme

image

packages/core/src/model/heatmap.ts
Akshat55: Yes, this is the same as what I was doing previously:
```js
if (index === 0 && value >= 0) {
	value = 0;
}
```

If the number is positive, it'll set the minimum in the domain to 0 - Cartesian service does something similar but for axis

theiliad (approved):

PR #1327, opened 2022-03-09
docs: update thumbnail backgrounds

mattrosno:
This PR fixes the background of 6 thumbnail images (noticeable when the SVGs sit on a non-white background).

Also, it apply `noIndex` to vanilla Network Diagram so it doesn't do this. We want to only see Network Diagram for the two libraries where it's been implemented, Angular and React.

image


theiliad (approved):
As far as Jeannie has approved, should be ok

PR #1330, opened 2022-03-14
chore(core): update carbon telemetry version to 0.1.0

theiliad:


Akshat55 (approved):

jdharvey-ibm:
Did a local test of the most-recent version of charts together with the v0.1.0 version of telemetry and was able to run it successfully on my end.

PR #1379, opened 2022-05-26
update @carbon/styles references in sandbox

theiliad:


theiliad:
Will merge after https://github.com/carbon-design-system/carbon/pull/11495 is released

PR #1391, opened 2022-06-14
fix(core): add aria-label attribute to toolbar container

### LINKED ISSUE #1390: [Accessibility]: Multiple toolbar components do not have unique labels
### Contact Details

_No response_

### Environment

Browser, Assistive technology used to verify

### What happened? What did you expect to happen instead?

When there are two or more charts with toolbar's enabled, the toolbar role needs to have an aria-label or aria-labelledby attribute (in this case, since there is no text - it should be aria-label)

image

### What WCAG 2.1 checkpoint does the issue violate?

4.1.2 Name, Role, Value

### Version

@carbon/charts@1.2.0

### Data & options used

```js
When there are two charts on the page, used IBM Accessibility Assessment
```

### Relevant log output

_No response_

### Codesandbox example

https://codesandbox.io/s/yu4jyv

Akshat55:
closes #1390

### Updates
- Each `chart` now has an id element
- Initializing chart id in init function (it is called before constructor is called 🤔)
- Decided to add `id` to the chart to clarify what the toolbar aria-label is referring too
- Added aria-label attribute to toolbar container
  - Aria label has the following label: `chart-${id} toolbar`
  - Using aria-label instead of aria-labelledby since we do not have text within container (only icon)

image
- 4.1.2 Name, Role, Value is no longer reported by the checker

- [ ] Demos all features
- [ ] Documented/annotated
- [ ] Matches UI/UX specs
- [ ] Accessible

### DIFF
--- a/packages/core/src/components/axes/toolbar.ts
+++ b/packages/core/src/components/axes/toolbar.ts
@@ -42,7 +42,12 @@ export class Toolbar extends Component {
 	}
 
 	render(animate = true) {
-		const container = this.getComponentContainer().attr('role', 'toolbar');
+		const container = this.getComponentContainer()
+			.attr('role', 'toolbar')
+			.attr(
+				'aria-label',
+				`chart-${this.services.domUtils.getChartID()} toolbar`
+			);
 
 		const isDataLoading = Tools.getProperty(
 			this.getOptions(),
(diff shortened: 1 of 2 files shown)


theiliad (approved):
LGTM

PR #1403, opened 2022-06-29
fix(core): remove dom-to-image dependency & add script to project

### LINKED ISSUE #1395: [Bug]: Remove Dom-2-Image as a dependency
### Contact Details

_No response_

### What happened?

Migrate dom-to-image script into the library & stop using the dependency. This should solve all vulnerability reports. Main issue with dom-2-image is that node_modules & bower files are uploaded to github/package - The package that was causing the jquery issue is no longer part of package.json but it still exists in node_modules which is most likely causing this issue.

### Version

@carbon/@charts1.2.1

### Data & options used

_No response_

### Relevant log output

_No response_

### Codesandbox example

_No response_

Akshat55:
fix #1395

### Updates
- Removes dom-to-image dependency
- Copy dom-to-image with permission notice to codebase


theiliad (approved):
LGTM, & @ninja511 was happy

davidicus:
@theiliad is there any plans to backport this to the pre v1.x versions? We are currently on `0.55.1` in the PAL and we are getting a lot of issues opened up about this.

PR #1408, opened 2022-07-11
feat(core): Show milliseconds in axis labels if necessary

hlyang397:
### Issue
If there are multiple data set in the same second (different millisecond) in time-serious stacked bar chart, only one data is displayed, and the other data will be ignored due to stacked key conflicts.

### Updates

- Include millisecond as part of time serious stacked keys to avoid conflicts

- Add a new short time interval stacked bar (all data are in the same second)
  -- http://localhost:9006/?path=/story/simple-charts-bar-stacked--vertical-stacked-bar-short-interval-time-series

image

- [ ] Demos all features
- [ ] Documented/annotated
- [ ] Matches UI/UX specs
- [ ] Accessible

### DIFF
--- a/packages/core/demo/data/bar.ts
+++ b/packages/core/demo/data/bar.ts
@@ -656,6 +656,43 @@ export const stackedBarTimeSeriesOptions = {
 	},
 };
 
+export const stackedBarShortIntervalTimeSeriesData = [
+	{ group: 'Dataset 1', date: new Date(2019, 0, 1, 8, 15, 20, 111), value: 0 },
+	{ group: 'Dataset 1', date: new Date(2019, 0, 1, 8, 15, 20, 222), value: 65000 },
+	{ group: 'Dataset 1', date: new Date(2019, 0, 1, 8, 15, 20, 333), value: 10000 },
+	{ group: 'Dataset 1', date: new Date(2019, 0, 1, 8, 15, 20, 444), value: 49213 },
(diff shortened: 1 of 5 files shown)


Akshat55 (changes requested):
Code looks good to me! 

But it would require the user to do some thinking & interacting with the chart to determine why the axis labels are the same... which isn't a good user experience.🤔

Could you please add milliseconds to the axis? In my opinion, you would need to remove the formatting in tooltip as well, thoughts on this @theiliad? 

Currently with both tooltip & axis formatted to **_not_** show milliseconds makes it difficult to determine what the chart is trying to portray.

hlyang397:
> Code looks good to me!
> 
> But it would require the user to do some thinking & interacting with the chart to determine why the axis labels are the same... which isn't a good user experience.🤔
> 
> Could you please add milliseconds to the axis? In my opinion, you would need to remove the formatting in tooltip as well, thoughts on this @theiliad?
> 
> Currently with both tooltip & axis formatted to **_not_** show milliseconds makes it difficult to determine what the chart is trying to portray.

@Akshat55 

Thanks for your review.

Please refer to my update https://github.com/carbon-design-system/carbon-charts/pull/1408/commits/9377efd8cf90c227763d441aa0ad5c9ecb3354f9.

Now the milliseconds shows in axis labels if necessary (when milliseconds is not 0)

Regarding to tooltip, it doesn't even show time (hh:mm:ss) right now. not sure if there are other considerations, ex the tooltip becomes too wide. 


image

Akshat55 (approved):

PR #1412, opened 2022-07-15
feat(core): allow theme to be set via options prop

### LINKED ISSUE #1411: [Enhancement]: Set theme via the `data-carbon-theme` option
### Contact Details

_No response_

### Summary

Allow users to set theme using options prop

### Justification

Instead of directly applying theme to them holder and putting that responsibility, enable users to set theme via options prop. By default, theme should be `white` to match the default theme of @carbon/styles.

NA

### "Must have" functionality

NA

NA

### Available extra resources

NA

Akshat55:
fix #1411

### Updates
- Updates docs to use `theme` option when `theme` radio value is selected.
- Marks svelte _additional_ `theme` as deprecated since V1
  - In next major version, should be removed
- Creates a new option called `theme` set to `white` by default. Accepts  `white`, `g10`, `g90`, `g100` as values, where `white` is default value.

### **Theme updates as you set change radio button**
image
### **Use `g100` chart theme in `g90` active theme**
image

- [ ] Demos all features
- [ ] Documented/annotated
- [ ] Matches UI/UX specs
- [ ] Accessible

### DIFF
--- a/packages/core/src/interfaces/charts.ts
+++ b/packages/core/src/interfaces/charts.ts
@@ -48,7 +48,7 @@ export interface BaseChartOptions {
 	 */
 	height?: string;
 	/**
-	 * Optionally specify the theme of the chart
+	 * Optionally specify a theme for the chart
 	 */
 	theme?: ChartTheme;
 	/**
--- a/packages/core/src/interfaces/charts.ts
+++ b/packages/core/src/interfaces/charts.ts
@@ -7,6 +7,7 @@ import {
 	TreeTypes,
 	DividerStatus,
 	ColorLegendType,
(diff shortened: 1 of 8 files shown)


packages/core/src/interfaces/charts.ts
```diff
@@ -46,6 +47,10 @@ export interface BaseChartOptions {
 	 * Optionally specify a height for the chart
 	 */
 	height?: string;
+	/**
+	 * Optionally specify the theme of the chart
```
Akshat55, line 51: Hmm this sounds weird, should I update this?

theiliad (commented):
LGTM

PR #1419, opened 2022-08-04
fix: show correct time format in ticks with milliseconds

hlyang397:
### Updates
- update the tick format with milliseconds to be consistent to format without millisecond
- update the demo data

Before the fix:
wrongFormat

After the fix:
correct

- [ ] Demos all features
- [ ] Documented/annotated
- [ ] Matches UI/UX specs
- [ ] Accessible


theiliad (changes requested):
I'm still seeing these wrong formats showing up.

I think these might have started when the milliseconds support was added. Is that correct @hlyang397 ?

image

hlyang397:
> I'm still seeing these wrong formats showing up.
> 
> I think these might have started when the milliseconds support was added. Is that correct @hlyang397 ?
> 
> image

Yes, that's a bug from my change, Thanks for catching this.
It is fixed in https://github.com/carbon-design-system/carbon-charts/pull/1419/commits/4856b788d96b3cd5c54ae4a2def7cbee9bece36e.

theiliad (changes requested):
Hi @hlyang397, seems like some of the charts are still having tooltip issues

image

I'm wondering why we need this piece altogether?
```js
try {
  // it's a correct ISO format Date string
  if (typeof value === 'string') {
    return format(Date.parse(value), 'MMM d, yyyy');
  }
} catch (e){
   // not a valid ISO format string
}
```

hlyang397:
> Hi @hlyang397, seems like some of the charts are still having tooltip issues
> 

> I'm wondering why we need this piece altogether?
> 
> ```js
> try {
>   // it's a correct ISO format Date string
>   if (typeof value === 'string') {
>     return format(Date.parse(value), 'MMM d, yyyy');
>   }
> } catch (e){
>    // not a valid ISO format string
> }
> ```

@theiliad 
1.  You use Chrome, right ? I just realize that Firefox and Chrome have different implementations in Data.parse(), so my fix works in Firefox but not in Chrome.
My latest update https://github.com/carbon-design-system/carbon-charts/pull/1419/commits/181de4326647b896bb467d07b9546603f44f4125 should make sure the date string is in correct format.

2. We need this function because now value could be ISO date string (ex: '2019-01-01T00:05:06.222Z') instead of Date object. We need this function to transform the value to a formatted string to display in tooltip.

theiliad (approved):
LGTM

PR #1429, opened 2022-08-12
fix(core): prevent toolbar buttons from being clicked if shouldBeDisabled returns true

### LINKED ISSUE #1424: [Bug]: ToolbarControl clickFunction incorrectly called when shouldBeDisabled returns true
### Contact Details

justindm@us.ibm.com

### What happened?

The clickFunction for a ToolbarControl is incorrectly called when the shouldBeDisabled returns true. A disabled item should not be interactive and should not trigger click events.

### Version

@carbon/charts-react@0.58.2

### Data & options used

```js
{
  data: [
    { group: "Thing 1", value: 20000 },
    { group: "Thing 2", value: 65000 }
  ],
  options: {
    title: "Donut",
    resizable: true,
    donut: {
      center: {
        label: "Things"
      }
    },
    height: "400px",
    width: "400px",
    toolbar: {
      enabled: true,
      controls: [
        {
          type: ToolbarControlTypes.CUSTOM,
          id: "custom-toolbar-control",
          text: "Disabled control",
          shouldBeDisabled: () => true,
          clickFunction: () => {
            alert("I should be disabled!");
          }
        }
      ]
    }
  }
}
```

### Relevant log output

_No response_

### Codesandbox example

https://codesandbox.io/s/disabled-toolbarcontrol-drives-clickfunction-biziqq

Akshat55:
fix #1424

### Updates
- Checks to see if shouldBeDisabled exists
- If it exists, runs the function
- If it doesn't exist, we automatically assume the toolbar button click function should execute

- [ ] Demos all features
- [ ] Documented/annotated
- [ ] Matches UI/UX specs
- [ ] Accessible

### DIFF
--- a/packages/core/src/components/axes/toolbar.ts
+++ b/packages/core/src/components/axes/toolbar.ts
@@ -401,6 +401,12 @@ export class Toolbar extends Component {
 
 	// Calls passed function && dispatches event
 	triggerFunctionAndEvent(control, event, element?) {
+		// Check if trigger is disabled
+		if (typeof control.shouldBeDisabled === 'function') {
+			if (control.shouldBeDisabled()) {
+				return;
+			}
+		}
 		// Call custom function only if it exists
 		if (typeof control.clickFunction === 'function') {
 			control.clickFunction(event);
--- a/packages/core/src/components/axes/toolbar.ts
+++ b/packages/core/src/components/axes/toolbar.ts
@@ -402,11 +402,13 @@ export class Toolbar extends Component {
 	// Calls passed function && dispatches event
 	triggerFunctionAndEvent(control, event, element?) {
 		// Check if trigger is disabled
-		if (typeof control.shouldBeDisabled === 'function') {
-			if (control.shouldBeDisabled()) {
-				return;
-			}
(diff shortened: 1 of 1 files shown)


packages/core/src/components/axes/toolbar.ts
```diff
@@ -401,6 +401,12 @@ export class Toolbar extends Component {
 
 	// Calls passed function && dispatches event
 	triggerFunctionAndEvent(control, event, element?) {
+		// Check if trigger is disabled
+		if (typeof control.shouldBeDisabled === 'function') {
+			if (control.shouldBeDisabled()) {
```
theiliad, line 406: why not have both conditions in 1 statement?

theiliad (changes requested):
LGTM

packages/core/src/components/axes/toolbar.ts
Akshat55: Short circuiting, wasn't used in other scenarios, but will switch to 1 statement.

theiliad (approved):
LGTM

Issue #1440, opened 2022-08-30
[Bug]: Svelte Carbon charts have upstream `sass` peerDep


josefaidt:
### Contact Details

n/a

### What happened?

When using the Svelte distribution of carbon charts, we receive missing peer dependency issues with `sass` when using package managers such as `pnpm` due to the upstream peer dep on [@carbon/styles](https://github.com/carbon-design-system/carbon/blob/main/packages/styles/package.json#L32). From what I can tell this appears to be optional, and I can disable the missing peer deps errors by providing the following `.npmrc` file

```
strict-peer-dependencies=false
```

However is it possible to include the following modifications to the `@carbon/chart`'s package.json to account for when developers opt to use the built CSS stylesheet?

```json
"peerDependenciesMeta": {
  "sass": {
    "optional": true
  }
}
```

### Version

```json
"@carbon/charts": "^1.5.2",
"@carbon/charts-svelte": "^1.5.2",
```

### Data & options used

```js
n/a
```

nstuyvesant:
@josefaidt - This modification (making sass an optional peerDependency) has been added to `@carbon/styles`. Would it be ok for you to close this issue?

theiliad:
I will close for now, pls let us know @josefaidt

PR #1441, opened 2022-08-30
chore: make sass an optional peer dependency of charts

josefaidt:
### Updates

- adds `sass` as an optional peer dep of `@carbon/charts`. The real peer dep is in `@carbon/styles` and rightfully so, therefore I'm not 100% certain this will work the way I expect it to, but it appears optional per the findings in #1440 

- [ ] Demos all features
- [ ] Documented/annotated
- [ ] Matches UI/UX specs
- [ ] Accessible

### DIFF
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -124,6 +124,11 @@
 		"webpack-cli": "3.3.11",
 		"webpack-dev-server": "3.7.0"
 	},
+	"peerDependenciesMeta": {
+		"sass": {
+			"optional": true
+		}
+	},
 	"publishConfig": {
 		"directory": "dist",
 		"access": "public"


nstuyvesant:
While `@carbon/charts` exports SCSS files, it is not required that an app that uses this component library has sass.

For example, you might import only the CSS files. Along those lines, my recommendation would be to not make `sass` a peerDependency.

IMO, I would reserve peerDependencies to indicate "this component library won't work with X or Y."

theiliad:
@nstuyvesant agreed

nstuyvesant:
@josefaidt - I ran into an issue where `@carbon/charts` StackBlitz examples require `sass` even though they are only importing CSS from `@carbon/styles`.

The problem is that `@carbon/styles` requires `sass` (unnecessarily). So I added [this PR](https://github.com/carbon-design-system/carbon/pull/14131) to that project to flag the `sass` peerDependency as optional.

Wrote this here as I'm wondering if this is actually your issue. Your PR for `@carbon/charts` is the same as my PR for `@carbon/styles`. The difference is that `@carbon/styles` has `sass` as a peerDependency while `@carbon/charts` does not.

If you think this solves your issue, please close this PR.

josefaidt:
Hey @nstuyvesant thanks for the update! I'll close this PR :)

nstuyvesant:

PR #1449, opened 2022-09-26
fix(core): wordcloud should not render if `value` is not a number

theiliad:
this PR adds a check where if the `value` field provided is not a number, it'll stop the rendering and throw an error

### DIFF
--- a/packages/core/src/components/graphs/wordcloud.ts
+++ b/packages/core/src/components/graphs/wordcloud.ts
@@ -51,27 +51,26 @@ export class WordCloud extends Component {
 			return;
 		}
 
-		const wordCloudData = [];
-		displayData.map(function (d) {
-			let value = d[fontSizeMapsTo];
-
-			if (typeof d[fontSizeMapsTo] !== 'number') {
-				throw Error(
-					'Badly formatted WordCloud data. `value` should only be an integer or float'
-				);
-			}
-
-			wordCloudData.push({
-				[groupMapsTo]: d[groupMapsTo],
-				text: d[wordMapsTo],
-				size: value,
-				value,
-			});
-		});
-
 		const layout = cloud()
 			.size([width, height])
-			.words(wordCloudData)
+			.words(
+				displayData.map(function (d) {
+					let value = d[fontSizeMapsTo];
+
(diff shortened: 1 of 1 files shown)


packages/core/src/components/graphs/wordcloud.ts
```diff
@@ -51,18 +51,27 @@ export class WordCloud extends Component {
...
+			}
+
+			wordCloudData.push({
+				[groupMapsTo]: d[groupMapsTo],
+				text: d[wordMapsTo],
+				size: value,
+				value,
+			});
+		});
+
```
Akshat55, line 71:
```suggestion

```

Akshat55: If this is the approach we have to take since some words are still rendering and then the error message is being shown, then instead of pushing, we should just be returning the object in the map function & `worldCloudData = displayData.map(function (d)....`

The current approach makes it seem that `map` is being used as a forEach

packages/core/src/components/graphs/wordcloud.ts
```diff
@@ -51,18 +51,27 @@ export class WordCloud extends Component {
...
-				displayData.map(function (d) {
-					return {
-						[groupMapsTo]: d[groupMapsTo],
-						text: d[wordMapsTo],
-						size: d[fontSizeMapsTo],
-						value: d[fontSizeMapsTo],
-					};
-				})
-			)
+			.words(wordCloudData)
```
Akshat55, line 74:
```suggestion
			.words(
				displayData.map(function (d) {
				   // Add check here & throw error if true. Otherwise continue? Error should be displayed instead of the render, correct?
				   if (typeof d[fontSizeMapsTo] !== 'number') {
				      throw Error(
					    'Badly formatted WordCloud data. `value` should only be an integer or float'
				      );
			       }
			       
					return {
						[groupMapsTo]: d[groupMapsTo],
						text: d[wordMapsTo],
						size: d[fontSizeMapsTo],
						value: d[fontSizeMapsTo],
					};
				})
			)
```

Akshat55 (changes requested):

Akshat55 (approved):

Issue #1400, opened 2022-06-24
[Question]: Export Time series data from line chart


jmarenin:
### Question for Carbon

I am currently trying to export the data that I am graphing on the carbon design line chart. However, when I export the data, instead of exporting the timestamp value such as '2022-06-23T17:50:00' it instead exports as 'June 24, 2022'. Is there any way to change this?

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/carbon-design-system/carbon/blob/f555616971a03fd454c0f4daea184adf41fff05b/.github/CODE_OF_CONDUCT.md)

jmarenin:
Any update on this?

naegelin:
I can confirm this is the case for me as well. Also when one hovers over the time series line chart the tooltip only shows the day even if the granularity of the data is minute, hour, etc.

naegelin:
After a bit more digging it appears the problem lies in two places:

For the tabular data modal / export feature referenced by @jmarenin the following line needs to be fixed:
https://github.com/carbon-design-system/carbon-charts/blob/master/packages/core/src/model/cartesian-charts.ts#L75 

For the issue with the tooltip:
https://github.com/carbon-design-system/carbon-charts/blob/master/packages/core/src/components/essentials/tooltip.ts#L253

and

https://github.com/carbon-design-system/carbon-charts/blob/master/packages/core/src/components/essentials/tooltip.ts#L259

I will see if I have some bandwidth this week to submit a PR.

naegelin:
If anyone has any thoughts as to the proper representation of the date please let me know. ScaleType is called "TIME" so rationally if should probably reflect an actual date + time format such as :

```
'MMM d, yyyy HH:mm:ss'

```

While the above is nicely formatted for humans it isn't a great datetime format for machines as say epoch or 

```
yyyy-MM-dd'T'HH:mm:ss*SSSZZZZ
```

Sheco:
I'm interested in this too.

theiliad:
This'll soon be address by our `locale` interface

theiliad:
You can also use `tabularRepModal` configs to customize values

https://charts.carbondesignsystem.com/documentation/interfaces/interfaces.BaseChartOptions.html#tabularRepModal

PR #1469, opened 2022-11-18
fixes #1468

### LINKED ISSUE #1468: [Bug]:  Tabular Data and Tooltip only show date level precision
### What happened?

This is a continuation of #1400 which I believe to be a bug.

Issue 1: When defining a scaletype of `time` in axis  options and the data passed into the chart is indeed date+time, the tooltip helper used on several charts will only show Date level granularity formatting. When using higher granularity time series (hourly, minute, etc) this becomes problematic as the tooltip will only ever just display the current date in the following format: `MMM d, yyyy` . 

An example of this can be seen here:
https://codesandbox.io/s/5v1s9b?file=/App.svelte
Notice the granularity of the data is hourly but the tooltip displays only the date.

The issue in code is in two locations:
https://github.com/carbon-design-system/carbon-charts/blob/master/packages/core/src/components/essentials/tooltip.ts#L253

and

https://github.com/carbon-design-system/carbon-charts/blob/master/packages/core/src/components/essentials/tooltip.ts#L259

Issue 2: When defining a scaletype of `time` in axis  options and the data passed into the chart is indeed date+time, the tabular representation made available by several charts will again only show Date level granularity. This means viewing hourly, minute-by-minute,etc data in the table representation is not possible because the values are again formatted as `MMM d, yyyy` .

An example of this can again be seen here:
https://codesandbox.io/s/5v1s9b?file=/App.svelte
Clicking on the tabular representation button will show that the table only displays the date even though the data is hourly.

The issue in code is in the following locations
https://github.com/carbon-design-system/carbon-charts/blob/master/packages/core/src/components/essentials/tooltip.ts#L259

I propose changing the formatting in all three places to `'MMM d, yyyy HH:mm:ss'` in order to properly represent up to second level precision in these two features. 

### Version

@carbon/charts@master

### Data & options used

_No response_

### Relevant log output

_No response_

### Codesandbox example

https://codesandbox.io/s/5v1s9b?file=/App.svelte

naegelin:
### Updates
Fixes #1468 

New Tabular Representation using  MMM d, yyyy HH:mm:ss

- [ ] Demos all features
- [ ] Documented/annotated
- [ ] Matches UI/UX specs
- [ ] Accessible

### DIFF
--- a/packages/core/src/components/essentials/tooltip.ts
+++ b/packages/core/src/components/essentials/tooltip.ts
@@ -250,13 +250,13 @@ export class Tooltip extends Component {
 		}
 
 		if (typeof value.getTime === 'function') {
-			return format(value, 'MMM d, yyyy');
+			return format(value, 'MMM d, yyyy HH:mm:ss');
 		}
 
 		try {
 			// it's a correct ISO format Date string
 			if (typeof value === 'string' && (/\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}.\d{3}Z/.test(value))) {
-				return format(Date.parse(value), 'MMM d, yyyy');
+				return format(Date.parse(value), 'MMM d, yyyy HH:mm:ss');
 			}
 		} catch (e){
 			// not a valid ISO format string
--- a/packages/core/src/model/cartesian-charts.ts
+++ b/packages/core/src/model/cartesian-charts.ts
@@ -72,9 +72,9 @@ export class ChartModelCartesian extends ChartModel {
 		const domainScaleType = cartesianScales.getDomainAxisScaleType();
 		let domainValueFormatter;
 		if (domainScaleType === ScaleTypes.TIME) {
-			domainValueFormatter = (d) => format(d, 'MMM d, yyyy');
+			domainValueFormatter = (d) => format(d, 'MMM d, yyyy HH:mm:ss');
 		}
-
+		
 		const result = [
 			[
 				'Group',


bx2 (approved):

aleksatr (approved):

FilipStamenkovic (approved):

theiliad:
Hi, unfortunately we can't modify the default date formatter in our tooltip to include seconds, as this might not apply to many users.

I suggest using the valueFormatter in options.tooltip to add formatting.

You can detect date values there and update the formatting https://charts.carbondesignsystem.com/documentation/interfaces/_interfaces_components_.tooltipoptions.html#valueformatter

Re. adding formatting to the toolbar, I suggest adding that as a feature in a PR. Some kind of a formatter that can run over the table entries.

naegelin:
The current tabular representation of a scale referred to as TIME literally does not show or export time.   

Your own demo which I have linked in #1468 shows the issue where multiple hours on the same day show the same date repeated over and over with different values because it omits hours.

I would also like to suggest that you not make generalized conclusions about what features may or may not apply to users without allowing for the actual community to comment and participate. In fact some users may be frustrated by the issue as apparent  in #1400 which has  gone unaddressed for months.

naegelin:
[Tabular representation bug](https://user-images.githubusercontent.com/1544569/202792200-d8abb9a0-108b-4f9d-b9c5-79ce8f75e721.jpg)

PR #1487, opened 2023-01-09
feat(core): Add choropleth chart

Akshat55:
### Updates
- Adds choropleth chart
  - Adds services & essentials to create other geospatial charts
- Reverses color scheme in dark theme for components using gradients (Heatmap & choropleth)
- Refactored heatmap & other models where necessary

#### Todo
- Setting appropriate tokens
- Data input format
- Currently, legend is the same as that of the heatmap - do we want to use a similar format to the one used [here](https://carbondesignsystem.com/data-visualization/complex-charts/#geospatial-charts)?

Light:
image
Dark:
image

- [ ] Demos all features
- [ ] Documented/annotated
- [ ] Matches UI/UX specs
- [ ] Accessible


jeanservaas:
@Akshat55 the only token we can leverage from Carbon is for the stroke value.

Originally you were using the _$layer-accent-01_ token for the stroke color around continents and between countries. Which is gray 20 in the light theme and gray 80 in the dark theme.

The problem is that when there's missing data (and sometimes missing continents even) the stroke was difficult to see. I suggested in Slack (after talking to Michael Gower) using _$border-subtle-selected-01_ instead, to pump up the contrast of the stroke a bit to gray 30 in the light theme and gray 70 in the dark theme.

Although gray 30/70 does not technically achieve 3:1 contrast against heir respective backgrounds, @mbgower said that we're trying to find a balance here between better contrast and not making the missing data so prominent that it looks like a fill accidentally and hurts the clarity of the design. i.e. we CAN settle for better definition while not meeting 3:1 for map outlines in a sitiuation like this.

There will also be NO FILL color in the areas with missing data, so it will just take on the background color.

@Akshat55 is there anything else you need?

jeanservaas:
@Akshat55 Sorry... I understand your concern about MISSING DATA looking like ZERO data if we just rely on stroke color and it's the same everywhere.

So there are two issues:

1) making sure missing data is differentiated from zero data
2) making sure there is enough contrast (even if we can't reach 3:1) on the strokes outlining regions on the map.

Heatmap uses hatch marks for missing data. Can we not also use this pattern for consistency?

image

mbgower:
> Making it lighter than the other strokes would actually make more sense, but it would have to be discernably lighter than the outline for regions with data

This is an interesting idea. There are a few considerations: what do you do on a border between a country with data and one without? In that situation, obviously the country with data would need an enhanced border, so there's no real difference. Logically the only time you're going to perceive that lighter 'no data' border is if the country abuts another country with no data, or where the country is on a coastline (assuming there aren't different stroke values for coastlines).

Also, what about the fact the stroke is going to be less noticeable on two abutting countries both with either dark fills OR  light fills?
Another way of thinking of this is that a potential way to tackle this is to have a 'shim' between each country which is just the page background. Then you have the stroke around each country by that 3:1 against the background colour. Then the fill colour does not factor into it in the same way.
(Of course, the inability to see the stroke colour is a wholly different consideration, hopefully solvable with our usual legend filtering abilities

jeanservaas:
I like the lines @Akshat55. I do worry about incredibly small land masses... but it's hard even to identify color in such small areas too.

In the smaller map the lines still seems a slight bit too spaced out.

image

I especially like the proximity of the lines when the map goes full screen

image

mbgower:
is there any reason the line spacing needs to alter so much with scale? When I zoom them up to about the same proportions, there is a marked difference on width.

Akshat55:
>I do worry about incredibly small land masses... but it's hard even to identify color in such small areas too.

Ideally we would want to offer a way to zoom in to regions. Users can also show a zoomed in map of ONLY Europe if they wish too. Identifying small land masses wouldn't be a problem in that case. 

@mbgower The hatched pattern doesn't scale with the the image when we zoom in, it's basically just a fill for the countries. If we were to compare the distance between each line in both zoomed in & normal state, the width should be the same. This is also the case with Heat map.

When we zoom in, more lines are required to fill the area of the country/cell because the area has increased, Number of lines shown is proportional to the width of the chart.

packages/core/src/model/choropleth.ts
```diff
@@ -0,0 +1,80 @@
...
+			const data = this.getDisplayData();
+			if (
+				!Tools.isEmpty(data) &&
+				!Tools.isEmpty(options.geoData.objects.countries)
+			) {
+				/**
+				 * @todo
+				 * We can either use name or id by default to generate this dictionary
+				 * Curently id & name are standard in geoJSON. Unfortunately, topojson does not have any standard
+				 * so feature objects can have any key. We suggest that they include name or id at the very least
```
theiliad, line 44: would geoJSON be the better option then? Would we only support that or both?

theiliad: Just trying to have some context

theiliad (changes requested):
@Akshat55 Could you pls share where you got the [packages/core/demo/data/topojson-110.ts](https://github.com/carbon-design-system/carbon-charts/pull/1487/files#diff-dbb4091c92832175c148095cd53d2e2b96b7619fe007d43803b7636985d717ba) file from?

Is it MIT-licensed?

packages/core/demo/data/CHART_TYPES.ts
```diff
@@ -29,6 +29,11 @@ export default {
 		angular: 'ibm-bullet-chart',
 		vue: 'ccv-bullet-chart',
 	},
+	ChoroplethChart: {
+		vanilla: 'EXPERIMENTAL_ChoroplethChart',
+		angular: 'experimental-ibm-choropleth-chart',
+		vue: 'experimental-ccv-choropleth-chart',
```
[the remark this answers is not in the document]
Akshat55: Can we get someone with experience with vue to verify if this will work. I've tried to run the storybook locally for vue & it states the select `EXPERIMENTAL-ccv-choropleth-chart` doesn't exist.

I'm not too sure how I can set a custom HTML tag for vue 🤔 Might be a storybook issue

packages/core/src/model/choropleth.ts
Akshat55: Going to create a new thread for this in this PR so we don't have to toggle resolved conversations in future.

Akshat55:
### Why use TopoJSON instead of GeoJSON?

Main reason I did not support GeoJSON for the time being is because TopoJSON has a reduced file size. In [TopoJSON repo](https://github.com/topojson/topojson/blob/master/README.md), it's stated that it offers a file size reduction of up to 80%. While working on this chart, I've observed varying GeoJSON files size from 200MB to even 2GB. To reduce bandwidth, it's highly recommended teams use TopoJSON with gzip compression. TopoJSON also significantly improves compression rounding off coordinate values.

One advantage of GeoJSON is that there is a [set format (RFC 7946)](https://geojson.org/). Based on the data I've seen for TopoJSON online, format seems to not match GeoJSON... this usually happens in the conversion phase. 


### Would we only support TopoJSON or both?
In the code, we convert TopoJSON to GeoJSON & render countries individually. So it's possible to add support for both in future.

TopoJSON has tools for manipulation like [topojson-cllient](https://github.com/topojson/topojson-client). Manipulation of GeoJSON _may_ be difficult without a library.

Akshat55:
> @Akshat55 Could you pls share where you got the [packages/core/demo/data/topojson-110.ts](https://github.com/carbon-design-system/carbon-charts/pull/1487/files#diff-dbb4091c92832175c148095cd53d2e2b96b7619fe007d43803b7636985d717ba) file from?
> 
> Is it MIT-licensed?

I got the data from [NaturalEarthData](https://www.naturalearthdata.com/about/). According to the [terms of use](https://www.naturalearthdata.com/about/terms-of-use/), it states:

> All versions of Natural Earth raster + vector map data found on this website are in the public domain. You may use the maps in any manner, including modifying the content and design, electronic dissemination, and offset printing. The primary authors, Tom Patterson and Nathaniel Vaughn Kelso, and all other contributors renounce all financial claim to the maps and invites you to use them for personal, educational, and commercial purposes.
>
> No permission is needed to use Natural Earth. Crediting the authors is unnecessary.

I used the following GeoJSON data (https://www.naturalearthdata.com/downloads/110m-cultural-vectors/110m-admin-0-countries/) & converted it to TopoJSON. I'm not too sure how frequently this data is updated... If we do find GeoJSON data that is under MIT on GitHub, I can convert it to TopoJSON and use that.

theiliad (changes requested):
@Akshat55 These buttons don't seem to work properly. Are we building these out?

image


image

theiliad (changes requested):
@Akshat55 
Could you pls:

1. Also disable knobs in storybooks in the other packages, they're still running slowly
2. Add the experimental flag in demo options so that the experimental notice shows up on the stories
3. I'm thinking we could even add `(experimental)` in the story category
image

Issue #1507, opened 2023-03-04
Make lodash includes tree-shaking friendly


wkeese:
### Summary

@carbon/charts ostensibly uses lodash-es to enable tree shaking.

However, it sabotages itself in tools.js, which pulls in all (or a lot) of lodash-es regardless of whether or not it's being used, via this code:

```js
export var Tools;
(function (Tools) {
    // Export these functions from lodash
    Tools.debounce = lodashDebounce;
    Tools.clone = lodashCloneDeep;
    Tools.merge = lodashMerge;
    Tools.unionBy = lodashUnionBy;
    Tools.removeArrayDuplicates = lodashUnique;
    Tools.clamp = lodashClamp;
    Tools.flatten = lodashFlatten;
    Tools.groupBy = lodashGroupBy;
    Tools.camelCase = lodashCamelCase;
    Tools.isEmpty = lodashIsEmpty;
    Tools.isEqual = lodashIsEqual;
    Tools.flatMapDeep = lodashFlatMapDeep;
    Tools.kebabCase = lodashKebabCase;
    Tools.fromPairs = lodashFromPairs;
    Tools.some = lodashSome;
```

That breaks tree shaking, at least in Rollup, even with the configuration `treeshake: "smallest"`.

I suggest you stop re-exporting lodash methods from the tools module and instead make the other modules directly import the lodash functions that they use.

Admittedly, your tools.ts file seems tree-shaking friendly:

```ts
// Functions
export namespace Tools {
	// Export these functions from lodash
	export const debounce = lodashDebounce;
	export const clone = lodashCloneDeep;
	export const merge = lodashMerge;
	export const unionBy = lodashUnionBy;
...
```

But the problem is that when you install from NPM, the above code has been transpiled into the tree-shaking *un*friendly:

```js
export var Tools;
(function (Tools) {
    // Export these functions from lodash
    Tools.debounce = lodashDebounce;
    Tools.clone = lodashCloneDeep;
    Tools.merge = lodashMerge;
    Tools.unionBy = lodashUnionBy;
...
```

And the situation is exacerbated when code like dom-utils.ts pulls in everything from Tools instead of just the functions it needs:

```js
import { Tools } from '../../tools';
...
const chartsprefix = Tools.getProperty(options, 'style', 'prefix');
```

So my other suggestion is to get rid of that

```js
export namespace Tools { ... }
```

wrapper around tools.ts' exports.   Can't it just export its methods directly?

### Justification

[...]

PR #1509, opened 2023-03-05
refactor(tools.ts): stop using a namespace for Tools

wkeese:
### Updates

- Stop using a namespace in tools.ts.  It's unnecessary and breaks tree shaking.
- Update other modules as per tools.ts changes.
- This is just part of the steps I think should be taken to make carbon-charts more tree-shaking friendly.  Other steps would include:
1) Replace `import * as Tools from "../tools"` with `import { mergeDefaultChartOptions, ... } from "../tools"`. 
2) Import lodash-es methods directly rather than re-exporting them from tools.ts.

Refs #1507.

Please use the [ignore spacing option](https://github.com/carbon-design-system/carbon-charts/pull/1509/files?diff=unified&w=1) when reviewing changes.

- [ ] Demos all features
- [ ] Documented/annotated
- [ ] Matches UI/UX specs
- [ ] Accessible


wkeese:

theiliad:
Hi @wkeese I think you need to push an update to get the DCO Assistant to update

theiliad (approved):
LGTM

theiliad:
recheck

Issue #1522, opened 2023-03-13
[Bug]: "default" is not exported by "../core/dist/services/essentials/dom-to-image.js", imported by "../core/dist/services/essentials/dom-utils.js".


nstuyvesant:
### Name

Nate Stuyvesant

### Are you an IBM employee?

- [ ] Yes

### Contact Details

nstuyvesant@gmail.com

### What happened?

As I was doing a build of a new version of charts-vue, got an error...
"default" is not exported by "../core/dist/services/essentials/dom-to-image.js", imported by "../core/dist/services/essentials/dom-utils.js".

Looking at packages/core/dist/services/essentials/dom-utils.js and packages/core/dist/services/essentials/dom-to-image.js, it looks like there is indeed no `export default` in front of the IIFE in dom-to-image.js.

### Version

@carbon/charts@1.6.14

### Relevant log output

```js
"default" is not exported by "../core/dist/services/essentials/dom-to-image.js", imported by "../core/dist/services/essentials/dom-utils.js".
file: /Users/nates/dev/carbon-charts/packages/core/dist/services/essentials/dom-utils.js:24:7
22: // MISC
23: import ResizeObserver from 'resize-observer-polyfill';
24: import domToImage from './dom-to-image';
           ^
25: var CSS_VERIFIER_ELEMENT_CLASSNAME = 'DONT_STYLE_ME_css_styles_verifier';
26: var DOMUtils = /** @class */ (function (_super) {
error during build:
RollupError: "default" is not exported by "../core/dist/services/essentials/dom-to-image.js", imported by "../core/dist/services/essentials/dom-utils.js".
    at error (file:///Users/nates/dev/carbon-charts/packages/vue/node_modules/rollup/dist/es/shared/node-entry.js:2125:30)
    at Module.error (file:///Users/nates/dev/carbon-charts/packages/vue/node_modules/rollup/dist/es/shared/node-entry.js:13316:16)
    at Module.traceVariable (file:///Users/nates/dev/carbon-charts/packages/vue/node_modules/rollup/dist/es/shared/node-entry.js:13701:29)
    at ModuleScope.findVariable (file:///Users/nates/dev/carbon-charts/packages/vue/node_modules/rollup/dist/es/shared/node-entry.js:12204:39)
    at FunctionScope.findVariable (file:///Users/nates/dev/carbon-charts/packages/vue/node_modules/rollup/dist/es/shared/node-entry.js:6950:38)
[...]

PR #1540, opened 2023-04-13
fix(1522): externalize dom-to-image-more

nstuyvesant:
### Updates
Externalizes dom-to-image to solve [issue 1522](https://github.com/carbon-design-system/carbon-charts/issues/1522). Problem occurs frequently when working in SvelteKit when running vite dev when there are imports like enums from charts core package.

- [ ] Demos all features
- [ ] Documented/annotated
- [ ] Matches UI/UX specs
- [ ] Accessible


theiliad (approved):
LGTM

nstuyvesant:
Now we're good!

theiliad:
> Hi @theiliad - Got several other PRS I wanted to submit (Storybook, vite everywhere, etc.) but was waiting for this one to get merged (as it gets rid of the offline packages). Any thoughts as to when this one can be merged?

Hi @nstuyvesant the issue right now is the `.yarn/cache` directory. Can we remove that? Because of that I can't see the full changes either, my browser just lags up 🙂

theiliad (changes requested):
😓 
I think there's an issue with yarn now on Netlify https://app.netlify.com/sites/carbon-charts-core/deploys/6440203d8a33c40008a82df5

This might be because of the `.yarnrc.yml` change?

nstuyvesant:
> Hi @nstuyvesant the issue right now is the `.yarn/cache` directory. Can we remove that? Because of that I can't see the full changes either, my browser just lags up 🙂

Just dropped two of the commits from this PR. That should resolve your issue. In general, I would like to get rid of the yarn-offline-mirror-cache from the repo as it's a storage challenge and makes contributing problematic.

The modified .gitignore but did not touch the directory for now. Perhaps a separate PR could be created to resolve some of the workspace issues around yarn and lerna?

theiliad:
> > Hi @nstuyvesant the issue right now is the `.yarn/cache` directory. Can we remove that? Because of that I can't see the full changes either, my browser just lags up 🙂
> 
> Just dropped two of the commits from this PR. That should resolve your issue. In general, I would like to get rid of the yarn-offline-mirror-cache from the repo as it's a storage challenge and makes contributing problematic.
> 
> The modified .gitignore but did not touch the directory for now. Perhaps a separate PR could be created to resolve some of the workspace issues around yarn and lerna?

It seems like builds are still showing the same error.

Why would we drop `yarn-offline-mirror-cache`? They seem to be working well... any specific reasons?

nstuyvesant:
Not all contributors have the available storage to keep all the dependencies for a fork in their GitHub account.

theiliad:
> Not all contributors have the available storage to keep all the dependencies for a fork in their GitHub account.

Oh interesting! I didn't know that. How does that happen?

nstuyvesant:
> Oh interesting! I didn't know that. How does that happen?

Let's say I update a bunch of dependencies or change the yarn version (which switches the cache from .tgz to .zip and changes all the cached dependencies), then I am incurring the storage for all the changed dependencies in my fork. This monorepo has a large set of dependencies so it's more likely to tip me over my limit. If others want to contribute, they will face the same thing if they change lots of dependencies on their forks.

Including the yarn-offline-mirror in the repo is like including node_modules. All of that can be reconstituted with a simple `yarn install`.

Also, yarn 1.22.19 is the last version of yarn to use the .yarnrc (without the .yml extension). The options for yarn-offline-mirror and yarn-offline-mirror-pruning no longer exist post 1.22.19.

nstuyvesant:
@theiliad - just pushed some small changes that keep yarn at 1.22.19 (for now) but solve problems for contributors while laying the foundation for a yarn upgrade (newer .yarnrc.yml format, packageManager in package.json, simplified workspace property in package.json).

Please let me know if you have any concerns or questions. After these items are merged, it will be much easier to cherry-pick non-breaking changes from my `next` branch.

theiliad (changes requested):
Awesome! Only thing I noticed is the contents of `[.yarn/releases/yarn-1.22.19.cjs](https://github.com/carbon-design-system/carbon-charts/pull/1540/files#diff-163fc4fe1e91bc008b60261f3e772e810f0f9a264356eddf36b654128883f00c)` seems weird, there are some weird symbols showing up in there.

Can we use the JS file available here instead?
https://github.com/yarnpkg/yarn/releases/tag/v1.22.19

I could also update it if you add me as a collab to your fork

theiliad:
> Awesome! Only thing I noticed is the contents of `.yarn/releases/yarn-1.22.19.cjs` seems weird, there are some weird symbols showing up in there.

Actually for licensing purposes, are we able to avoid having that file in the repo?

nstuyvesant:
> > Awesome! Only thing I noticed is the contents of `.yarn/releases/yarn-1.22.19.cjs` seems weird, there are some weird symbols showing up in there.
> 
> Actually for licensing purposes, are we able to avoid having that file in the repo?
> 
> Other than that looks good to me 💯 thanks for working on these

Hi again. That file represents the public sources to yarn 1.22.19. I didn't see any odd characters myself but it is very long. The license is referenced on line 1979. Many public repos have these like:
- Babel (https://github.com/babel/babel/tree/main/.yarn/releases)
- Angular (https://github.com/angular/angular/tree/main/.yarn/releases)
- Storybook (https://github.com/storybookjs/storybook/tree/next/.yarn/releases) 

The newer .yarnrc.yml expects the yarn sources to be there (and part of the project). If it's removed, the reference to it would also have to be removed from .yarnrc.yml but yarn versions after 1.22.19 will require it.

Here's a Q&A on the yarn website about not gitignoring the .yarn/releases folder... https://yarnpkg.com/getting-started/qa#which-files-should-be-gitignored.

If you are using modern yarn (versions 2.0 to 3.5.0), here's the install process. Once performed you can set a project to use 1.22.19 (or apply it globally).

The yarn release file is created there by the command, `yarn set version 1.22.19`.

Hope this helps!

theiliad:
Thanks @nstuyvesant 💯

MRumpold:
Hi, 
I believe the bug fix you mentioned is causing an issue in my project. (I'm using SvelteKit, started with vite dev as mentioned at the very beginning). When I call a page with an embedded chart, the following message appears on the console:
`
X [ERROR] Could not resolve "dom-to-image-more"

    node_modules/.pnpm/@carbon+charts@1.7.1_d3@7.8.4_sass@1.62.0/node_modules/@carbon/charts/services/essentials/dom-utils.js:24:23:
      24 │ import domToImage from 'dom-to-image-more';
         ╵                        ~~~~~~~~~~~~~~~~~~~

  You can mark the path "dom-to-image-more" as external to exclude it from the bundle, which will remove this error.
  `
  A vite build fails with the following error message:
  `
  [vite]: Rollup failed to resolve import "dom-to-image-more" from "D:/.../node_modules/.pnpm/@carbon+charts@1.7.1_d3@7.8.4_sass@1.62.0/node_modules/@carbon/charts/services/essentials/dom-utils.js".
  `
  These are my current Carbon dependencies from the package.json:

- "carbon-components-svelte": "^0.72.3",
- "carbon-icons-svelte": "^11.4.0",
- "carbon-preprocess-svelte": "^0.9.1",
- "@carbon/charts": "^1.7.1",
- "@carbon/charts-svelte": "^1.7.1",

I have already tried a lot including:
- clearing the pnpm cache 
- installing the dom-to-image-more package -> but it resulted in more errors. 

I'm not sure if this is a bug or a combination with the configuration.

Can someone please help me to get the project running again?

nstuyvesant:
If you are using rollup...
```
export default {
  // ...
  external: ['dom-to-image-more'],
  // ...
}
```

For vite...
```
export default {
  // ...
  optimizeDeps: {
    exclude: ['dom-to-image-more']
  }
}
```

Can potentially do the exclusion in the charts-svelte package itself. Previously, dom-to-image was part of the core project and caused problems with SvelteKit when running dev (vite). Now that it's externalized, it's easier to mitigate (with the two examples above).

nstuyvesant:
I just submitted https://github.com/carbon-design-system/carbon-charts/pull/1545 to simplify things for projects and packages that consume `@carbon/charts`

MRumpold:
Thank you very much, it works now with these settings.

MRumpold:
Hello,

I was a bit too hasty, unfortunately it does not always work. 
In vite.config.ts I added the following code as mentioned above:
```build: {
    rollupOptions: {
        external: ['dom-to-image-more']
    }
}```
I installed dom-to-image-more as a module, and added the following dependency to the package.json file:
```"dom-to-image-more": "^3.1.5"```
It works when I start it with **vite dev**, but not with **vite preview** or **vite build**. The browser console shows the following error message:
```TypeError: Failed to resolve module specifier "dom-to-image-more". 
Relative references must start with either "/", "./", or "../".```
Do you have any suggestions on how to fix this error? 
Thank you for your help!

nstuyvesant:
That is odd. On a separate note, there is a conflict between rollup and dom-to-image-more that I hope will be resolved by tomorrow on the dom-to-image-more side. At that point, I have a PR that will bundle dom-to-image-more within `@carbon/charts`.

As a temporary solution, perhaps downgrade to 1.6.14 until this is resolved. It should not take long.

nstuyvesant:
The update to `dom-to-image-more` was released this evening. This PR, once merged, should resolve your issue...
https://github.com/carbon-design-system/carbon-charts/pull/1545

PR #1558, opened 2023-05-03
Monorepo update - Proposed for 1.9.0

nstuyvesant:
#### After merging
- [ ] remove minor flag from lerna publish command in deploy.sh
- [ ] Remove reference to `pr-1554` in https://github.com/carbon-design-system/carbon-charts/blob/3964ece30c6395b94d130405a670d0dc902522fe/scripts/next.sh

### Updates
- Node.js 18.16.0
- Yarn 3.5.1 with workspaces and `yarn upgrade-interactive` plugin
- Lerna 7.0.2
- Implemented vite as the bundler for core, react, svelte, and vue (and their storybooks). Bundles are ES and UMD.
- Dev dependencies updated
- Root-level bash scripts don't assume lerna is installed globally
- Publishing content is controlled by package.json for each package via `files` array and publishConfig.directory (latter only applies to Angular)
- Root-level config for prettier used by packages
- Core package's CSS (min and maps) copied to dist folders of dependent packages on builds to remove need for developers to have `@carbon/charts` as a dependency if using `@carbon/charts-react` just for styles.
- Git ignore file in monorepo root used everywhere
- Eslint ignore file in monorepo root used everywhere (and shared by prettier)
- Packages no longer need bash shell scripts - scripts are in package.json
- Packages with dependencies on core are linked via yarn workspace 'workspace:*' rather than symbolic links
- Several package.json scripts run concurrently
- Corrected issue with d3: d3-sankey and d3-cloud are not included in d3 and therefore need to be peerDependencies and treated differently in vite.config.ts.
- tsconfig templates for IDE and Vite in monorepo root propagate to packages
- Corrected over 500 TypeScript lint issues on core project (bundled code is unchanged)
- Core tests run in vitest
- Added explicit exports to packages that were missing them (also solves monorepo dev issues without symbolic links)
- Made exports for barrel files in core more explicit to improve tree-shaking and for secure development (only export what's needed) and ensure generated docs don't miss anything.
- Added top-level exports to core to prevent developers from having to know folder structure / namespace (non-breaking - still have the folder structure)
- Removed unused imports that were there to overcome a bug in an old version of TypeScript (no longer applies)
- Updated Storybook to 7.x which adds vite support, dark mode and improvements for editing JSON options and data.
- Converted Storybook docs to use its support for MDX2.
- Common Storybook MDX docs are copied from core to the other packages to make it easier to browse the documentation.
- Added typing to several packages where it was missing
- Svelte package uses SvelteKit's built-in component library generation (vite). Bundle is not affected. SvelteKit is not required for those using the component library (really no change to generated bundle as it's just source code).
- Vue package uses TypeScript in single file components providing better error messages.
- TypeScript updated to 5.1.3 for all packages.
- Declaration files in dist are downleveled to TypeScript 3.4 for compatibility.
- Angular 16.1.x support
- StackBlitz examples for all packages accessed via Storybook demos
- SCSS cleanup for core
- Support for Svelte 3 and 4
- Script for prereleases with Lerna

Should be a non-breaking change despite all the changes.

- [ ] Demos all features
- [ ] Documented/annotated
- [ ] Matches UI/UX specs
- [ ] Accessible


theiliad (changes requested):
Hi @nstuyvesant, great PR!

Seeing a couple of issues right off the bat

1. the react storybook seems to no longer load, with a `yd is not defined` error

image

2. If you choose a theme other than g100, and you switch to another story, it switches back to g100

theiliad:
> @theiliad - The area where I need help is testing `@carbon/charts-angular` on old Angular versions. I've tested it back to Angular 12 but could use some help testing even older versions (not sure how far back you need to go).

I think what we could do is generate a build, and use locally in some older angular apps. I can help once everything is sorted out

nstuyvesant:
What's interesting about the `yd` error is that it only appears on Netlify. Try going to package/react and running `yarn storybook` and it functions perfectly. The source map is included so you can see the odd line Storybook added.

The error refers to something that is injected by Storybook. So it's an issue but not in the component library. I will take a look at it tomorrow. Initially, I was getting it when running the storybook locally but I hadn't added optimizeDeps.include in the vite.config.ts (but after doing that it ran fine locally).

Regardingreverting from a selected theme back to G100 as the default, that was somewhat intended but I imagine we could just store the selected theme in localStorage and recall it. Pretty easy to add.

theiliad:
> Regarding reverting from a selected theme back to G100 when switching, that was somewhat intended but I imagine we could just store the selected theme in localStorage and recall it. Pretty easy to add.

yeah I can drop that in

nstuyvesant:
@theiliad - Regarding the `yd` not defined thing... I know it's related to Storybook-injected code and not the component library but I'm not sure why it's happening.

Here's a list of what I've tried so far:

1. Upgraded React on the package to 18 to match what Storybook is using. Thought there might be some conflict.
2. Turned off Storybook telemetry (not carbon) in case that was interfering.
3. Tried disabling the stories that are imported from core. Storybook has issues with MDX files coming from outside the package root (might need to eventually have a package.json script to just copy them, to the local package on builds).

Are you able to run the storybook for react on your local computer?

nstuyvesant:
Also, if you are able to run it locally, can you think of any Netlify environmental issue that may be causing the issue?

packages/core/src/services/essentials/dom-utils.ts
```diff
@@ -1,5 +1,5 @@
-import { select, Selection } from 'd3-selection'
-import ResizeObserver from 'resize-observer-polyfill' // only needed for testing with jsdom as modern browsers (>1%) already support it
+import { type BaseType, select, Selection } from 'd3-selection'
+// import ResizeObserver from 'resize-observer-polyfill' // only needed for testing with jsdom as modern browsers (>1%) already support it
```
theiliad: do we need this then?

packages/angular/.storybook/main.ts
```diff
@@ -0,0 +1,53 @@
...
+      // BUG: In theory, this line is supposed to workaround the Storybook issue but doesn't (webpack vs. vite)
+      '@storybook/blocks': dirname(require.resolve('@storybook/blocks/package.json'))
+    }
+    return config
+  },
+  features: {
+    storyStoreV7: false // required for storiesOf API
+  }
+}
+export default config
```
theiliad, line 53:
```suggestion
export default config

```

packages/angular/.storybook/manager-head.html
```diff
@@ -1,45 +1,22 @@
...
-
-<!-- Storybook override -->
-<script>
-	setTimeout(function () {
-		document.querySelector('div.sidebar-header > div > a').innerHTML =
-			'Carbon Charts - Angular';
-	}, 1000);
-</script>
+<meta name="twitter:image" content="https://charts.carbondesignsystem.com/angular/share.png" />
+<meta name="twitter:site" content="@_carbondesign" />
```
theiliad, line 22:
```suggestion
<meta name="twitter:site" content="@_carbondesign" />

```

packages/angular/.storybook/preview.ts
```diff
@@ -0,0 +1,46 @@
+import type { Preview } from '@storybook/angular'
+import { themes } from '@storybook/theming'
+
+// import { setCompodocJson } from '@storybook/addon-docs/angular'
+// import docJson from '../../../demo/documentation.json'
```
theiliad, line 5: do we need these?

packages/angular/angular.json
```diff
@@ -1,68 +1,84 @@
...
-			"prefix": "ibm",
-			"styleext": "css"
+			"prefix": "ibm"
 		},
 		"@schematics/angular:directive": {
 			"prefix": "ibm"
 		}
 	}
-}
+}
```
theiliad, line 84:
```suggestion
}

```

package.json
```diff
@@ -2,30 +2,40 @@
...
+		"commitizen": "^4.3.0",
+		"cz-conventional-changelog": "^3.3.0",
+		"husky": "^8.0.3",
+		"lerna": "^6.6.1",
+		"pinst": "^3.0.0",
+		"prettier": "^2.8.8",
+		"prettier-config-carbon": "^0.9.0"
+	},
+	"resolutions": {
+		"axios": "0.21.1"
```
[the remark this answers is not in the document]
nstuyvesant: Codesandbox was the culprit on this...
```
yarn why axios
├─ codesandbox@npm:2.2.3
│  └─ axios@npm:0.18.1 (via npm:^0.18.1)
```
I moved axios up to the min allowed by yarn but just bumped it 1.4.0 which is what Lerna's nx dependency uses.
Codesandbox for Vue seems to be working fine. I just pushed a commit.

If we see any codesandbox issues, can switch to StackBlitz as  two packages (Angular and React) already use it.

packages/angular/.storybook/carbon-charts-theme.ts
```diff
@@ -0,0 +1,8 @@
+import { create } from '@storybook/theming'
+
+export default create({
+	base: 'dark',
+	brandTitle: 'Carbon Charts - Angular',
+	brandTarget: '_self',
+	// fontBase: '"IBM Plex Sans", Arial, sans-serif' // throws off Storybook's icon alignment (only applies to left navigation)
```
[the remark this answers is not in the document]
nstuyvesant: Only affects the left navigation in storybook - not the stories themselves. While it can be enabled, the navigation icons aren't going to align without some smithing in CSS.

packages/angular/.storybook/main.ts
nstuyvesant: Just pushed an update with this

packages/angular/.storybook/preview-head.html
```diff
@@ -1,10 +1,6 @@
...
-	href="https://fonts.googleapis.com/css?family=IBM+Plex+Sans+Condensed:300,400|IBM+Plex+Sans:400,600&display=swap"
-	rel="stylesheet"
-	crossorigin="anonymous"
-/>
+<link rel="preconnect" crossorigin="anonymous" href="https://fonts.googleapis.com" />
+<link href="https://fonts.googleapis.com/css?family=IBM+Plex+Sans+Condensed:300,400|IBM+Plex+Sans:400,600&display=swap"
+	rel="stylesheet" crossorigin="anonymous" />
+<script>
+	window.global = window
+</script>
```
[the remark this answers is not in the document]
nstuyvesant: Prettier appears to drop the final line break for HTML and JSON files.

packages/angular/.storybook/preview.ts
nstuyvesant: Just took them out. It's part of Storybook for generating docs but I don't really see any benefit.

packages/angular/.storybook/preview.ts
```diff
@@ -0,0 +1,46 @@
...
+import '@carbon/charts/demo/styles.css'
+
+// setCompodocJson(docJson)
```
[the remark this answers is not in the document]
nstuyvesant: Just took that out as well.

packages/angular/angular.json
nstuyvesant: This is a prettier thing. It doesn't want the final CR or LF at the end of JSON or HTML files.

packages/angular/.storybook/manager-head.html
nstuyvesant: Also an issue of prettier automatically stripping final CR or LFs from JSON and HTML files.

packages/angular/.storybook/carbon-charts-theme.ts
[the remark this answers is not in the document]
nstuyvesant: Are you ok with leaving the left navigation as it is now?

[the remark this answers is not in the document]
theiliad: Yeah that's fine. I'll just look into it separately

nstuyvesant:
@theiliad - here's the first part of the minified code that triggers the `yd` error for the react storybook when it's deployed to Netlify...

It can found in packages/react/demo/bundle/assets/iframe-b41c5865.js. The iframe is where the storiesOf() story renders in storybook.
```js
;yd.__docgenInfo={description:"",methods:[{name:"addLineEventListener",docblock:null,modifiers:[],params:[],returns:null},{name:"addNodeEventListener",docblock:null,modifiers:[],params:[],returns:null},{name:"traverse",docblock:null,modifiers:[],params:[{name:"e",type:null},{name:"t",type:null},{name:"n",type:null}],returns:null},{name:"getRightArrowIcon",docblock:null,modifiers:[],params:[],returns:null},{name:"destroy",docblock:null,modifiers:[],params:[],returns:null}],displayName:"yd"};bd.__docgenInfo={description:"",methods:[{name:"init",docblock:null,modifiers:[],params:[],returns:null},{name:"destroy",docblock:null,modifiers:[],params:[],returns:null}],displayName:"bd"};nB.__docgenInfo={description:"",methods:[{name:"addEventListeners",docblock:null,modifiers:[],params:[],returns:null},{name:"getInvisibleAxisRef",docblock:null,modifiers:[],params:[],returns:null},{name:"getTitleRef",docblock:null,modifiers:[],params:[],returns:null},{name:"getNumberOfFittingTicks",docblock:null,modifiers:[],params:[{name:"e",type:null},{name:"t",type:null},{name:"n",type:null}],returns:null},{name:"destroy",docblock:null,modifiers:[],params:[],returns:null}],displayName:"Ks"};xd.__docgenInfo={description:"",methods:[{name:"addBoxEventListeners",docblock:null,modifiers:[],params:[],returns:null},{name:"addCircleEventListeners",docblock:null,modifiers:[],params:[],returns:null}],displayName:"xd"};Ed.__docgenInfo={description:"",methods:[{name:"init",docblock:null,modifiers:[],params:[],returns:null},{name:"addEventListeners",docblock:null,modifiers:[],params:[],returns:null},{name:"destroy",docblock:null,modifiers:[],params:[],returns:null}],displayName:"Ed"};Fi.__docgenInfo={description:"",methods:[{name:"init",docblock:null,modifiers:[],params:[],returns:null},{name:"createClipPath",docblock:null,modifiers:[],params:[],returns:null}],displayName:"Fi"};Od.__docgenInfo={description:"",methods:[{name:"unhighlightChildren",docblock:null,modifiers:[],params:[{name:"e",type:null}],returns:null},{name:"highlightChildren",docblock:null,modifiers:[],params:[{name:"e",type:null}],returns:null},{name:"getZoomClass",docblock:null,modifiers:[],params:[{name:"e"
```

Appears to be related to `@storybook/blocks`. The other packages have similar code injected but don't have issues.

nstuyvesant:
Added https://github.com/storybookjs/storybook/issues/22406

nstuyvesant:
Wondering if this is related to our `yd.__docgenInfo` issue?
https://github.com/styleguidist/react-docgen-typescript/issues/481

nstuyvesant:
The `yd` issue is definitely tied to `storybook:react-docgen-plugin` and appears to be related to tree-shaking. The issue is likely in the underlying `react-docgen` package. There is an issue logged that looks similar: https://github.com/styleguidist/react-docgen-typescript/issues/481

When I disable it in packages/react/.storybook/main.ts, the storybook site works on Netlify. The MDX stories don't load with it disabled.

Will look into some workarounds.

nstuyvesant:
@theiliad - Worked around the `yd` issue by switching from `@storybook/react-vite` to `@storybook/react-webpack5`. I left the vite build code in packages/react/.storybook/main.ts so we can switch back once Storybook fixes the open issue.

At this point, the open issues are:
1. Using localStorage to retain theme setting (though this probably shouldn't hold things up as the current release also doesn't retain the theme for the graph itself. Try going to https://charts.carbondesignsystem.com/?path=/story/simple-charts-bar-simple--vertical-simple-bar-discrete, change to G100 then go to the next chart on the left navigation. The background of the chart will switch to white.
2. Testing against old versions of Angular
3. Testing against old versions of React

theiliad:
> @theiliad - Worked around the `yd` issue by switching from `@storybook/react-vite` to `@storybook/react-webpack5`. I left the vite build code in packages/react/.storybook/main.ts so we can switch back once Storybook fixes the open issue.
> 
> At this point, the open issues are:
> 
> 1. Testing against old versions of Angular
> 2. Testing against old versions of React

I can help with these

nstuyvesant:
@theiliad - would it be possible to publish a non-release like `1.8.0-next.1` from the pr-1544 branch? This might make the setup of test environments a little easier.

packages/core/.storybook/preview.ts
```diff
@@ -0,0 +1,64 @@
...
+Classes that do not seem to be defined/used:
+	theme--white on div
+	intro
+	welcome__container
+	welcome__content
+	welcome__heading
+	welcome__heading--subtitle
+	welcome__heading--other
+	netlify
+*/
```
theiliad, line 64: Should we remove this comment?

packages/core/LICENSE.md
theiliad, line 1: Can we pls revert changes to these license files?

packages/core/.storybook/preview-head.html
```diff
@@ -1,10 +1,8 @@
...
-/>
+<link rel="preconnect" crossorigin="anonymous" href="https://fonts.googleapis.com" />
 <link
 	href="https://fonts.googleapis.com/css?family=IBM+Plex+Sans+Condensed:300,400|IBM+Plex+Sans:400,600&display=swap"
 	rel="stylesheet"
-	crossorigin="anonymous"
-/>
+	crossorigin="anonymous" />
+<script>
+	window.global = window
```
theiliad, line 7: do we need this line?

theiliad (changes requested):

package.json
```diff
@@ -2,30 +2,40 @@
...
+		"commitizen": "^4.3.0",
+		"cz-conventional-changelog": "^3.3.0",
+		"husky": "^8.0.3",
+		"lerna": "^6.6.2",
+		"pinst": "^3.0.0",
+		"prettier": "^2.8.8",
+		"prettier-config-carbon": "^0.9.0"
+	},
+	"resolutions": {
+		"axios": "^1.4.0"
```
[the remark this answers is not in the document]
nstuyvesant: Looks like this issue was logged 2 years ago (and codesandbox hasn't been updated for the time period). Another option would be to replace codesandbox with stackblitz as this process was already started.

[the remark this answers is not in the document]
nstuyvesant: BTW, the resolutions could be moved to the svelte and vue packages since they are the only two that still use codesandbox.

packages/angular/package.json
```diff
@@ -2,13 +2,91 @@
...
+    "build": "yarn build:package && yarn build:demo",
+    "storybook": "ng run charts-angular:storybook",
+    "start": "yarn storybook",
+    "lint:angular": "ng lint --ignore-path ../../.eslintignore",
+    "lint": "eslint . --ignore-path ../../.eslintignore",
+    "format": "prettier --write src/ --ignore-path ../../.eslintignore",
+    "clean": "rm -rf dist demo/bundle"
+  },
+  "peerDependencies": {
+    "@angular/common": "*",
```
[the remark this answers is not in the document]
nstuyvesant: Feel free to change that to what you prefer.

packages/angular/package.json
```diff
@@ -2,13 +2,91 @@
...
+    "@angular/core": "*"
```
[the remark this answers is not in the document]
nstuyvesant: Same as above - feel free to change that to what you prefer.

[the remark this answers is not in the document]
nstuyvesant: Actually, I'll update it.

LICENSE.md
```diff
@@ -2,180 +2,142 @@
                            Version 2.0, January 2004
```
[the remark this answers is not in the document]
nstuyvesant: I should mention that the build process now copies the license file from the top-level to each package before it's published so everyone who installs a package gets a copy.

package.json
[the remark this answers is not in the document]
nstuyvesant: Moved the resolutions to svelte and vue package.json files since only they use codesandbox.

packages/angular/package.json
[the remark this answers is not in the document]
nstuyvesant: Changed in latest commit.

packages/angular/package.json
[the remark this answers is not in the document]
nstuyvesant: Changed in latest commit

packages/svelte/static/favicon.png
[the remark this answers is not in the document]
nstuyvesant: Changed in latest commit. Also changed the HTML template to point to the new SVG.

packages/core/LICENSE.md
nstuyvesant: Previously, the packages like core had no license file. The new build script copies the one from the top-level of the monorepo to the package directory so it gets published by lerna.

And there were no wording changes to the license file. The whitespace was altered by prettier. You can confirm by reviewing with an IDE that allows you to turn off whitespace differences.

Are you OK with that?

packages/core/.storybook/preview.ts
nstuyvesant: Done in latest commit

packages/core/LICENSE.md
theiliad: Let's add them to the ignore file pls and revert the changes, I don't want to update any license files. I can help with this piece

nstuyvesant: I just pushed the reverted LICENSE.md (and its copies in each package) and added it to the eslint ignore file that prettier shares.

LICENSE.md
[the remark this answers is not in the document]
nstuyvesant: Reverted

packages/vue/LICENSE.md
```diff
@@ -0,0 +1,161 @@
+                                 Apache License
```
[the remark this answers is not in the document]
nstuyvesant: There was no license file in that directory. This is part of the build that copies it. I restored the root-level license's whitespace so it will get copied to each package's directory.

packages/vue/.storybook/preview-head.html
```diff
@@ -1,10 +1,6 @@
...
-<link
-	href="https://fonts.googleapis.com/css?family=IBM+Plex+Sans+Condensed:300,400|IBM+Plex+Sans:400,600&display=swap"
-	rel="stylesheet"
-	crossorigin="anonymous"
-/>
+<link rel="preconnect" crossorigin="anonymous" href="https://fonts.googleapis.com" />
+<link href="https://fonts.googleapis.com/css?family=IBM+Plex+Sans+Condensed:300,400|IBM+Plex+Sans:400,600&display=swap"
+	rel="stylesheet" crossorigin="anonymous" />
+<script>
+	window.global = window
```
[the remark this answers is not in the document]
nstuyvesant: Per the other comment, this is part of Storybook's template so I recommend leaving it in place.

packages/angular/src/lib/charts/base-chart.component.ts
```diff
@@ -1,102 +1,96 @@
...
 	 * Chart height
 	 */
-	@Input() height: any;
+	@Input() height?: string
 
 	/**
 	 * Chart container element ref
 	 */
-	@ViewChild("nChart") chartRef;
+	@ViewChild('chartRef') chartRef!: ElementRef<HTMLElement>
```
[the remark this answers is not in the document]
nstuyvesant: I believe this was some cruft from an older version. Only area-chart.component.ts referred to `nChart` in its template (and it really should have had an empty template like all the other charts).
 The name change in this abstract class should not have any effect on the consumption of the components in a project.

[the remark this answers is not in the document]
nstuyvesant: Update pushed removing the line and the import of ViewChild.

packages/vue/LICENSE.md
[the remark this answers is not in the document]
theiliad: thanks

packages/core/src/demo/styles.scss
```diff
@@ -230,62 +266,61 @@ div.container {
...
+	// 		font-size: 14px;
+	// 		line-height: 1.4;
+
+	// 		code {
+	// 			background-color: unset;
+	// 			color: white;
+	// 			padding: 0;
+	// 		}
+	// 	}
+	// }
```
theiliad, line 323: should we remove these then?

packages/core/src/stories/tutorials/1-color-palette.stories.mdx
```diff
@@ -0,0 +1,73 @@
...
+> This version relies on <b>Carbon v11</b>. If you're using Carbon v10, <a href="https://carbon-charts-0x.netlify.app" target="_blank" rel="noreferrer">see the legacy demo site</a>.
+
+Carbon charts recently updated the default color palettes for `@carbon/charts` to follow the new IBM Design Language data vizualization palette.
+This means more accessibility and usability in terms of the color variants provided by the library.
+
+> For more guidance on using the IBM Design Language dataviz palette, see [here](https://www.carbondesignsystem.com/data-visualization/color-palettes/)
+
+<br/>
+## Using a default palette
+You can add the following chart options if you want to choose a color palette option from the IBM dataviz palette.
```
theiliad, line 14:
```suggestion
You can add the following chart options if you want to choose a color palette option from the Carbon dataviz palette.
```

packages/react/LICENSE.md
```diff
@@ -0,0 +1,201 @@
+                                 Apache License
```
theiliad, line 1: we should revert this as well

packages/react/src/stories/2-diagrams-cards.stories.disabled-mdx
```diff
@@ -0,0 +1,14 @@
...
+	CardNode // This line causes an error for no particular reason
+	CardNodeColumn,
+	CardNodeSubtitle,
+	CardNodeLabel,
+	CardNodeTitle
+} from '../'
+
+<Meta title="Diagrams/Cards" />
+
+# CardNode
```
theiliad, line 14: this file seems incomplete

packages/react/src/stories/shapes.css
```diff
@@ -0,0 +1,3 @@
+.innerZoomElementWrapper {
```
theiliad, line 1: what is this element?

packages/svelte/LICENSE.md
```diff
@@ -0,0 +1,201 @@
+                                 Apache License
```
theiliad, line 1: another license file

theiliad (changes requested):

packages/core/src/services/essentials/dom-utils.ts
```diff
@@ -393,140 +362,144 @@ export class DOMUtils extends Service {
...
+			const computedStyles = getComputedStyle(cssVerifierElement as any)
+
+			if ( // true even when styles are properly loaded
 				computedStyles.getPropertyValue('overflow') !== 'hidden' ||
 				computedStyles.getPropertyValue('opacity') !== '0'
 			) {
 				console.error(
 					'Missing CSS styles for Carbon Charts. Please read the Carbon Charts getting started guide.'
-				);
+				)
```
theiliad, line 415: Are we removing this piece?

theiliad (changes requested):

packages/core/src/services/essentials/dom-utils.ts
nstuyvesant: It's logging messages that there are missing CSS styles even when they are loaded as documented. Also, it's creating a `g` tag within the `div` which is not really a valid child for anything other than an `svg`.

A few options...

1. Fix the test to see if the required CSS is properly loaded (perhaps check for the presence of CSS custom properties) instead of overflow and opacity.
2. Add a CSS verifier element but do it as a `span` or `div` instead of as a `g`
3. Do #1 and #2
4. Take out that verification altogether and trust the implementer to follow the documentation regarding loading of the styles.

Any preference?

packages/angular/package.json
```diff
@@ -2,13 +2,91 @@
...
+    "storybook": "ng run charts-angular:storybook",
+    "start": "yarn storybook",
+    "lint:angular": "ng lint --ignore-path ../../.eslintignore",
+    "lint": "eslint . --ignore-path ../../.eslintignore",
+    "format": "prettier --write src/ --ignore-path ../../.eslintignore",
+    "clean": "rm -rf dist demo/bundle"
+  },
+  "peerDependencies": {
+    "@angular/common": "^12.0.0 || ^13.0.0 || ^14.0.0 || ^15.0.0 || ^16.0.0",
+    "@angular/core": "^12.0.0 || ^13.0.0 || ^14.0.0 || ^15.0.0 || ^16.0.0"
```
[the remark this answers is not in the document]
nstuyvesant: I can extend the versions. What would you prefer it go back to?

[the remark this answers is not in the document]
nstuyvesant: Added peerDependencies back to 6 though I think from a practical perspective, usage for sub version 12 is unlikely at this point.

packages/react/src/stories/2-diagrams-start-here.stories.tsx
```diff
@@ -0,0 +1,255 @@
...
+					style={{
+						width: '100%',
+						height: '500px',
+						marginTop: '2em',
+						border: 0,
+						borderRadius: '4px',
+						overflow: 'hidden'
+					}}
+					title="carbon-charts-react-elkjs-diagram"
+					allow="accelerometer; ambient-light-sensor; camera; encrypted-media; geolocation; gyroscope; hid; microphone; midi; payment; usb; vr; xr-spatial-tracking"
```
theiliad, line 104: oh interesting, we should probably remove these

packages/core/src/stories/all.stories.ts
```diff
@@ -0,0 +1,153 @@
...
+const introStories = storiesOf('Docs', module)
+
+// Loop through the demos for the group
+introStories.add(
+	'Welcome',
+	() => {
+		// container creation
+		const container = document.createElement('div')
+		container.setAttribute('class', 'container intro')
+		// container.style.backgroundColor = DEFAULT_THEME
```
theiliad, line 24: do we need this then?

packages/core/vite.config.ts
```diff
@@ -0,0 +1,57 @@
...
+	optimizeDeps: {
+		include: [
+			'@carbon/colors',
+			'@carbon/layout',
+			'@carbon/themes',
+			'@carbon/utils-position',
+			'carbon-components',
+			'date-fns', // widely used - perhaps make a peerDependency and externalize
+			'dom-to-image-more',
+			'lodash-es' // widely used - perhaps make a peerDependency and externalize
```
theiliad, line 37: good idea, probably not a huge deal right now

packages/react/.storybook/main.ts
```diff
@@ -0,0 +1,52 @@
...
+	// viteFinal: (config) => {
+	// 	if (config.build) {
+	// 		config.build.chunkSizeWarningLimit = 1800
+	// 	}
+
+	// 	// Remove vite:dts - no need for declarations
+	// 	config.plugins = config.plugins!.filter((plugin) => plugin!.name !== 'vite:dts' /* || plugin!.name !== 'storybook:react-docgen-plugin'*/)
+
+	// 	return config
+	// },
```
theiliad, line 17: do we need these?

theiliad (changes requested):

packages/core/src/services/essentials/dom-utils.ts
theiliad: let's not take it out for now, that could be considered a breaking change. Can we just change it to a `span`?

I haven't seen it throw errors incorrectly ever yet... every time someone reported a false error, I look into it and realize that it is in fact correctly working. Did you find examples where it's actually not working as expected?

packages/angular/src/stories/diagrams/examples/examples.disabled-mdx
```diff
@@ -0,0 +1,80 @@
...
+<ng-template #wikiTemplate>
+  <svg ibmIconWikis size="20"></svg>
+</ng-template>
+```
+
+
+A simple composed diagram, using statically defined x and y coordinates.
+
+2. Programatic static layout
+
```
theiliad, line 80: can this be removed as well then?

theiliad (changes requested):

theiliad:
I've reviewed this as well as I can, there are tons of files and lots of changes, I might be missing something...

I'm asking all maintainers in Carbon to please review this PR

theiliad:
@nstuyvesant have you had a chance to review the latest of my comments? 🙂

nstuyvesant:
> @nstuyvesant have you had a chance to review the latest of my comments? 🙂

I think I addressed each one but please let me know if there were any that I missed or if my reply did not successfully address the concern.

theiliad:
Test results:
- React
   - 16 ✅ 
   - 18 ✅ 
- Vue ✅ 
- Svelte ✅ 
- Angular
   - 15 ✅

theiliad:
> Test results:
> 
> * React
>   
>   * 16 ✅
>   * 18 ✅
> * Vue ✅
> * Svelte ✅
> * Angular
>   
>   * 15 ✅

@nstuyvesant I think we also introduced some build errors in netlify

nstuyvesant:
@theiliad - added a `yarn next` script (scripts/next.sh) which should correctly publish pr-1554 as version 1.9.0-next.2 with the tag "next". It should also correctly translate `workspace:*` for packages/angular/dist when it publishes. If you can run that command and go ahead and push the version changes back to pr-1554, I'd appreciate it.

theiliad (approved):
LGTM

UserComponent:
This PR reverted the file-download options enhancement from [fc658e](https://github.com/carbon-design-system/carbon-charts/commit/fc658e03e188884a27ed60e95ded3f41db5d50ba). Thought I'd raise a comment, because this doesn't seem to be intentional, and I was troubleshooting why the options seemed to have vanished.

nstuyvesant:
Definitely not intentional! Seems to be a merge issue. Can you either open a PR and reapply these changes to packages/core/src/interfaces/chart.ts, model/model.ts, and services/essentials/dom-utils.ts or an issue so this gets the right visibility?

PR #1594, opened 2023-06-27
fix(1592): cp styles from dist to packages/core

nstuyvesant:
### Updates
- Copies styles.css and styles.min.css from packages/core/dist to packages/core

- [ ] Demos all features
- [ ] Documented/annotated
- [ ] Matches UI/UX specs
- [ ] Accessible


packages/core/styles.css
```diff
@@ -0,0 +1,6345 @@
+.cds--chart-holder {
```
theiliad, line 1: ouuu we might need these gitignored

theiliad (approved):
LGTM

PR #1602, opened 2023-06-28
fix(shadowed-vars)

theiliad:
### Updates
There were many shadowed variables, and seems like eslint-typescript wasn't catching these


theiliad:
CC @nstuyvesant

PR #1606, opened 2023-07-05
fix(react): remount issues

nstuyvesant:
### Updates
- In packages/react/src/charts/BaseChart.tsx:componentDidMount(), it only creates the chart if this.chart is undefined. Changed the componentWillUnmount() to set the instance method to undefined.
- Minor package bumps within semver except prettier.
- Improved comments on core/vite.config.ts so it's clear why that package needs to suppress compilation error (due to importing other package's package.json files that don't need type declarations).


theiliad (approved):
LGTM

PR #1608, opened 2023-07-06
fix(prettier), fix(vite.config - optimize html-to-image)

nstuyvesant:
### Updates
- Adds .vscode settings to read prettier config and share .eslintignore (tweaked .gitignore so this is added to repo)
- Convert prettier config from CommonJS to ESM (main prettier 3 benefit)
- Removed unneeded prettier devDependencies for individual packages (only needed at root level)
- Removed unneeded prettier config pointer in package.json files for packages
- Properly typed core/src/demo/utils/package-versions.ts
- Fixed core/vite.config.ts to optimize new `html-to-image` package
- Bumped packages (vite, vitest, @babel/core - latter only used by Storybook) within semver range and tested
- Added more comments for `vite-plugin-dts` in core since its version is going to be held back


theiliad (approved):
LGTM

PR #1609, opened 2023-07-06
fix(lerna): revert prettier to 2.8.8 for now

nstuyvesant:
### Updates
- Revert to prettier 2.8.8 for lerna publish (no idea why it cares about the prettier version but it does)
- Revert config file to CJS for now
- Update .vscode/settings to look for CJS file
- Update root-level package.json to point to prettier.config.cjs
- Bug fix for Svelte 4
- Bug fix for Vite


theiliad (approved):
LGTM

PR #1612, opened 2023-07-07
fix(svelte): make ref and chart props optional

nstuyvesant:
### Updates
- Svelte charts were giving TypeScript warnings if the user didn't provide the `ref` or `chart` property (when they weren't needed). Added defaults for those properties in BaseChart.svelte which aligns better to 1.8.0.
- Based on what we found yesterday with React, adjusted the destroy method of BaseChart.svelte (reset chart prop to null after destroying components but not the holder)
- Added CSS class 'cds--chart-holder' to the chart holder to align Svelte behavior with core rendering. 
- Fixed issue where toolbar on Svelte charts was displaying vertically initially until resized
- Added randomly generated HTML id to chart in case users need one (was in 1.8.0)
- Republished essential types like ChartTabularData and options for each chart type so users won't need `@carbon/charts` dependency. Ditto for most commonly-used enums.
- Updated `vite-plugin-dts` (just an update to the docs I contributed)
- Vite bug fix (postcss related)
- Updated `@carbon/styles` with `sass` marked as an optional peerDependency. Should allow us to drop `sass` from StackBlitz speeding up the loading (will test after next publish).
- Bumped Angular to 16.1.4 (components work better with older @angular/router)


theiliad (approved):
LGTM

PR #1615, opened 2023-07-10
fix(svelte dts), fix(docs), fix(example), fix(lerna w/prettier 3), fix(vite-plugin-dts/core)

nstuyvesant:
### Updates
- Make `chart` and `ref` properties optional for all chart types
- Remove unneeded vite optimizeDeps. End-users no longer need to optimize `@carbon/charts` dependencies because that's handled in core's vite config.
- Add `@carbon/charts` as peerDependency for `@carbon/charts-svelte` since svelte component library is in source form (not bundled)
- Update `@carbon/charts-svelte` StackBlitz example's vite config to remove unneeded optimizeDeps. 
- Corrected README.md files
- Fix vite-plugin-dts config for core so it works with 3.2.0
- Lerna fix for prettier 3
- Minor semver bump for dependencies


theiliad (approved):
@nstuyvesant are the `prettier  x   lerna` bugs resolved now?

also @metonym, any thoughts on this PR?

nstuyvesant:
@theiliad - the lerna patch that is included is intended to fix the prettier 3 compatibility. I can't test lerna publish to verify but others have commented on the original issue logged to their GH and said it resolved the issue. I left the prettier config in CJS for now to not tempt fate.

packages/svelte/README.md
```diff
@@ -10,67 +10,38 @@ Carbon Charts Svelte is a thin Svelte wrapper around the vanilla JavScript `@car
...
-// vite.config.js
-import { svelte } from '@sveltejs/vite-plugin-svelte'
-
-/** @type {import('vite').UserConfig} */
-const config = {
-	plugins: [svelte()],
-	optimizeDeps: {
-		include: ['@carbon/charts', 'carbon-components'],
-		exclude: ['@carbon/telemetry']
+		noExternal: process.env.NODE_ENV === 'production' ? ['@carbon/charts'] : []
```
metonym, line 44: I just tried this with my `sveltekit` example in [this repo](https://github.com/metonym/carbon-charts-svelte-examples/tree/master/sveltekit).

Development works fine, but I believe `carbon-components` still needs to be externalized.

After building for production and running `yarn preview`, I saw this error:

```sh
import { Modal as Modal$2 } from "carbon-components";
         ^^^^^
SyntaxError: Named export 'Modal' not found. The requested module 'carbon-components' is a CommonJS module, which may not support all module.exports as named exports.
CommonJS modules can always be imported via the default export, for example using:

import pkg from 'carbon-components';
const { Modal: Modal$2 } = pkg;

```

nstuyvesant: Hi @metonym. I just did the following...
```bash
git clone https://github.com/metonym/carbon-charts-svelte-examples.git
cd carbon-charts-svelte-examples/sveltekit
yarn install
yarn build
yarn preview
```
Then I invoked a modal on one of the charts...

image


I did not get the error you got related to `carbon-components`. Can you share some of your environmental details? I noticed an older version of `@carbon/charts-svelte` (^1.6.3) in the package.json. Can you check to see if your node_modules has version 1.11.8 of both `@carbon/charts` and `@carbon/charts-svelte`? The core `@carbon/charts` package switched to using vite/esbuild in 1.9.0 and above. That's when the optimizeDeps were added for `carbon-components` there. Wondering if your local env might not be up to date.

nstuyvesant:
@metonym - Just pushed a change to the package.json for the svelte package that I believe solves your issue with telemetry. It was a devDependency but needed to be a dependency.

theiliad:
@metonym let us know if everything's working now, and I can merge 🙂

metonym:
SGTM – thanks for getting to the bottom of this

metonym (approved):

PR #1617, opened 2023-07-12
Stop re-exporting lodash from tools.

### LINKED ISSUE #1507: Make lodash includes tree-shaking friendly

wkeese:
Second part of refactor to make lodash-imports friendly to tree shaking.

I don't see any reason for tools.ts to re-export the lodash methods.  It makes tree shaking more difficult and it also gives other code two ways to import the lodash methods, which could lead to inconsistency (ex: ruler-binned.ts).

This mainly changes where methods are imported from, but it also reverts clone() to its original name cloneDeep(), and reverts removeArrayDuplicates() to its original name uniq().

Fixes #1507.

### DIFF
--- a/packages/core/src/charts/area.ts
+++ b/packages/core/src/charts/area.ts
@@ -1,4 +1,4 @@
-import { cloneDeep } from "lodash-es";
+import { cloneDeep } from 'lodash-es'
 import { AxisChart } from '@/axis-chart'
 import { options } from '@/configuration'
 import { mergeDefaultChartOptions } from '@/tools'
--- a/packages/core/src/charts/area.ts
+++ b/packages/core/src/charts/area.ts
@@ -1,6 +1,7 @@
+import { cloneDeep } from "lodash-es";
 import { AxisChart } from '@/axis-chart'
 import { options } from '@/configuration'
-import { clone, mergeDefaultChartOptions } from '@/tools'
+import { mergeDefaultChartOptions } from '@/tools'
 import type { ChartConfig } from '@/interfaces/model'
 import type { AreaChartOptions } from '@/interfaces/charts'
 import { Skeletons } from '@/interfaces/enums'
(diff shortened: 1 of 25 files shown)


nstuyvesant:
@theiliad - LGTM

theiliad (approved):
LGTM

PR #1623, opened 2023-07-15
update(docs, templates, lerna, eslint, vite), cleanup(tasks.json - wrong folder/not used), chore(contributors - add self)

nstuyvesant:
### Updates
- Adjust GitHub templates to reference StackBlitz instead of Codesandbox
- Bump vite, eslint, vite-plugin-dts and vue-tsc (bug fixes)
- Remove tasks.json which was intended for VS Code but wasn't placed in .vscode directory so it was dormant. Also, not needed.


theiliad (approved):
LGTM

theiliad (approved):

PR #1624, opened 2023-07-18
fix(otherVersions link), bump(storybook)

nstuyvesant:
### Updates
- Fixed otherVersions widget that appears in core and react demos providing links to the equivalent story in the other package's storybooks
- Bumped Storybook to 7.1 (within semver)

image


theiliad (approved):
LGTM

PR #1628, opened 2023-07-20
fix(lerna version): hoist prettier-plugin-svelte due to lerna bug

nstuyvesant:
### Updates
- lerna version has some checks for prettier for some unknown reason. One of those checks is to look for `prettier-plugin-svelte` in the top-level node_modules. But it normally resides in packages/svelte/node_modules. This update adds it as a devDependency to the top-level package.json to force it to be hoisted.
- [An issue](https://github.com/lerna/lerna/issues/3788) has been opened on Lerna's GitHub to get the problem corrected.

### DIFF
--- a/package.json
+++ b/package.json
@@ -33,7 +33,8 @@
     "lerna": "^7.1.4",
     "pinst": "^3.0.0",
     "prettier": "^3.0.0",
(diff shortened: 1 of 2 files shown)


theiliad (approved):
Let's try it

PR #1629, opened 2023-07-20
fix(lerna): lerna version doesn't know we are sharing eslintignore file

nstuyvesant:
### Updates
- `lerna version` for some unknown reason messes with prettier. Because prettier's ignore is set in .vscode/settings.json (and points to eslint's ignore file), it tries to parse yarn.lock. This PR creates a prettier ignore file that is identical to the eslint ignore file so lerna knows not to parse yarn.lock.

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -1,0 +1,17 @@
+# Root-level only
+.git
+.yarn
+pages
+yarn.lock
+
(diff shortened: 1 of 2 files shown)


theiliad (approved):
LGTM

PR #1630, opened 2023-07-23
Update release.yml actions/checkout to v3 which does not leave head in detached state

nstuyvesant:
### Updates
- Changed release.yml to use actions/checkout@v3 (from v1). Newer version does not leave head in detached state (and does faster checkout.
- Commented out line in scripts/deploy.sh that performs `git checkout master` as this should not be needed.
- Bump eslint-plugin-vue within semver.


theiliad (approved):
LGTM

Issue #1638, opened 2023-08-02
[Bug]: Generated styles.css should use CSS custom properties for font-family


nstuyvesant:
### Name

Nate Stuyvesant

### Are you an IBM employee?

- [ ] Yes

### Summary

Currently, the generated styles.css references `IBM Plex Sans Condensed` or `IBM Plex Sans` in 12 different selectors and then in 24 CSS custom properties.

The problem is that the 12 selectors use IBM Plex faces directly rather than using any of the defined CSS custom properties. If users are loading styles as CSS, they will have to override the 12 selectors individually to change the font.

Here are the 12 selectors in the generated styles.css where the font is set without custom CSS properties:
```css
.cds--cc--axes g.axis g.tick text {
  fill: var(--cds-text-secondary, #525252);
  font-family: 'IBM Plex Sans Condensed', system-ui, -apple-system, BlinkMacSystemFont, '.SFNSText-Regular', sans-serif;
}

.cds--cc--axes g.axis .axis-title {
  font-family: 'IBM Plex Sans', system-ui, -apple-system, BlinkMacSystemFont, '.SFNSText-Regular', sans-serif;
  font-weight: 600;
  fill: var(--cds-text-primary, #161616);
}

.cds--cc--title p.title {
  color: var(--cds-text-primary, #161616);
  font-size: 16px;
  font-family: 'IBM Plex Sans', system-ui, -apple-system, BlinkMacSystemFont, '.SFNSText-Regular', sans-serif;
  font-weight: 600;
  white-space: nowrap;
  overflow: hidden;
  text-overflow: ellipsis;
  padding-right: 15px;
}

.cds--cc--tooltip {
  background-color: var(--cds-layer-02, #ffffff);
  pointer-events: none;
  transition-timing-function: cubic-bezier(0.4, 0.14, 0.3, 1);
  display: inline;
  visibility: visible;
  box-shadow: 0 1px 6px 0 rgba(0, 0, 0, 0.2);
  position: absolute;
  word-wrap: break-word;
  z-index: 1059;
  font-family: 'IBM Plex Sans Condensed', system-ui, -apple-system, BlinkMacSystemFont, '.SFNSText-Regular', sans-serif;
  transition: visibility 0s linear 0.1s, opacity 0.1s;
}

.cds--cc--threshold--label {
  background-color: #fa4d56;
  pointer-events: none;
  transition: opacity 0.1s;
[...]

PR #1639, opened 2023-08-02
fix(core/scss): Replace hard-coded font-family set by mixin to CSS custom properties then reuse

nstuyvesant:
### Updates
- In the generated styles.css file, 12 selectors had hard-coded font-families for two font faces. This seemed to be unintentional as CSS custom properties are used elsewhere. Restyling required users to explicitly restyle all 12 selectors just to change the font.
- Nothing changes with the look-and-feel of the fonts (still use IBM Plex Sans and IBM Plex Sans Condensed) but now a custom CSS property assignment like this can propagate to axes, tooltips, etc.
```scss
.cds--cc--chart-wrapper {
  --cc-font-family: Verdana;
  --cc-font-family-condensed: 'DIN Condensed';
}
```
- Closes https://github.com/carbon-design-system/carbon-charts/issues/1638

### DIFF
--- a/packages/core/scss/_type.scss
+++ b/packages/core/scss/_type.scss
@@ -3,7 +3,11 @@
 @use '@carbon/layout' as *;
 
 .#{$prefix}--#{$charts-prefix}--chart-wrapper {
-	font-family: font-family('sans-condensed');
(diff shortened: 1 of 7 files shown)


theiliad (approved):
LGTM

Issue #1643, opened 2023-08-08
[Bug]: UMD bundle for core not getting built


nstuyvesant:
### Name

Nate Stuyvesant

### Are you an IBM employee?

- [ ] Yes

### What happened?

When core is built via vite, it is creating es and cjs versions. It should have been building es and umd.

I will submit a PR to fix this. Otherwise, this documented approach to loading Carbon Charts will not work...

```html
<!DOCTYPE html>
<html>
  <head>
    <link rel="stylesheet" href="@carbon/styles/css/styles.css" />
    <link rel="stylesheet" href="https://unpkg.com/@carbon/charts/styles.css" />
  </head>
  <body>
    <div id="my-bar-chart"></div>

    <script src="https://unpkg.com/@carbon/charts/dist/umd/bundle.umd.js"></script>

    <script>
      // grab chart holder DOM element
      const chartHolder = document.getElementById('my-bar-chart')

      const stackedBarData = [
        // refer to tabular data format tutorial
      ]

      const stackedBarOptions = {
        // refer to chart specific options
      }

      // initialize the chart
      new CarbonCharts.StackedBarChart(chartHolder, {
        data: stackedBarData,
        options: stackedBarOptions,
      })
    </script>
  </body>
</html>
```

### Version

`@carbon/charts@1.11.17`

### What priority level would this be in your opinion?

P0

PR #1644, opened 2023-08-08
fix(umd): close issue 1643 - UMD bundle not generated

nstuyvesant:
# Problem solved
- Closes https://github.com/carbon-design-system/carbon-charts/issues/1643 (surprisingly, no one appears to have opened a case on this so I assume everyone is either using a very old version of the UMD bundle or using bundling tools)

### Updates
- Setup a separate vite config to generate UMD. This was necessary because you cannot have multiple entry points to the library for UMD. Those entry points were necessary for the namespaces like /interfaces, /components, etc.
- Updated core's package.json so the UMD bundle is properly exported
- UMD bundle includes dependencies so it can work properly in a standalone web page
- Updated the vanilla JS documentation to align with the configuration
- Bump dependencies within semver


theiliad (approved):
LGTM

Issue #1633, opened 2023-07-27
[Bug]: Charts with more than 100 data points throw an error when debouncing mouse events


a88zach:
### Name

Zach Tindall

### Are you an IBM employee?

- [ ] Yes

### What happened?

I believe a change to use an arrow function is causing the "this" context to be undefined

[packages/core/src/tools.ts](https://github.com/carbon-design-system/carbon-charts/commit/e4f6905218ec0f1a85799e32b124981311825893#diff-afb2f69fdf5248ae4491b017fe2eb68b06068246e24fd21215edc0cabe5af211)

![Screenshot 2023-07-27 at 9 03 10 AM](https://github.com/carbon-design-system/carbon-charts/assets/1724822/3f9d3df5-ba86-4ed3-a2d4-8830055381f0)

Whenever I render a chart with more than 100 data points, I'm getting this following error when I hover over the chart
> Cannot set properties of undefined (setting 'mousePosition')

This code is only called when charts have more than 100 data points as seen here:
https://github.com/carbon-design-system/carbon-charts/blob/master/packages/core/src/components/axes/ruler.ts#L221

### Version

@carbon/charts@^1.9.0

### What priority level would this be in your opinion?

P3

justindm234:
I believe I am also hitting this after upgrading @carbon/charts and @carbon/charts-react from 1.8.0 to 1.11.16

raizdev:
Me either.

theiliad:
I will address this asap

PR #1645, opened 2023-08-09
fix(1633): use of arrow function lost lexical this

nstuyvesant:
### Updates
- packages/core/src/tools.ts:debounceWithD3MousePosition():12 returned arrow function where there is no lexical this. Changed to return standard function.
- closes https://github.com/carbon-design-system/carbon-charts/issues/1633


theiliad (approved):

theiliad (approved):

raizdev:
Tested and fixed! Thanks for your quick response and handling, your guys doing a great job with that!

Issue #1646, opened 2023-08-09
[Bug]: UMD bundle not working as expected


nstuyvesant:
### Name

Nate Stuyvesant

### Are you an IBM employee?

- [ ] Yes

### What happened?

Have this StackBlitz HTML example that is supposed to use the UMD bundle for core:
https://stackblitz.com/edit/q34jth?file=index.html

Getting a console error: `Error: Charts.SimpleBarChart is not a constructor`

### Version

`@carbon/charts@1.11.18`

### Data & options used

```js
These are in the StackBlitz example
```

### Codesandbox example

https://stackblitz.com/edit/q34jth?file=index.html

### What priority level would this be in your opinion?

P1

nstuyvesant:
@theiliad - the generated dist/umd/bundle.umd.js is a bit hard to read but it looks like it exports everything. One approach we can try is to create a separate entry point for the UMD generation that leaves out types and classes that would not be directly used. I'll create a PR but interested in your opinion. The vite.umd.config.ts file looks OK to my eyes.

nstuyvesant:
@theiliad - Even with the simplification, the UMD bundle is not working. This is what I'm using to test it...
```html
<!DOCTYPE html>
<html>
  <head>
    <title>Carbon Charts Vanilla JavaScript Example</title>
    <meta charset="UTF-8" />

    <!-- Load Carbon Charts as Charts (UMD) -->
    <script src="https://unpkg.com/@carbon/charts@latest/dist/umd/bundle.umd.js"></script>

    <!-- Load required stylesheets -->
    <link
      href="https://unpkg.com/@carbon/styles@latest/css/styles.css"
      rel="stylesheet"
      crossorigin="anonymous"
    />
    <link
      href="https://unpkg.com/@carbon/charts@latest/dist/styles.css"
      rel="stylesheet"
      crossorigin="anonymous"
    />

    <!-- Load font used by Carbon Charts -->
    <link
      href="https://fonts.googleapis.com/css?family=IBM+Plex+Sans+Condensed:300,400|IBM+Plex+Sans:400,600&display=swap"
      rel="stylesheet"
      crossorigin="anonymous"
    />

    <style>
      .p-1 {
        padding: 2rem;
      }
    </style>
  </head>

  <body>
    <div class="p-1">
      <div id="app" style="width: 100%; height: 100%"></div>
    </div>
    <script>
      document.addEventListener('DOMContentLoaded', function () {
[...]

PR #1647, opened 2023-08-10
fix(umd): simplify exports

nstuyvesant:
### Updates
- As the generated dist/umd/bundle.umd.js for core is not functioning, https://github.com/carbon-design-system/carbon-charts/issues/1646, even though the vite.umd.config.ts looks OK, this PR reduces the exports for that bundle to just the bare essentials. The assumption is that anyone who needs to do something more advanced would be using the es modules via import.
- Even if the reduction of exports is not desirable in the long-run, we need to start somewhere with the troubleshooting.
- Bump lerna, sass, angular, @types/react within semver
- Note: due to a bug in `@carbon/styles`, restricted sass version for core. See https://github.com/carbon-design-system/carbon/issues/14415
- Bump node
- Bump JamesIves/github-pages-deploy-action@3.7.1 to JamesIves/github-pages-deploy-action@4.1.1

Note: The UMD bundle hasn't been working for some time and no one has logged an issue which suggests this is not the way developers consume `@carbon/charts`.


packages/core/package.json
```diff
@@ -137,7 +137,7 @@
     "react": "^18.2.0",
     "react-dom": "^18.2.0",
     "resize-observer-polyfill": "^1.5.1",
-    "sass": "^1.64.2",
+    "sass": "~1.64.2",
```
nstuyvesant, line 140: Prevent core from moving up to sass@1.65.x due to bug in @carbon/styles (until issue is resolved).

package.json
```diff
@@ -22,15 +22,15 @@
...
+    "node": "^18.17.1"
   },
   "packageManager": "yarn@3.6.1",
   "prettier": "./prettier.config.mjs",
   "devDependencies": {
     "commitizen": "^4.3.0",
     "cz-conventional-changelog": "^3.3.0",
     "husky": "^8.0.3",
-    "lerna": "^7.1.4",
+    "lerna": "^7.1.5",
```
nstuyvesant, line 33: Fixes some yarn workspace issues

theiliad (approved):
LGTM

PR #1649, opened 2023-08-11
fix(umd): Issue 1646 - change process.env.NODE_ENV to string

### LINKED ISSUE #1646: [Bug]: UMD bundle not working as expected

nstuyvesant:
### Updates
- Uses rollup to convert process.env.NODE_ENV to "production" just like pre-1.9.0 versions of core did.
- Also makes "warning" external just to be safe.
- Should close #1646


theiliad (approved):
LGTM

PR #1652, opened 2023-08-15
feat(scss): Optimized SCSS to remove need to separately load @carbon/styles

nstuyvesant:
# Problems this solves...
- Previously, `@carbon/charts` and the packages for React, Angular, Svelte, and Vue required that `@carbon/styles/css/styles.css` be loaded IN ADDITION TO the styles for the chart component libraries. This caused the entire  stylesheet for `@carbon/styles` to be loaded including styles for components not used by `@carbon/charts` increasing download times. It also added extra work to implement `@carbon/charts` (more moving parts and more documentation).
- Loading `@carbon/styles/css/styles.css` also applies a CSS reset which messes up Bootstrap and Material Design's CSS reset. That forces developers who do not want body-level `@carbon/styles` styling to use the shadow DOM for components.
- `@carbon/styles` was only needed for the toolbar and its modal (tabular representation). These are overflow-menu, button, modal, and data-table modules in `@carbon/styles` that should have been loaded by their corresponding Carbon Charts SCSS components.
- `@carbon/charts` SCSS used `@import` which is deprecated
- `@carbon/import-once` was used as a workaround because `@import` was used instead of `@use` for network diagram components.
- There were many cases of `@use 'something' as *;` which makes it difficult to understand where variables originate.
- The regular and condensed font CSS custom properties were associated with the chart wrapper which does not include the tooltip and modal dialog for the tabular representation of the data.
- Some SCSS modules were not getting loaded at all. These turned out to be duplicates of modules that were being loaded from a different folder.
- SCSS globals appeared in index.scss rather than their own file that could be loaded with `@use` by other modules.
- Some SCSS modules loaded other modules they did not use.
- StackBlitz examples needed to load `@carbon/styles` and `sass` (latter to avoid an install prompt by StackBlitz).
- In general, the SCSS flow was very difficult to follow previously.
- `@carbon/charts-angular` exports specified the dist folder for styles.css and styles.min.css but because the dist folder is what's published (rather than the parent), these paths were wrong.

# How to test this
Because StackBlitz normally loads styles.css from the latest version of `@carbon/charts`, please use [this StackBlitz example](https://stackblitz.com/edit/qcmxzn-hzmz2d?file=styles.css). The file packages/core/dist/styles.css has been copied and pasted into a file in the example then loaded. Please note that the Zoom button is always hidden in StackBlitz examples (not specific to this PR).


theiliad (approved):
LGTM

theiliad (changes requested):
@nstuyvesant something is wrong with the fonts here

before

image


after

image

nstuyvesant:
Wi

> @nstuyvesant something is wrong with the fonts here
> 
> before
> 
> image
> after
> 
> image

Will look into this. At a glance, there is a style identifier for Wordcloud missing that sets the font-family. It doesn't appear to inherit the font from the container.

nstuyvesant:
@theiliad - the Wordcloud issue is that the font-family being displayed in "IBM Plex Sans" instead of "IBM Plex Sans Condensed". Just need to explicitly use the CSS custom property for the latter instead of the default font for the container which is the former. Will fix later today.

nstuyvesant:
> @nstuyvesant something is wrong with the fonts here
> 
> before
> 
> image
> after
> 
> image

Fixed with latest push.

theiliad (approved):
LGTM

PR #1660, opened 2023-09-15
Donut and Pie label fix

RiyaJethwa:
### Updates
-added property percentageValue to formatter method data
-user can now display and edit labels for donut and pie charts 

image

### DIFF
--- a/packages/core/src/tools.ts
+++ b/packages/core/src/tools.ts
@@ -199,11 +199,16 @@ export function capitalizeFirstLetter(word: string) {
  * @param {string} key
  * @returns The percentage in the form of a number (1 significant digit if necessary)
  */
-export function convertValueToPercentage(item: any, fullData: any, key = 'value') {
+export function convertValueToPercentage(item: any, fullData: any, key = 'value',entireValue=false) {
 	const percentage =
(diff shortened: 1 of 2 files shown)


packages/core/src/tools.ts
```diff
@@ -199,11 +199,16 @@ export function capitalizeFirstLetter(word: string) {
...
 		(item / fullData.reduce((accum: number, val: any) => accum + val[key], 0)) * 100
-	// if the value has any significant figures, keep 1
-	return percentage % 1 !== 0 ? parseFloat(percentage.toFixed(1)) : percentage
+	//in need for entire float percentage value
+	if(entireValue){
+		return percentage
+	}else{
+		// if the value has any significant figures, keep 1
+		return percentage % 1 !== 0 ? parseFloat(percentage.toFixed(1)) : percentage
+	}
```
theiliad, line 211:
```suggestion
	if(entireValue){
		return percentage
	}

	// if the value has any significant figures, keep 1
	return percentage % 1 !== 0 ? parseFloat(percentage.toFixed(1)) : percentage
```

theiliad (changes requested):
@RiyaJethwa can you pls also run `prettier` to format the code?

Issue #1663, opened 2023-09-20
Change chart area theme


aaaknot30:
Hi - 

We are using svelte and carbon charts.  
Is there a way to change the color or theme of the chart area?  It is always white no matter the theme.

Thanks,
Kyle

nstuyvesant:
See the options on this example: https://charts.carbondesignsystem.com/svelte/?path=/story/simple-charts-area-stacked--stacked-area-time-series

The g100 theme has a dark background (black)...
```json
{
  "title": "Stacked area (time series)",
  "axes": {
    "left": {
      "stacked": true,
      "scaleType": "linear",
      "mapsTo": "value"
    },
    "bottom": {
      "scaleType": "time",
      "mapsTo": "date"
    }
  },
  "curve": "curveMonotoneX",
  "height": "400px",
  "theme": "g100"
}
```

aaaknot30:
Hi Nate - 
Could you double check why the StackBlitz version doesn't change chart area color? 
The StackBlitz version is what I deploy to local server. 
Thanks,
Kyle

nstuyvesant:
The StackBlitz version doesn't have the theme property in its options so it defaults to white. Adding the theme 'g100' doesn't automatically set the chart box to black. It's just adjusted for dark colors but you still have to provide that dark color.

You can set the CSS custom property `--cds-grid-bg` to `black` or whatever you want, it will change.

See https://stackblitz.com/edit/t3e7yt?file=index.html

aaaknot30:
Hi Nate - 
Yet, the theme automatically sets the chart area to dark or light in the first example you showed, (https://charts.carbondesignsystem.com/svelte/?path=/story/simple-charts-area-stacked--stacked-area-time-series).  
Can you show me where we would add or change the CSS class each time we changed the theme?
Thanks,
Kyle

nstuyvesant:
First, it seems like an omission in the generated CSS for the component library. The simplest way to put this is that the Storybook demos use a different generated CSS file than what is exported by the component library.

The demo CSS includes these definitions (while the base does not):
```css
.cds--chart-holder {
  --cds-grid-bg: #ffffff;
}

.cds--chart-holder[data-carbon-theme=g10] {
  --cds-grid-bg: #ffffff;
}

.cds--chart-holder[data-carbon-theme=g90] {
  --cds-grid-bg: #161616;
}

.cds--chart-holder[data-carbon-theme=g100] {
[...]

PR #1666, opened 2023-09-22
fix(scss): load tokens at beginning of _chart-holder.scss

nstuyvesant:
### Updates
- Moves loading of SCSS tokens to top of _chart-holder.scss so all CSS custom properties like --cds-grid-bg are set correctly.
- Simplies index.scss as the loading of the tokens needed to happen earlier anyway
- bump vitest
- Closes https://github.com/carbon-design-system/carbon-charts/issues/1663


theiliad (approved):
LGTM

Issue #1448, opened 2022-09-26
[Bug]: DonutChart - Issue with numeric values in Tooltip when Truncation type is set to none.


MariaStthpl:
### Contact Details

mariastthpl@gmail.com

### What happened?

Upon specifying the tooltip options of a Donut Chart as follows:

```
tooltip: {
    truncation: {
        type: "none"
    }
},
```
the decimal point of the number disappears.

| BEFORE      | AFTER |
| ----------- | ----------- |
| Truncation type is unset      | Truncation type is set to none  |
| ![NoTruncation](https://user-images.githubusercontent.com/33152738/192288636-9c3eae71-63a8-493e-9b72-117120ba550b.png) | ![Truncation](https://user-images.githubusercontent.com/33152738/192288699-527f8794-978a-4725-bf5b-7d1e32849855.png) |

Is this an expected behavior? I would expect that the decimal points remain intact regardless of the truncation type.

### Version

"@carbon/charts": "1.5.1"

MariaStthpl:
Let me mention additionally that `valueFormatter` is also not taken into consideration when we set `TRUNCATION.NONE`.

PR #1667, opened 2023-09-22
fix(tooltip-#1448):Issue with numeric value in tooltip when truncation type set to none

RiyaJethwa:
### Updates
-[#1448](https://github.com/carbon-design-system/carbon-charts/issues/1448)
-Numeric value is formatted when truncation is set to none.

### Demo screenshot or recordingimage


RiyaJethwa:

theiliad (approved):

Issue #1669, opened 2023-09-22
[Bug]: When charts styles are imported after carbon styles, some components break


janhassel:
### Name

Jan Hassel

### Are you an IBM employee?

- [X] Yes

### What happened?

When the styles from carbon charts are imported after carbon, they have precedence over the core styles and break some components.

```scss
@use '@carbon/styles' with ($font-path: '@ibm/plex');

@use '@carbon/charts/styles.css' as charts;
```

For example, CodeSnippet (inline variant):
```jsx
<CodeSnippet type="inline">Hello world</CodeSnippet>
```

See reproduction link below.

<img width="143" alt="image" src="https://github.com/carbon-design-system/carbon-charts/assets/28265588/43b61a5e-2d99-4790-a812-64fe2735e0ba">

<img width="169" alt="image" src="https://github.com/carbon-design-system/carbon-charts/assets/28265588/069a5b3c-1368-4b8d-86b9-404864677b60">

### Version

`@carbon/charts-react@1.13.3`

Note: I've upgraded from `@carbon/charts-react@1.6.3` where this issue was not appearing.

### Codesandbox example

https://stackblitz.com/edit/github-skkfec?file=package.json,src%2FApp.jsx

### What priority level would this be in your opinion?

P3

nstuyvesant:
@janhassel - This StackBlitz fork of your example uses Carbon Charts 1.0.1 https://stackblitz.com/edit/github-skkfec-q6ryuo?file=package.json and the problem appears to still be there. Please take a look.

PR #1671, opened 2023-09-25
fix(CSS pollution)

nstuyvesant:
## Overview
- Carbon Charts generated CSS with selectors not encapsulated within the top-level selector `--cds--chart-holder`. This caused the style sheet to overwrite CSS custom properties set by `@carbon/styles` and related packages.
- Particularly problematic components were modal and the overflow-menu both used by the toolbar component. The imported CSS set CSS custom properties on root. These were refactored to avoid that issue.
- Components grid-brush.scss and title only needed a limited number of SCSS variables and not generated classes. These now pull in the appropriate module where the variable is set to avoid generating unused CSS that conflicts with `@carbon/styles`.
- Restructured toolbar into a folder with modules for modal and the toolbar buttons since this is interdependent. If someone doesn't need the toolbar, they can write an optimized SCSS script loading only what they need.
- Closes https://github.com/carbon-design-system/carbon-charts/issues/1669
- Bumps Storybook, Eslint, @vitejs/plugin-react, SvelteKit and some @types within semver

This StackBlitz example submitted with bug 1669 has the newly generated CSS pasted in place of the import:
https://stackblitz.com/edit/github-skkfec-q6ryuo?file=src%2Findex.scss


theiliad (approved):
LGTM

Issue #1570, opened 2023-06-13
[Accessibility]: a11y scan violations found with charts-react


kalyanixraut:
### Name

Kalyani Raut

### Are you an IBM employee?

- [X] Yes

### Environment

Browser

### What happened? What did you expect to happen instead?

Accessibility scan violations are found with charts.

- Message: The 'id' "modal-title, modal-description" specified for the ARIA property 'aria-labelledby, aria-describedby' value is not valid
      
Level: violation
      XPath: /html[2]/body[1]/div[1]/div[1]/div[1]/div[3]
      Snippet: `<div tabindex="-1" aria-describedby="modal-description" aria-labelledby="modal-title" aria-modal="true" role="dialog" data-modal="true" class="bx--modal cds--modal">`
      Help: https://able.ibm.com/rules/archives/09March2023/doc/en-US/Rpt_Aria_ValidIdRef.html#%7B%22message%22%3A%22The%20'id'%20%5C%22modal-title%2C%20modal-description%5C%22%20specified%20for%20the%20ARIA%20property%20'aria-labelledby%2C%20aria-describedby'%20value%20is%20not%20valid%22%2C%22snippet%22%3A%22%3Cdiv%20tabindex%3D%5C%22-1%5C%22%20aria-describedby%3D%5C%22modal-description%5C%22%20aria-labelledby%3D%5C%22modal-title%5C%22%20aria-modal%3D%5C%22true%5C%22%20role%3D%5C%22dialog%5C%22%20data-modal%3D%5C%22true%5C%22%20class%3D%5C%22bx--modal%20cds--modal%5C%22%3E%22%2C%22value%22%3A%5B%22VIOLATION%22%2C%22FAIL%22%5D%2C%22reasonId%22%3A%22Fail_1%22%2C%22ruleId%22%3A%22Rpt_Aria_ValidIdRef%22%7D

<img width="1027" alt="image" src="https://github.com/carbon-design-system/carbon-charts/assets/22093326/542fe2b4-fd13-4e34-a8ab-22d826b06bd1">

      
      

### What WCAG 2.1 checkpoint does the issue violate?

The 'id' "modal-title, modal-description" specified for the ARIA property 'aria-labelledby, aria-describedby' value is not valid

### Version

`carbon/charts-react: 0.55.1`
`carbonV10`
`accessibility-checker: 3.1.47`

### What priority level would this issue be in your opinion?

P2

PR #1680, opened 2023-10-31
fix(modal-#1570):ARIA properties added to modal whilst being unopened

RiyaJethwa:
### Updates
- [#1570](https://github.com/carbon-design-system/carbon-charts/issues/1570)
- ARIA properties removed when modal is closed

image

### DIFF
--- a/packages/core/src/components/essentials/modal.ts
+++ b/packages/core/src/components/essentials/modal.ts
@@ -23,6 +23,15 @@ export class Modal extends Component {
 	}
 
 	handleShowModal = () => {
+		this.modal
+			.attr('data-modal', true)
+			.attr('class', 'cds--modal cds--modal')
+			.attr('role', 'dialog')
+			.attr('aria-modal', true)
+			.attr('aria-labelledby', 'modal-title')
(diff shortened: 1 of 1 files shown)


RiyaJethwa:

packages/core/src/components/essentials/modal.ts
```diff
@@ -23,13 +23,37 @@ export class Modal extends Component {
 	}
 
 	handleShowModal = () => {
+		this.modal
+			.attr('data-modal', true)
+			.attr('class', 'cds--modal cds--modal')
```
theiliad, line 28: why `cds--modal` twice?

theiliad, line 28:
```suggestion
			.attr('class', 'cds--modal')
```

theiliad (changes requested):

packages/core/src/components/essentials/modal.ts
RiyaJethwa: image

code already has been that way before the change was made, I'll remove the extra `cds--modal` and make a PR again

Issue #1718, opened 2024-01-02
[Bug]: Jest no longer works since v1.13.10


dabrad26:
### Name

David Bradshaw

### Are you an IBM employee?

- [X] Yes

### What happened?

Jest tests fail since version 1.13.10 (last working version is 1.13.8).  

```
FAIL src/components/ModalTransferDetails/ModalTransferDetails.test.tsx
  ● Test suite failed to run

    Cannot find module '@carbon/charts-react' from 'src/components/ModalTransferDetails/index.tsx'

    Require stack:
      src/components/ModalTransferDetails/index.tsx
      src/components/ModalTransferDetails/ModalTransferDetails.test.tsx

      1 | import React from 'react';
      2 | import '../../styles/components/ModalTransferDetails.scss';
    > 3 | import {GaugeChart} from '@carbon/charts-react';
        | ^
      4 | import {ChartTabularData, GaugeChartOptions, GaugeTypes, Statuses} from '@carbon/charts';
      5 | import {Observable, Subject, debounceTime} from 'rxjs';
      6 | import {SubManager} from '../../classes/SubManager';
```

Possible issues include the removal of UMD from the build for Vite.

[https://github.com/carbon-design-system/carbon-charts/compare/v1.13.8...v1.13.10#diff-28dac5df05657e5e97b4a2230a2[…]920e563809a5422407526e813291aR11](https://github.com/carbon-design-system/carbon-charts/compare/v1.13.8...v1.13.10#diff-28dac5df05657e5e97b4a2230a26f8314d5920e563809a5422407526e813291aR11)
https://github.com/carbon-design-system/carbon-charts/pull/1691

### Version

    "@carbon/charts": "1.13.17",
    "@carbon/charts-react": "1.13.17",

### Data & options used

```js
"jest": {
    "globalSetup": "./jest-setup.js",
    "resetMocks": false,
    "testEnvironment": "jsdom",
    "moduleNameMapper": {
      "\\.(jpg|ico|jpeg|png|gif|eot|otf|webp|svg|ttf|woff|woff2|mp4|webm|wav|mp3|m4a|aac|oga)$": "identity-obj-proxy",
      ".+\\.(css|less|scss)$": "babel-jest"
    },
    "transformIgnorePatterns": [
      "node_modules/(?!(@aspera-ui/starter|@ibm-aspera/connect-sdk-js|axios|@tauri-apps/api|@carbon/charts)/)"
    ]
  },
```

babel:

```
[...]

PR #1719, opened 2024-01-03
Fix 1718 - Old versions of jest rely on UMD bundles rather than ESM

nstuyvesant:
### Updates
- Closes https://github.com/carbon-design-system/carbon-charts/issues/1718
- Modified packages/react/vite.config.ts to ensure ESM and UMD output (lib.formats default to both so took out explicit one for ESM).
- Bumped storybook, @nrwl/devkit, vue, @vue/eslint-config-prettier, sass, typedoc, classnames, @sveltejs/package


packages/react/vite.config.ts
```diff
@@ -8,12 +8,16 @@ export default defineConfig({
 		chunkSizeWarningLimit: 600,
 		lib: {
 			entry: 'src/index.ts',
-			formats: ['es']
```
nstuyvesant, line 11: This was the main problem. The default value is ['es', 'umd'] so omitting it restores the UMD build.

packages/react/vite.config.ts
```diff
@@ -8,12 +8,16 @@ export default defineConfig({
...
+			name: 'ChartsReact',
+			fileName: (format) => `index.${format === 'es' ? 'm' : ''}js`
 		},
 		rollupOptions: {
 			external: ['react', 'react-dom'],
 			output: {
-				entryFileNames: '[name].mjs'
+				globals: {
```
nstuyvesant, line 17: Required for UMD

packages/react/vite.config.ts
```diff
@@ -8,12 +8,16 @@ export default defineConfig({
...
 		lib: {
 			entry: 'src/index.ts',
-			formats: ['es']
+			name: 'ChartsReact',
+			fileName: (format) => `index.${format === 'es' ? 'm' : ''}js`
 		},
 		rollupOptions: {
 			external: ['react', 'react-dom'],
 			output: {
-				entryFileNames: '[name].mjs'
```
nstuyvesant, line 16: Not needed as lib.filename outputs both - mjs for ESM and js for UMD.

theiliad (approved):

Issue #1716, opened 2023-12-29
[Accessibility]: getChartID not being generated/available causing accessibility issues


huineng:
### Name

Guy Huinen

### Are you an IBM employee?

- [X] Yes

### Environment

Automated testing tool and ruleset

### What happened? What did you expect to happen instead?

IBM Equal Access Accessibility Checker revealed several issues, but they are mainly caused by the absent of the ChartId

### What WCAG 2.1 checkpoint does the issue violate?

element_id_unique

### Version

 "@carbon/charts": "^1.13.15",

![image](https://github.com/carbon-design-system/carbon-charts/assets/3482445/8e0178a0-ed9c-4bce-a1a9-f02501ac96b7)

### Data & options used

```js
This is not really an issue if there's only one chart on the page
But it is when you have 2 or more because of the absence of the chartID, many elements are named the same
```

### Relevant log output

```js
you can see from the picture above in inspector that the ChartId is empty when rendering
```

### What priority level would this issue be in your opinion?

P0

huineng:
I see your latest release solved this issue.
Thanks, i will close this ticket

Issue #1523, opened 2023-03-14
[Accessibility]: Problems with charts' dialogs


wkeese:
### Name

Bill Keese

### Are you an IBM employee?

- [X] Yes

### Contact Details

wkeese@jp.ibm.com

### Environment

Automated testing tool and ruleset

### What happened? What did you expect to happen instead?

Every subclass of Chart creates a Modal like

```
<div class="cds--modal cds--modal" data-modal="true" role="dialog" aria-modal="true" aria-labelledby="modal-title" aria-describedby="modal-description" tabindex="-1"></div>
```

This was apparently designed for accessibility, i.e. so blind people could get the chart data even though they can't see the chart.  Ironically though it has a few accessibility problems of its own.  The two obvious problems here are:

* A chart’s dialog isn’t actually created until you hover the tooltip (or something like that), so initially the references to "modal-title" and "modal-description" are meaningless, causing accessibility warnings and also confusion for people that use screen readers.

The error is reported by accessibility-checker (although it would be a problem even if accessibility-checker failed to report it).  Strangely accessibility-checker only reports it when run from Node, not inside Chrome.  Not sure why.

```
    - Message: The 'id' "modal-title, modal-description" specified for the ARIA property 'aria-labelledby, aria-describedby' value is not valid
      Level: violation
      XPath: /html[2]/body[1]/div[1]/div[1]/div[1]/div[2]/div[1]/div[1]/div[1]/table[1]/tbody[1]/tr[15]/td[4]/div[1]/div[3]
      Snippet: <div tabindex="-1" aria-describedby="modal-description" aria-labelledby="modal-title" aria-modal="true" role="dialog" data-modal="true" class="bx--modal cds--modal">
[...]

PR #1722, opened 2024-01-04
fix(modal-domUtils):chartID initialization and addition of chartID in modal ARIA attributes

RiyaJethwa:
### Updates
- [#1523](https://github.com/carbon-design-system/carbon-charts/issues/1523)
- https://github.com/carbon-design-system/carbon-charts/issues/1716
- `chartID` previously not getting initialized by `init()` has been fixed
- addition of `chartID` in `aria-labelledby` and `aria-describedby` attributes

image


RiyaJethwa:

packages/core/src/components/essentials/modal.ts
```diff
@@ -23,13 +23,14 @@ export class Modal extends Component {
...
+		const id = this.services.domUtils.getChartID()
 		this.modal
 			.attr('data-modal', true)
 			.attr('class', 'cds--modal')
 			.attr('role', 'dialog')
 			.attr('aria-modal', true)
-			.attr('aria-labelledby', 'modal-title')
-			.attr('aria-describedby', 'modal-description')
+			.attr('aria-labelledby', `${id}__modal-title`)
+			.attr('aria-describedby', `${id}__modal-description`)
```
theiliad, line 33: these just point to an ID, but you also need to set these IDs to the modal

(currently the modal IDs are set to `modal-title` & `modal-description`, so we need to add the `${id}__` in there

theiliad: over here

image

theiliad (changes requested):

theiliad (approved):

Issue #1728, opened 2024-01-24
[Accessibility]: an auto generated id is included in the aria-label of the chart toolbar


kadiryazgan:
### Name

Kadir Yazgan

### Are you an IBM employee?

- [ ] Yes

### Environment

Browser

### What happened? What did you expect to happen instead?

Toolbar of carbon charts displays a hardcoded aria-label which includes a random ID.
This ID makes no sense to screenreader users. Until there is a better way to provide a custom aria-label value, removing the ID from the aria-label could be a solution.

<img width="889" alt="image" src="https://github.com/carbon-design-system/carbon-charts/assets/2729020/08b81707-6142-4e81-9603-94160c3ed5c6">

### What WCAG 2.1 checkpoint does the issue violate?

1.1.1

### Version

@carbon/charts@1.13.23

### Relevant log output

```js
<div class="cds--cc--toolbar" width="100%" height="100%" role="toolbar" aria-label="chart-114b94d7d9477 toolbar"><div class="cds--overflow-menu-options cds--overflow-menu--flip cds--overflow-menu-options cds--overflow-menu--flip" tabindex="-1"><ul role="menu"></ul></div><div class="toolbar-control cds--overflow-menu" aria-disabled="false" aria-label="Show as table">
			<button class="cds--overflow-menu__trigger cds--overflow-menu__trigger" aria-haspopup="true" aria-expanded="false" id="chart-114b94d7d9477-control-toolbar-showasdatatable" aria-label="Show as table">
				<svg focusable="false" preserveAspectRatio="xMidYMid meet" xmlns="http://www.w3.org/2000/svg" class="cds--overflow-menu__icon cds--overflow-menu__icon" viewBox="0 0 32 32" aria-hidden="true" style="will-change: transform; width: 20px; height: 20px;">
[...]

PR #1729, opened 2024-01-24
including a random id in the aria-label does not help screenreader users

kadiryazgan:
### Updates
removed the chartID from aria-label attribute.

related issue: https://github.com/carbon-design-system/carbon-charts/issues/1728


kadiryazgan:

theiliad (approved):

Issue #1737, opened 2024-02-02
[Bug]: StackBlitz examples for @carbon/charts-angular display zone.js errors


nstuyvesant:
### Application/Team

N/A

### What happened?

Steps to reproduce:
1. Goto the @carbon/charts-angular storybook - https://charts.carbondesignsystem.com/angular/?path=/story/simple-charts-area-simple--area-time-series-natural-curve
2. Click `Open in StackBlitz`

Initially, StackBlitz will display...

<img width="558" alt="image" src="https://github.com/carbon-design-system/carbon-charts/assets/5770938/41628bc6-4e4c-4b1a-80fc-54577833770a">

and then display this error:
```
Error in src/main.ts

The entrypoint for 'zone.js' could not successfully be resolved. The file '/turbo_modules/zone.js@0.14.2/dist/zone' does not exist. Please log an issue with the package author or check for an updated version of the package.
```

The StackBlitz Angular TypeScript template being used is supposed to include everything needed without adding all normal dependencies to the example's package.json. If those dependencies are added (zone.js, rxjs, @angular/common, @angular/core), then the error changes to:
```
Error in src/main.ts

v.getResolvedModule is not a function. (In 'v.getResolvedModule(r,e.text,AI(r,e))', 'v.getResolvedModule' is undefined)
```

### Version

`@carbon/charts-angular@1.13.27`

### Data & options used

```js
export default {
  title: 'Area (time series - natural curve)',
  axes: {
    bottom: {
      title: '2019 Annual Sales Figures',
      mapsTo: 'date',
      scaleType: 'time'
    },
    left: {
      mapsTo: 'value',
      scaleType: 'linear'
    }
  },
  curve: 'curveNatural',
  height: '400px'
}


```
export default [
  {
    group: 'Dataset 1',
    date: new Date(2019, 0, 1),
    value: 0
  },
  {
    group: 'Dataset 1',
    date: new Date(2019, 0, 6),
    value: -37312
  },
  {
    group: 'Dataset 1',
    date: new Date(2019, 0, 8),
    value: -22392
  },
  {
    group: 'Dataset 1',
    date: new Date(2019, 0, 15),
    value: -52576
  },
  {
    group: 'Dataset 1',
    date: new Date(2019, 0, 19),
    value: 20135
  },
  {
    group: 'Dataset 2',
    date: new Date(2019, 0, 1),
    value: 47263
  },
  {
    group: 'Dataset 2',
[...]

PR #1738, opened 2024-02-02
fix(stackblitz): env changes for angular examples to match latest StackBlitz Angular TypeScript template

nstuyvesant:
### Updates
- Closes https://github.com/carbon-design-system/carbon-charts/issues/1737
- Adjusts dependencies and devDependencies for Angular StackBlitz examples to match latest template from StackBlitz for Angular TypeScript (aside from CLI analytics which is now turned off). This includes using Angular 17.1.2.
- Demos: changed values that should be enums to enums. They still get transpiled to JS before going to StackBlitz but that won't necessarily be true in the future as using the enums properly is a best practice.
- Adjusted typing for properties that expected an enum to also allow string. As many of the examples are TypeScript but the contents of the options.ts were transpired to JS, this caused errors for demos that use properties typed as enums.
- Bumped @carbon/icons, @carbon/icons-react, @carbon/themes, @carbon/styles, @types/react, @nrwl/devkit, husky
- Prevented @carbon/icon-helpers from upgrading to the next minor version as it breaks the build


packages/core/src/demo/utils/stackblitz/angular.ts
```diff
@@ -71,98 +93,111 @@ bootstrapApplication(App)
...
-              "sourceMap": false,
-              "extractCss": true,
-              "namedChunks": false,
-              "aot": true,
-              "extractLicenses": true,
-              "vendorChunk": false,
-              "buildOptimizer": true
+    "$schema": "./node_modules/@angular/cli/lib/config/schema.json",
+    "cli": {
+      "analytics": "1e1de97b-a744-405a-8b5a-0397bb3d01ce"
```
theiliad, line 98: what's this? do we need it? can we remove it?

theiliad (changes requested):

packages/core/src/demo/utils/stackblitz/angular.ts
nstuyvesant: It's part of StackBlitz's Angular TypeScript template. The problem they are trying to solve is to prevent this prompt from coming up and stopping the execution...
`Would you like to share anonymous usage data with the Angular Team at Google under
Google’s Privacy Policy at https://policies.google.com/privacy? For more details  and how to change this setting, see http://angular.io/analytics. (y/N)`

I changed it to false and this appears to suppress the Angular prompt.

theiliad (approved):

PR #1751, opened 2024-02-13
fix(locale-configuration):localeCode defaults to en-US if navigator.language is not present

RiyaJethwa:
### Updates
- In case of navigator.language not being present then 'en-US' is used as localeCode.


RiyaJethwa:

theiliad (approved):

Issue #1759, opened 2024-02-20
[Bug]: Hide heatmap legend if there is no data


chrisdudley:
### Application/Team

IBM API Connect

### What happened?

When there is no data the heatmap legend displays NaN.
I couldn't work out how to disable the legend if there is no data.
I tried options.legend.enabled=false and it didnt seem to have any effect.

![image](https://github.com/carbon-design-system/carbon-charts/assets/7387884/b4a72a6f-dc63-4557-b64f-f8cd557d0fa7)

### Version

@carbon/charts-react: 1.14.1

### Data & options used

```js
I don't think this really affects things but this is how we're constructing the chart:

  const { t } = useTranslation("mfe_analytics")
  const localeObject = getLocaleObject()

  let domain = []
  for (let i = 23; i >= 0; i--) {
    const hour = i < 10 ? `0${i}:00` : `${i}:00`
    domain.push(hour)
  }

  const options = {
    title,
    tooltipTotal,
    axes: {
      bottom: {
        title: t("date"),
        mapsTo: "date",
        scaleType: "labels",
      },
      left: {
        title: t("hour"),
        mapsTo: "hour",
        scaleType: "labels",
        domain,
      },
    },

    heatmap: {
      colorLegend: {
        title: t("api_calls"),
      },
    },

    color: {
      gradient: {
        colors: [
          "#6929c4",
          "#4589ff",
          "#1192e9",
          "#08bdba",
          "#3ddbd9",
          "#6fdc8c",
          "#a7F0ba",
          "#fddc69",
          "#f1c21b",
          "#F7A361",
          "#eb6200",
          "#da1e28",
        ],
      },
    },
    fileDownload: {
      fileName: title,
    },
    data: {
      loading: loading,
    },
    timeScale: {
      addSpaceOnEdges: 1,
    },

    toolbar: {
      enabled: true,
      numberOfIcons: 3,
      controls: [
        {
          type: "Show as data-table",
        },
        {
          type: "Make fullscreen",
        },
        {
          type: "Custom",
          title: t("export_as_csv"),
          clickFunction: () => exportCSV(data, `${title}.csv`),
        },
        {
          type: "Custom",
          title: t("export_as_json"),
          clickFunction: () => exportJSON(data, `${title}.json`),
        },
        {
          type: "Export as PNG",
          title: t("export_as_png"),
        },
        {
          type: "Export as JPG",
          title: t("export_as_jpg"),
        },
      ],
    },
[...]

PR #1763, opened 2024-02-22
fix(heatmap-choropleth):legend hidden when data array is empty

RiyaJethwa:
### Updates
- https://github.com/carbon-design-system/carbon-charts/issues/1759
- hiding legend when data array is empty for heatmap and choropleth charts

image

image


RiyaJethwa:

PR #1768, opened 2024-02-27
fix(guage): number formatter not respected when not number

dabrad26:
A PR seems to have broken the use of numberFormatter for some components.  This is quite a large PR so hard to go through and fix each. But I think a quick look:

`localeNumberFormatter(Number(numberFormatter(d)), localeCode))`

numberFormatter should not be wrapped by Number since numberFormatter should allow the consumer to return a string numberFormatter?: (value: number) => string;

Commit introducing this:
https://github.com/RiyaJethwa/carbon-charts/commit/60ef1ec0003be80b3dadbeae26ceaab0327e4729

Checking for NaN was the best option I could find since numberFormatter had a default thing coming from `const gaugeChart: GaugeChartOptions = merge({}, chart, {` which would have been riskier to alter.

Example of the issue:

Screenshot 2024-02-27 at 12 07 29

### DIFF
--- a/packages/core/src/components/graphs/gauge.ts
+++ b/packages/core/src/components/graphs/gauge.ts
@@ -210,7 +210,10 @@ export class Gauge extends Component {
 			.merge(valueNumber as any)
 			.style('font-size', `${fontSize}px`)
 			.attr('text-anchor', 'middle')
-			.text((d: any) => localeNumberFormatter(Number(numberFormatter(d)), localeCode))
+			.text((d: any) => {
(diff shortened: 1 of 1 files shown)


dwosk (approved):

hsu022210 (approved):

dabrad26:
```
Eliad Moosavi:
We probably shouldn’t check based on NaN though, it should probably just check if number formatter is provided, and if it is, it’d just call it
```

PR #1798, opened 2024-04-26
feat(docs): Creation of new documentation website (PLEASE SQUASH COMMITS)

nstuyvesant:
# What is this?

- Created a new @carbon/charts documentation site in React that replaces Storybooks.
- Ripped out Storybook. These depended on a deprecated API (storiesOf).
- Each package now has a test page that can be viewed via `yarn dev` from the package subdirectories
- `npx lerna run build:test` builds the test pages in pages/test for each package
- Fixed eslint and prettier across all packages
- Fixed husky and set it to run eslint and prettier only on staged content (instead of entire monorepo)
- Added DOM tests for all packages which can be run in parallel by `yarn test:dom` from the monorepo root (`yarn test:dom:rebase` rebuilds the baseline).

image

image

image

image


packages/core/src/interfaces/axis-scales.ts
```diff
@@ -109,6 +109,10 @@ export interface BasedAxisOptions {
 	 * axis that's being broken into bins
 	 */
 	binned?: boolean
+	/**
```
nstuyvesant, line 112: The `correspondingDatasets` property belongs in AxisOptions rather than ComboChartAxisOptions because it's available for line and scatter charts as well. For a long time, there have been demo chart configs that use this property (and displayed an error in the IDE because the property didn't exist).

packages/core/src/interfaces/charts.ts
```diff
@@ -509,11 +508,11 @@ export interface TreeChartOptions extends BaseChartOptions {
  */
 export interface CirclePackChartOptions extends BaseChartOptions {
 	circlePack?: {
-		circles: {
+		circles?: {
 			fillOpacity: number
 		}
 		// depth of nodes to display
-		hierachyLevel: number
```
nstuyvesant, line 516: Spelling error

packages/core/src/interfaces/charts.ts
```diff
@@ -602,8 +601,8 @@ export interface ChoroplethChartOptions extends ThematicChartOptions {
 			title?: string
 			type: ColorLegendType | string
 		}
-		geoData: Topology
```
nstuyvesant, line 605: This was a bug. The prop `geoData` was expected to be at the top-level of `ChoroplethChartOptions` - not in the `choropleth` property. If you try to set choropleth.geoData, it does not work.

The original demo for Choropleth set the `geoData` at the top-level which is how I found this bug.

thumbnails/alluvial.svg
nstuyvesant, line 1: All of the thumbnails moved to the packages/docs/public/images

theiliad (changes requested):

packages/angular/charts-angular-test/src/app/app.component.ts
```diff
@@ -0,0 +1,65 @@
...
+		'ibm-meter-chart': MeterChartComponent,
+		'ibm-pie-chart': PieChartComponent,
+		'ibm-radar-chart': RadarChartComponent,
+		'ibm-scatter-chart': ScatterChartComponent,
+		'ibm-simple-bar-chart': SimpleBarChartComponent,
+		'ibm-stacked-area-chart': StackedAreaChartComponent,
+		'ibm-stacked-bar-chart': StackedBarChartComponent,
+		'ibm-tree-chart': TreeChartComponent,
+		'ibm-treemap-chart': TreemapChartComponent,
+		'ibm-wordcloud-chart': WordCloudChartComponent
```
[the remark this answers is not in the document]
nstuyvesant: Hi @theiliad - wasn't sure what you meant by this one as this is only for the test harness (charts-angular-test) and not related to the exported components for charts-angular.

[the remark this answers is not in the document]
theiliad: yes sorry that was a mistake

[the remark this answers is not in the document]
nstuyvesant: No worries - think this was the only open comment.

packages/core/package.json
```diff
@@ -91,30 +83,30 @@
 		"dompurify": "^3.1.5",
 		"html-to-image": "^1.11.11",
 		"lodash-es": "^4.17.21",
+		"prettier": "^3.3.2",
```
nstuyvesant, line 86: Should be a devDependency

packages/core/package.json
```diff
@@ -91,30 +83,30 @@
...
+		"publint": "^0.2.8",
```
nstuyvesant, line 87: Should be a devDependency

packages/core/package.json
nstuyvesant: fixed

packages/core/package.json
nstuyvesant: fixed

packages/docs/package.json
```diff
@@ -0,0 +1,100 @@
...
+		"@carbon/icons-react": "^11.43.0",
+		"@carbon/react": "^1.59.0",
+		"@stackblitz/sdk": "^1.10.0",
+		"highlight.js": "^11.9.0",
+		"react": "^18.3.1",
+		"react-dom": "^18.3.1",
+		"react-router-dom": "^6.23.1"
+	},
+	"devDependencies": {
+		"@carbon/icon-helpers": "^10.48.0",
```
nstuyvesant: Not needed.

theiliad (approved):
LGTM. This is a very large PR, so I've tried my best to review here. I would like others to also participate here

PR #1799, opened 2024-04-30
fix(react): remove react-jsx-runtime from bundles

theiliad:
react-jsx-runtime logic was showing up in our react bundles, this should remove it


theiliad:
Nvm, seems like it's good practice to keep it in there

Issue #1819, opened 2024-05-06
[Bug]: @use no longer works for importing style sheets


finken2:
### Application/Team

IBM Cloud Satellite

### What happened?

We were previously using `@use '@carbon/charts/styles'` with @carbon/charts 1.6.14, but this isn't working with newer version. I've tried many variations with 1.15.7 and can't get any of them to work.
`@import '@carbon/charts/styles.css';` does work, but i'd prefer to use `@use` if possible. 

```
SassError: Can't find stylesheet to import.
  ╷
4 │ @use '@carbon/charts/styles';
  │ ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  ╵
  client/scss/app.scss 4:1  root stylesheet
SassError: SassError: Can't find stylesheet to import.
  ╷
4 │ @use '@carbon/charts/styles';
  │ ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  ╵
  client/scss/app.scss 4:1  root stylesheet
    at Object.loader (/Users/edwardfink/containers/satellite-ui/node_modules/sass-loader/dist/index.js:56:14)
```

### Version

`@carbon/charts 1.15.7`

### What priority level would this be in your opinion?

P3

nstuyvesant:
@finken2 - what bundler are you using?

What happens if you do...
```scss
@use '@carbon/charts/scss'
```
or
```scss
@use '@carbon/charts/scss/index.scss'
```

finken2:
We're using webpack.  The first one fails the same way, but the 2nd one does work. A little weird to need to the `index.scss` there, but at least it works.  I tried like 100 different approaches, and i swear that was one of them, but who knows. Thanks for the suggestion.

I started this conversation on Slack by asking "what's the recommended way to load styles in 1.x?" and was directed to open an issue.  We should get the docs updated to whatever the expected working approach is. I doubt i'm the only one who got tripped up by the change between versions.

nstuyvesant:
Webpack is finicky about resolving paths and it's unfortunately not good at using the modern conventions for package.json exports.

There is a new docs website in a PR under review. I just added the reference for this file to it.

BTW... if you have an option to use vite instead of webpack, I think you'll find it's faster and requires less configuration.

crhuff-ibm:
[...]

PR #1824, opened 2024-05-17
fix: Update package.json style export

crhuff-ibm:
fix the restricted exports provided by package.json to allow scss to import

### Updates
- adds in a more generic ./styles import so A. you don't have to provide an explicit import of a .css file and B. scss can now handle loading that file using @use instead of @import


crhuff-ibm:
Fixes #1819

theiliad (approved):

PR #1834, opened 2024-06-03
chore(angular): linting and add types

nstuyvesant:
### Updates

- Removed explicit types for ChartTabularData that were already established in the BaseChartComponent.
- Updated comments
- Added @experimental to JSDoc comment for Choropleth
- Republished common types so users don't have to import `@carbon/charts` just for types
- Submitting separately from docs PR to break it down into smaller pieces

### DIFF
--- a/packages/angular/src/lib/index.ts
+++ b/packages/angular/src/lib/index.ts
@@ -48,3 +48,35 @@ export {
 	ShapeNodeComponent,
 	ShapeNodeModule
 } from './diagrams'
+
+// Republish essential types from core so it's not a required dependency for TypeScript projects
+export type {
+	ChartTabularData,
+	ChartOptions,
+	AlluvialChartOptions,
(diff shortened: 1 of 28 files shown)


packages/angular/src/lib/index.ts
```diff
@@ -48,3 +48,35 @@ export {
...
+	LineChartOptions,
+	LollipopChartOptions,
+	MeterChartOptions,
+	PieChartOptions,
+	RadarChartOptions,
+	ScatterChartOptions,
+	StackedAreaChartOptions,
+	TreeChartOptions,
+	TreemapChartOptions,
+	WorldCloudChartOptions as WordCloudChartOptions
```
nstuyvesant, line 78: This is non-breaking because WorldCloudChartOptions was not previously exported.

nstuyvesant:
Merged with other PR

PR #1846, opened 2024-06-14
chore(HeatmapChartOptions): include typings for axes on HeatmapChartOptions

lajtmaN:
### Updates

The Heatmap component supports axes (which are also used on the stories in storybook), but the typings for HeatmapCharOptions does not include any axes.

I wasn't sure whether it is enough to just add the `axes` prop, or we instead should extend the `AxisChartOptions` instead of `BaseChartOptions`


lajtmaN:

theiliad (approved):
LGTM

theiliad:
This has already been addressed

Issue #1845, opened 2024-06-14
[Bug]: StackedBarChartOptions is missing from the typescript exports for carbon charts react


chrisdudley:
### Relevant package(s)

- [ ] @carbon/charts
- [ ] @carbon/charts-svelte
- [X] @carbon/charts-react
- [ ] @carbon/charts-angular
- [ ] @carbon/charts-vue
- [ ] Documentation website

### Carbon Charts Version

1.16.4

### Which bundler are you using?

None (UMD)

### What happened and what did you expect to happen?

Converting JS to typescript, and we have converted many of our charts over successfully but hit an issue on StackedBarChart because the options object was not exported - all the other chart Options objects are.

![image](https://github.com/carbon-design-system/carbon-charts/assets/7387884/167daba0-86f2-4f8b-950b-4062382fd434)

I assume this is a simple omission so have submitted a PR to add it alongside the others:
https://github.com/carbon-design-system/carbon-charts/pull/1844

### IBM Application/Team (if relevant)

IBM API Connect

### What priority level would this be in your opinion?

P2 (Medium)

Issue #1841, opened 2024-06-12
Does `@carbon/charts-react` support React 17?


s100:
We use `@carbon/charts-react@1.16.3` and we also use React 17 at present. In the `package.json` for `@carbon/charts-react`, we see:

https://github.com/carbon-design-system/carbon-charts/blob/ca97715d352a589fb69e788174fac7313f8b2426/packages/react/package.json#L48-L53

In the production dependencies, React 18 appears to be the minimum required version. This is causing us peer dependency issues and we are having to specify an override in our `package.json` to get a successful `npm install`. But in the peer dependencies, it seems as if React 16, 17 and 18 are all acceptable. I'm not certain why `react` and `react-dom` are listed as both production and peer dependencies or why the version ranges disagree with one another. I'd recommend eliminating the peer dependencies entirely. And, if CCR supports React 16 and 17, could you please update the production dependencies to reflect this?

s100:
Following #1849, we now see that `@carbon/charts-react@1.16.9` has no React `dependencies`, only `peerDependencies` on React 16, 17 or 18. Thank you.

PR #1849, opened 2024-06-15
fix(all packages): missing options, package.json and lint fixes

nstuyvesant:
### Updates

1. Added `StackedBarChartOptions` to exports for all packages. Closes https://github.com/carbon-design-system/carbon-charts/issues/1845
2. For `@carbon/charts-react` move react and react-dom to devDependencies and make peerDependencies to match `@carbon/react`. Closes https://github.com/carbon-design-system/carbon-charts/issues/1841
3. Similar to item 2 but for Vue (proactive).
4. Minor lint changes to remove errors related to demos (missing props, prop types, const vs. let).
5. More JSDoc comments.
6. Drop D3 as a peerDependency for core as it's inadvertently been getting bundled all along (total JavaScript bundle size is 206.3K gzipped). The size has come down quite a bit due to more efficient imports implemented back in 1.9.x. Benefit: setup process is simpler and the size trade-off is minimal (and again, was already in the bundle).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,15 @@
 All notable changes to this project will be documented in this file. See
 [Conventional Commits](https://conventionalcommits.org) for commit guidelines.
 
+## 1.16.6 (2024-06-15)
+
+**Note:** Version bump only for package @carbon/charts-monorepo
+
+# Change Log
+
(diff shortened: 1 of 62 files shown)


theiliad (approved):
LGTM

Issue #1850, opened 2024-06-15
[Bug]: Redefined variables violating no-shadow rule


nstuyvesant:
### Relevant package(s)

- [X] @carbon/charts
- [ ] @carbon/charts-svelte
- [ ] @carbon/charts-react
- [ ] @carbon/charts-angular
- [ ] @carbon/charts-vue
- [ ] Documentation website

### Carbon Charts Version

1.16.7

### Which bundler are you using?

Vite

### What happened and what did you expect to happen?

Eslint identified 4 bugs in core:

1. packages/core/src/selectedGroups.spec.ts - prefix is imported but then redefined
2. packages/core/src/components/graphs/gauge.ts - value const defined on line 170 then redefined in arrow function on line 214
3. packages/core/src/components/graphs/heatmap.ts - domainIdentifier const defined on line 58 then redefined in arrow function on 116
4. packages/core/src/components/graphs/radar.ts - const data defined on line 54 and arrow function parameter of same name defined on line 97

### Chart data and options (automatically formatted so no need for backticks)

```js
N/A
```

### JavaScript console or build output (if relevant)

```shell
N/A
```

### What priority level would this be in your opinion?

P1 (High)

PR #1851, opened 2024-06-15
fix(1850): redefined variables violated no-shadow rule

nstuyvesant:
### Updates

- Closes https://github.com/carbon-design-system/carbon-charts/issues/1850


theiliad (approved):
LGTM

PR #1868, opened 2024-07-12
feat(search): add search capability to docs site

nstuyvesant:
### Updates

- Adds search capability to docs website
- Bumps node, @eslint/compat, eslint-plugin-jsdoc, eslint-plugin-svelte, svelte-eslint-parser, angular, sass, vitest, styled-components, typedoc, fuse.js


packages/docs/src/routes/Axes.tsx
```diff
@@ -5,8 +5,16 @@ import javascript from 'highlight.js/lib/languages/javascript'
 import { CodeSnippet } from '@carbon/react/es'
 import PageHeader from '../components/PageHeader'
 import ConfigExamples from '../components/ConfigExamples'
+import styled from 'styled-components'
 
-import './Axes.scss'
+// const ExampleImage = styled.img`
+// 	width: 100%;
+// `
```
theiliad, line 12: do we need this?

packages/docs/src/routes/Anatomy.tsx
```diff
@@ -1,7 +1,11 @@
 import { Link } from 'react-router-dom'
 import PageHeader from '../components/PageHeader'
 // import { Grid, Column, OrderedList, ListItem } from '@carbon/react'
-// import './Anatomy.scss'
+// import styled from 'styled-components'
+
+// const DiagramImage = styled.img`
+// 	width: 100%;
+// `
```
theiliad, line 8: do we need this?

theiliad (changes requested):

packages/docs/src/routes/Axes.tsx
nstuyvesant: It's for the commented content that was omitted in order to not conflict with the design folks. I recommend keeping this in case we ever bring the content below back.

packages/docs/src/routes/Anatomy.tsx
nstuyvesant: Same as my last comment. The styles are there to support the content that was commented out. If we ever bring that content back, this CSS will be needed.

PR #1869, opened 2024-07-12
fix(docs): media breakpoints for PageHeaderTitle

nstuyvesant:
### Updates

- Drop redirects file since netlify.toml addresses it
- Scale PageHeaderTitle down for smaller viewport widths


theiliad (approved):
LGTM

PR #1881, opened 2024-08-27
fix(treemap): assign leaf text color based on custom colors

### LINKED ISSUE #1880: [Bug]: Text color of tree map leaves is incorrect when using custom colors
### Relevant package(s)

- [X] @carbon/charts
- [ ] @carbon/charts-svelte
- [ ] @carbon/charts-react
- [ ] @carbon/charts-angular
- [ ] @carbon/charts-vue
- [ ] Documentation website

### Carbon Charts Version

1.15.8

### Which bundler are you using?

Webpack

### What happened and what did you expect to happen?

Expected:

Text color (black or white) for tree map leaf is determined based on the color provided via color scale options.

Actual: 

The text color is not dependent on the provided color scale; the same color can sometimes show black and other times show white.

### Chart data and options (automatically formatted so no need for backticks)

```js
// data
export default [
  {
    name: 'Oceania',
    children: [
      {
        name: 'A',
        value: 800
      },
      {
        name: 'B',
        value: 200
      },
      {
        name: 'C',
        value: 100
      },
      {
        name: 'D',
        value: 900
      }
    ]
  },
  {
    name: 'Europe',
    children: [
      {
        name: 'France',
        value: 2800,
        showLabel: true
      },
      {
        name: 'Germany',
        value: 10000,
        showLabel: true
      },
      {
        name: 'Sweden',
        value: 500
      },
      {
        name: 'England',
        value: 500
      },
      {
        name: 'Italy',
        value: 200
      }
    ]
  },
  {
    name: 'America',
    children: [
      {
        name: 'U.S.',
        value: 3500,
        showLabel: true
      },
      {
        name: 'Brazil',
        value: 3000,
        showLabel: true
      },
      {
        name: 'Mexico',
        value: 2000,
        showLabel: true
      },
      {
        name: 'AA',
        value: 500
      },
      {
        name: 'BB',
        value: 100
      },
      {
        name: 'CC',
        value: 500
      },
      {
        name: 'DD',
        value: 500
      },
      {
        name: 'EE',
        value: 400
      },
      {
        name: 'FF',
        value: 600
      },
      {
        name: 'GG',
        value: 200
      },
      {
        name: 'HH',
        value: 800
      },
      {
        name: 'II',
        value: 900
      },
      {
        name: 'JJ',
        value: 100
      },
      {
        name: 'KK',
        value: 900
      }
    ]
  },
  {
    name: 'Australia',
    children: [
      {
        name: 'KH',
        value: 2000
      },
      {
        name: 'LL',
        value: 400
      },
      {
        name: 'MM',
        value: 500
      },
      {
        name: 'NN',
        value: 100
      },
      {
        name: 'OO',
        value: 1000
      },
      {
        name: 'PP',
        value: 800
      },
      {
        name: 'QQ',
        value: 700
      },
      {
        name: 'RR',
        value: 200
      },
      {
        name: 'ST',
        value: 300
      }
    ]
  },
  {
    name: 'Africa',
    children: [
      {
        name: 'Nigeria',
        value: 2300,
        showLabel: true
      },
      {
        name: 'TT',
        value: 2000
      },
      {
        name: 'UU',
        value: 500
      },
      {
        name: 'VV',
        value: 1200
      },
      {
        name: 'WW',
        value: 2000
      },
      {
        name: 'XX',
        value: 800
      },
      {
        name: 'YY',
        value: 2000
      },
      {
        name: 'ZZ',
        value: 500
      },
      {
        name: 'ABC',
        value: 1200
      },
      {
        name: 'CBA',
        value: 1500
      }
    ]
  },
  {
    name: 'Asia',
    children: [
      {
        name: 'China',
        value: 12500,
        showLabel: true
      },
      {
        name: 'Iran',
        value: 22500,
        showLabel: true
      },
      {
        name: 'Myanmar',
        value: 12500,
        showLabel: true
      }
    ]
  }
]

// options
export default {
  title: 'Treemap',
  height: '600px',
  data: {
    groupMapsTo: 'name',
  },
  color: {
    scale: {
      Oceania: 'green',
      Europe: 'green',
      America: 'green',
      Australia: 'green',
      Africa: 'green',
      Asia: 'green',
    },
  },
};
```

### JavaScript console or build output (if relevant)

_No response_

### StackBlitz repro

 https://stackblitz.com/edit/zj5ef7?file=options.js,index.js

### IBM Application/Team (if relevant)

_No response_

P3 (Low)

SMassola:
- updated treemap chart to set the background color in data for the leaf text node so it can be referenced when determining the text color

Fix #1880

https://github.com/user-attachments/assets/296a09aa-3085-4e9f-a07a-24308498f0d1

### DIFF
--- a/packages/core/src/components/graphs/treemap.ts
+++ b/packages/core/src/components/graphs/treemap.ts
@@ -25,14 +25,15 @@ const findColorShade = (hex: string) => {
 	return null
 }
 
-const textFillColor = function () {
-	const correspondingLeaf = select(this.parentNode).select('rect.leaf') as any
-	const correspondingLeafFill = getComputedStyle(correspondingLeaf.node(), null).getPropertyValue(
-		'fill'
-	)
-	const cl = d3Color(correspondingLeafFill) as any
+const textFillColor = function (data) {
+	const correspondingLeaf = select(this.parentNode).select('rect.leaf')
+	const correspondingLeafFill: string =
+		data.backgroundColor ??
+		getComputedStyle(correspondingLeaf.node() as Element, null).getPropertyValue('fill')
+	const cl = d3Color(correspondingLeafFill)
 
 	let colorShade: any
+
 	if (cl) {
 		colorShade = findColorShade(cl ? cl.hex() : null)
 	}
@@ -199,7 +200,7 @@ export class Treemap extends Component {
 					return [
 						{
 							text: d.data.name,
-							color: color.l < 0.5 ? 'white' : 'black'
+							backgroundColor: color
 						}
 					]
 				},
--- a/packages/core/src/components/graphs/treemap.ts
+++ b/packages/core/src/components/graphs/treemap.ts
@@ -196,11 +196,11 @@ export class Treemap extends Component {
 
 					let parent = d
 					while (parent.depth > 1) parent = parent.parent
-					const color = hsl(this.model.getFillColor(parent.data.name))
+
 					return [
 						{
 							text: d.data.name,
-							backgroundColor: color
+							backgroundColor: this.model.getFillColor(parent.data.name)
 						}
 					]
 				},
--- a/packages/docs/src/lib/treemap/index.ts
+++ b/packages/docs/src/lib/treemap/index.ts
@@ -16,6 +16,21 @@ const options: TreemapChartOptions = {
 	height: '600px'
 }
 
+const customColorOptions: TreemapChartOptions = {
+	title: 'Treemap (Custom colors)',
+	height: '600px',
+	color: {
+		scale: {
+			Oceania: '#d9fbfb',
+			Europe: '#9ef0f0',
+			America: '#3ddbd9',
+			Australia: '#08bdba',
+			Africa: '#009d9a',
+			Asia: '#081a1c'
+		}
+	}
+}
+
 const data: ChartTabularData = [
 	{
 		name: 'Oceania',
@@ -110,6 +125,11 @@ export const examples: Example[] = [
 	{
 		data,
 		options,
+		tags: ['test']
+	},
+	{
+		data,
+		options: customColorOptions,
 		tags: ['test']
 	}
 ]
--- a/packages/docs/src/lib/treemap/index.ts
+++ b/packages/docs/src/lib/treemap/index.ts
@@ -21,12 +21,12 @@ const customColorOptions: TreemapChartOptions = {
 	height: '600px',
 	color: {
 		scale: {
-			Oceania: '#d9fbfb',
-			Europe: '#9ef0f0',
-			America: '#3ddbd9',
-			Australia: '#08bdba',
-			Africa: '#009d9a',
-			Asia: '#081a1c'
+			Oceania: '#3ddbd9',
+			Europe: '#08bdba',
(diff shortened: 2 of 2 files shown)


packages/core/src/components/graphs/treemap.ts
```diff
@@ -199,7 +200,7 @@ export class Treemap extends Component {
 					return [
 						{
 							text: d.data.name,
-							color: color.l < 0.5 ? 'white' : 'black'
+							backgroundColor: color
```
theiliad, line 203: seems like you removed the `color` field here. is there a reason for that?

theiliad (changes requested):
Thanks @SMassola, could you pls add a demo as well in the demo data so that this example is documented?

packages/core/src/components/graphs/treemap.ts
SMassola: The only reason I removed it was because it didn't seem to be utilized anywhere; if this is not the case I can revert this change.

theiliad: that sets the text fill based on backgroundColor

theiliad (changes requested):

SMassola:
> Thanks @SMassola, could you pls add a demo as well in the demo data so that this example is documented?

I added an example to the docs which uses custom colors and appended a short demo video to the PR description.

packages/core/src/components/graphs/treemap.ts
SMassola: I think the `color` property within the text node's `data` wasn't being used; instead the text node's color is set by the `textFillColor` function which utilizes either the new `backgroundColor` property that I've added or the fill color of the parent `rect.leaf` node.

The main issue was that when the `textFillColor` called `getComputedStyles` on the `rect.leaf` node, the `fill` property isn't set to the custom color early enough, so it always ended up determining the text color based on the default colors.

theiliad: yes @SMassola the text color values seem to be getting set properly in cases where you're providing custom colors, but if you don't, it seems like your PR has broken that case

SMassola: > yes @SMassola the text color values seem to be getting set properly in cases where you're providing custom colors, but if you don't, it seems like your PR has broken that case

You're right, I've modified the code to send the `backgroundColor` in the proper format (simply a string) to the `textFillColor` function.

theiliad (changes requested):
Thanks @SMassola, just a final point, could you pls use a set of colors that offer larger contrast in your `custom colors` demo?

image

some of these colors wouldn't pass a 3:1 test

SMassola:
> Thanks @SMassola, just a final point, could you pls use a set of colors that offer larger contrast in your `custom colors` demo?
> 
> image
> some of these colors wouldn't pass a 3:1 test

I updated the chart colors to use teal 30-80 from the [continuous monochromatic color palette](https://carbondesignsystem.com/data-visualization/color-palettes/#monochromatic) (option 4).

SMassola:
@theiliad this is what teal 30-80 looks like; let me know if this range of colors is appropriate:

Screenshot 2024-08-29 at 9 25 21 AM

PR #1888, opened 2024-09-08
fix(release): remove auth-token from actions/setup-node@v4 and use env for publish

nstuyvesant:
### Updates

- Remove auth-token from actions/setup-node@v4 in release workflow
- Set env variable NODE_AUTH_TOKEN to secrets.NPM_TOKEN for `npx lerna publish`
- Bump concurrently dependency


theiliad (approved):
LGTM

PR #1928, opened 2024-12-06
fix: exclude jsx-runtime from bundle

### LINKED ISSUE #1927: [Bug]: jsx-runtime included in bundle
### Relevant package(s)

- [ ] @carbon/charts
- [ ] @carbon/charts-svelte
- [x] @carbon/charts-react
- [ ] @carbon/charts-angular
- [ ] @carbon/charts-vue
- [ ] Documentation website

### Carbon Charts Version

1.22.5

### Which bundler are you using?

Webpack

### What happened and what did you expect to happen?

The final bundle of your package includes the jsx-runtime.  This should be excluded from the bundle to allow users to upgrade to React v19

Image

### Chart data and options (automatically formatted so no need for backticks)

```js

```

### JavaScript console or build output (if relevant)

```shell

```

### StackBlitz repro

_No response_

### IBM Application/Team (if relevant)

_No response_

P1 (High)

a88zach:
### Updates

Exclude react/jsx-runtime from the final bundle.  This should be excluded to allow users to update to React v19.

Including this dependency in the bundle results in 
`"TypeError: Cannot read properties of undefined (reading 'ReactCurrentDispatcher')"`

fixes #1927

### DIFF
--- a/packages/react/package.json
+++ b/packages/react/package.json
@@ -43,8 +43,8 @@
 		"@ibm/telemetry-js": "^1.8.0"
 	},
 	"peerDependencies": {
-		"react": "17.0.1 || ^18.2.0",
-		"react-dom": "^17.0.1 || ^18.2.0"
+		"react": "^16.8.6 || ^17.0.1 || ^18.2.0",
+		"react-dom": "^16.8.6 || ^17.0.1 || ^18.2.0"
 	},
 	"devDependencies": {
 		"@playwright/test": "^1.49.0",
--- a/packages/react/package.json
+++ b/packages/react/package.json
@@ -43,8 +43,8 @@
 		"@ibm/telemetry-js": "^1.8.0"
 	},
 	"peerDependencies": {
-		"react": "^16.8.6 || ^17.0.1 || ^18.2.0",
-		"react-dom": "^16.8.6 || ^17.0.1 || ^18.2.0"
+		"react": "17.0.1 || ^18.2.0",
+		"react-dom": "^17.0.1 || ^18.2.0"
 	},
 	"devDependencies": {
 		"@playwright/test": "^1.49.0",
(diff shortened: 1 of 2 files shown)


theiliad:
I don't think `'react/jsx-runtime'` can just be dropped, that could break support for react17 or 18

a88zach:
> I don't think `'react/jsx-runtime'` can just be dropped, that could break support for react17 or 18

This doesn't necessarily drop it, it just excludes it from the bundle.  Then instead of the 3rd party package being bundled, it is required at the top of the bundle.  This way the version used matches the version installed by the user which changes between React 18 and 19

theiliad:
> > I don't think `'react/jsx-runtime'` can just be dropped, that could break support for react17 or 18
> 
> This doesn't necessarily drop it, it just excludes it from the bundle. Then instead of the 3rd party package being bundled, it is required at the top of the bundle. This way the version used matches the version installed by the user which changes between React 18 and 19
> 
> It's the same reason you are already excluding react and react-dom. You want the code to use the installed version and not an explicit version

Are you having these issues with the UMD bundle? or with the ESM?

a88zach:
I'm using NextJs, so it's using the ESM bundle, but it would also be a problem when using the UMD bundle.  Although not officially supported yet by this library, if you try to use the library with React 19, you'll see the issue.  If you try to search the bundles for `react-jsx-runtime`, you can see that it gets bundled.  

React dependencies should be peer dependencies and not bundled

dabrad26 (approved):
@theiliad any update here? This seems logical to require peer deps and this is stopping successful migration to React 19.  I confirmed no issue with React 19 and the main Carbon library.  The issue is only with the charts package.  

```
    TypeError: Cannot read properties of undefined (reading 'ReactCurrentDispatcher')

      1 | import React from 'react';
    > 3 | import {GaugeChart} from '@carbon/charts-react';
```

theiliad (approved):
LG in my testing

PR #1937, opened 2025-01-07
fix(vite): only externalize jsx-runtime

theiliad:


dabrad26:
I am seeing issues with latest carbon charts.  Seems it is bundling React as well?  I feel we should not bundle react or any of the peer dependencies?

```
    A React Element from an older version of React was rendered. This is not supported. It can happen if:
    - Multiple copies of the "react" package is used.
    - A library pre-bundled an old copy of "react" or "react/jsx-runtime".
    - A compiler tries to "inline" JSX instead of using the runtime.
```

PR #1942, opened 2025-01-11
update angular to v19 and move to standalone components

tom-wegener:
Currently the charts are not compatible with angular v19 because they are not standalone and not explicitly marked as not standalone.

### Updates

- make components standalone
- make example app standalone
- update angular to v19

There were no issues for that. I'm happy to create one.

no visual changes.

If there is anything you want to have changed, please feel free to say so. My yarn couldn't find the commit-script, so I just used git commit.

I left the ngmodules files in, I would personally also remove them but I wasn't sure if they are still required for some legacy functionality.


tom-wegener:

theiliad:
Hi @tom-wegener, there seems to be issues with the alluvial chart rendering

image

and some others if you scroll down further
https://deploy-preview-1942--carbon-charts-angular.netlify.app

which do look ok in the vanilla test harness
https://deploy-preview-1942--carbon-charts-core.netlify.app

tom-wegener:
Hey @theiliad 

Yes, I accidentally committed three wrong lines where I tested signals for the base component. Thank you for pointing it out.

When checking my commit again I also found some old ngswitchcase and ngif in the card-node and shape-node which I could change but I could not find a test for that. Where could I find one?

And is there a need for the *.modules.ts-files?

packages/angular/src/lib/charts/base-chart.component.ts
```diff
@@ -46,16 +46,16 @@ export class BaseChartComponent implements AfterViewInit, OnInit {
 	get options() {
 		return this._options
 	}
-
+	
 	/**
 	 * Chart width
 	 */
-	@Input() width?: string
+	@Input() width?: string;
```
nstuyvesant, line 53: Super minor but curious why the semicolon was added given the prettier settings. Same for line 58.

packages/angular/src/lib/charts/charts.module.ts
```diff
@@ -32,7 +31,9 @@ import {
 } from './index'
 
 @NgModule({
-	declarations: [
+	declarations: [],
+	imports: [
+		WordCloudChartComponent,
```
nstuyvesant, line 36: Odd the list got reordered. Perhaps the components should be alphabetized?

packages/angular/package.json
```diff
@@ -23,29 +23,29 @@
...
+		"@angular/platform-browser-dynamic": "^19.0.6",
 		"@carbon/icons": "^11.53.0",
 		"@playwright/test": "^1.49.0",
 		"angular-eslint": "^18.4.2",
 		"concurrently": "^9.1.0",
 		"eslint": "^9.15.0",
-		"ng-packagr": "^18.2.1",
+		"ng-packagr": "^19.0.1",
 		"prettier": "^3.4.1",
 		"typescript": "<5.6.0",
```
nstuyvesant, line 50: Angular 19 now supports TypeScript 5.6 so please change to ">=5.5.0 <5.7.0" per https://angular.dev/reference/versions

packages/angular/src/lib/charts/charts.module.ts
tom-wegener: I've ordered the list alphabetically, there were a few more out of order.

packages/angular/src/lib/charts/base-chart.component.ts
tom-wegener: Fair, I've run prettier again and not it is fixed.

nstuyvesant:
> Hey @theiliad
> And is there a need for the *.modules.ts-files?

Regarding the module.ts files, with standalone, not really needed but the documentation package in this monorepo would have to be revised (slightly) if they were removed. For example, packages/docs/src/lib/stackblitz/angular.ts includes `import { ChartsModule } from '@carbon/charts-angular'` for the generated StackBlitz examples. It might be beneficial to change that file anyway to update it to use Angular 19 instead of 18. You would have to check the specific version of 19 that StackBlitz supports as they aren't always on top of the latest.

tom-wegener:
> > Hey @theiliad
> > And is there a need for the *.modules.ts-files?
> 
> Regarding the module.ts files, with standalone, not really needed but the documentation package in this monorepo would have to be revised (slightly) if they were removed. For example, packages/docs/src/lib/stackblitz/angular.ts includes `import { ChartsModule } from '@carbon/charts-angular'` for the generated StackBlitz examples. It might be beneficial to change that file anyway to update it to use Angular 19 instead of 18. You would have to check the specific version of 19 that StackBlitz supports as they aren't always on top of the latest.

Thank you, I will look into it. Is there an easy way to test it besides starting the complete documentation-site?

tom-wegener:
Thank you. How do I import the local angular package?

nstuyvesant:
> Thank you. How do I import the local angular package?

Sorry if I am misunderstanding but if you are asking how to use @carbon/charts-angular from your local clone rather than the one published on npmjs...

- Connect to the directory where you cloned the monorepo  and go to packages/angular
- `npm link` makes it globally available on your local machine
- Navigate to the root directory of your other project
- npm link @carbon/charts-angular

tom-wegener:
> > Thank you. How do I import the local angular package?
> 
> Sorry if I am misunderstanding but if you are asking how to use @carbon/charts-angular from your local clone rather than the one published on npmjs...
> 
>     * Connect to the directory where you cloned the monorepo  and go to packages/angular
> 
>     * `npm link` makes it globally available on your local machine
> 
>     * Navigate to the root directory of your other project
> 
>     * npm link @carbon/charts-angular

I'm sorry my question was not precise enough. I'm unsure how to use the local version of the angular package with the stackblitz examples if I'm starting the docs site via `yarn dev`. It would be easy to test it if the package was published but at the moment I only have the local version.

nstuyvesant:
> I'm sorry my question was not precise enough. I'm unsure how to use the local version of the angular package with the stackblitz examples if I'm starting the docs site via `yarn dev`. It would be easy to test it if the package was published but at the moment I only have the local version.

Oh, sorry. Now, I understand. 

This blog post suggests StackBlitz supports tarball dependencies... https://blog.stackblitz.com/posts/announcing-native-package-manager-support/

In theory, you could use `npm pack` to generate a tarball then either put that tarball on a publicly accessible website or upload to StackBlitz then do this on its package.json...
```json
{
  "dependencies": {
    "@carbon/charts-angular": "file:./charts-angular-5.6.0.tgz"
  }
}```

I never tried this myself so your mileage may vary.

tom-wegener:
Then I would propose that I drop the last two commits from this PR, we get this merged and published and I'll open a new one and test the changes against the new package?

nstuyvesant:
> Then I would propose that I drop the last two commits from this PR, we get this merged and published and I'll open a new one and test the changes against the new package?

I think that makes sense. It will be a shorter path.

nstuyvesant:
@theiliad - LGTM.

theiliad (approved):

theiliad:
Thank you @tom-wegener!

Issue #1956, opened 2025-02-12
[Bug]: Uncaught TypeError: Cannot read properties of undefined (reading 'A')


davidebibm:
### Relevant package(s)

- [ ] @carbon/charts
- [ ] @carbon/charts-svelte
- [x] @carbon/charts-react
- [ ] @carbon/charts-angular
- [ ] @carbon/charts-vue
- [ ] Documentation website

### Carbon Charts Version

1.22.16

### Which bundler are you using?

Webpack

### What happened and what did you expect to happen?

With last upgrade from 1.22.15 to 1.22.16, we are getting the following error on any chart.
Was working on version 1.22.15

### Chart data and options (automatically formatted so no need for backticks)

```js
All charts no matter options or data passed.
```

### JavaScript console or build output (if relevant)

```shell
Uncaught TypeError: Cannot read properties of undefined (reading 'A')
    at c (index.mjs:22076:17)
    at f (index.mjs:22172:27)
    at ki.jsx (index.mjs:22219:14)
    at aU.render (index.mjs:22251:31)
    at finishClassComponent (react-dom.development.js:19776:31)
    at updateClassComponent (react-dom.development.js:19722:24)
    at beginWork (react-dom.development.js:21645:16)
    at HTMLUnknownElement.callCallback (react-dom.development.js:4164:14)
    at Object.invokeGuardedCallbackDev (react-dom.development.js:4213:16)
    at invokeGuardedCallback (react-dom.development.js:4277:31)
c @ index.mjs:22076
f @ index.mjs:22172
ki.jsx @ index.mjs:22219
render @ index.mjs:22251
finishClassComponent @ react-dom.development.js:19776
updateClassComponent @ react-dom.development.js:19722
beginWork @ react-dom.development.js:21645
callCallback @ react-dom.development.js:4164
invokeGuardedCallbackDev @ react-dom.development.js:4213
invokeGuardedCallback @ react-dom.development.js:4277
beginWork$1 @ react-dom.development.js:27485
performUnitOfWork @ react-dom.development.js:26594
workLoopSync @ react-dom.development.js:26500
renderRootSync @ react-dom.development.js:26468
performConcurrentWorkOnRoot @ react-dom.development.js:25772
workLoop @ scheduler.development.js:266
flushWork @ scheduler.development.js:239
performWorkUntilDeadline @ scheduler.development.js:533Understand this errorAI
[...]

PR #1958, opened 2025-02-13
Resolve compatibility issues: docs website, @carbon/charts-react with jsx-runtime, @carbon/charts-svelte for non-TypeScript users

nstuyvesant:
### Updates

- Used yarn resolutions to upgrade `react-is` in order to overcome `@carbon/react` issue with React 19
- Removed JSX from `@carbon/charts-react` to eliminate dependency on `react/jsx-runtime` for compatibility reasons (specifically React 16 does not use jsx-runtime so externalizing it breaks 16 and bundling the jsx-runtime breaks everything but React 19)
- On docs site, fixed issues with div tags inside of p tags (React 19 complains about it)
- Ensure `@carbon/charts-svelte` emits JavaScript rather than TypeScript components to improve compatbility
- Updated search index for docs website
- Updated Playwright test baselines
- Bumped dependencies within semver


packages/docs/src/routes/Area.tsx
```diff
@@ -9,14 +8,6 @@ import { chartTypesStacked, examplesStacked } from '../lib/area/examplesStacked'
 import '@carbon/charts-react/styles.css'
 
 export default function Area() {
-	const [selectedIndex, setSelectedIndex] = useState(0)
```
nstuyvesant, line 12: Unused code.

packages/docs/src/routes/Area.tsx
```diff
@@ -59,7 +50,7 @@ export default function Area() {
 
 			<StackBlitzLauncherExplanation />
 
-			<Tabs selectedIndex={selectedIndex} onChange={handleTabChange}>
```
nstuyvesant, line 62: Unused code

packages/docs/src/routes/Area.tsx
```diff
@@ -68,10 +59,10 @@ export default function Area() {
 				<TabPanels>
 					<TabPanel>
 						{examples.map((example, index) => (
-							<p key={index} className="chart">
+							<div key={index} className="chart">
```
nstuyvesant, line 62: React 19 complains if you have paragraphs that contain divs (StackBlitzLauncher)

packages/docs/src/routes/Area.tsx
```diff
@@ -81,10 +72,10 @@ export default function Area() {
 						</p>
 
 						{examplesStacked.map((example, index) => (
-							<p key={index} className="chart">
+							<div key={index} className="chart">
```
nstuyvesant, line 75: React 19 complains if you have paragraphs that contain divs (StackBlitzLauncher)

packages/docs/src/routes/Bar.tsx
```diff
@@ -1,4 +1,3 @@
-import { useState, useEffect } from 'react'
```
nstuyvesant, line 1: Inert code

packages/docs/src/routes/Bar.tsx
```diff
@@ -11,14 +10,6 @@ import { chartTypesStacked, examplesStacked } from '../lib/bar/examplesStacked'
 import '@carbon/charts-react/styles.css'
 
 export default function Bar() {
-	const [selectedIndex, setSelectedIndex] = useState(0)
```
nstuyvesant, line 14: Inert code

packages/docs/src/routes/Bar.tsx
```diff
@@ -39,7 +30,7 @@ export default function Bar() {
 
 			<StackBlitzLauncherExplanation />
 
-			<Tabs selectedIndex={selectedIndex} onChange={handleTabChange}>
```
nstuyvesant, line 42: Inert code

packages/react/src/charts/AlluvialChart.ts
nstuyvesant, line 1: All chart components other than BaseChart.tsx were pure TypeScript already. Just needed to rename them.

packages/react/src/charts/BaseChart.ts
```diff
@@ -38,6 +38,6 @@ export default class BaseChart<
 	}
 
 	render() {
-		return <div ref={this.chartRef} className="chart-holder"></div>
+		return React.createElement('div', { ref: this.chartRef, className: 'chart-holder' })
```
nstuyvesant, line 41: One-liner to not use jsx-runtime that was introduced in React 17.

packages/react/src/diagrams/CardNode.ts
```diff
@@ -26,15 +26,16 @@ const CardNode: React.FC<
...
 
-	return (
-		<Component
-			className={cardClasses}
-			style={{ borderColor: color, position }}
-			tabIndex={0}
-			{...rest}>
-			{children}
-		</Component>
+	return React.createElement(
```
nstuyvesant, line 29: Minor changes to diagrams to remove dependency on jsx-runtime.

packages/react/vite.config.ts
```diff
@@ -12,7 +12,7 @@ export default defineConfig({
 			fileName: format => `index.${format === 'es' ? 'm' : 'umd.c'}js`
 		},
 		rollupOptions: {
-			external: ['react', 'react-dom'],
+			external: ['react', 'react-dom' /*, 'react-dom/client', 'react/jsx-runtime'*/],
```
nstuyvesant, line 15: Leaving these in as comments in case we need to use them in conjunction with remove of JSX dependencies.

packages/svelte/README.md
```diff
@@ -42,6 +42,8 @@ yarn add -D @carbon/charts-svelte
 
 The required styles should be imported from `@carbon/charts-svelte/styles.css`.
 
+A release in the near future will move components to runes mode for Svelte 5 and beyond. As this would be a breaking change because event handlers will change to component properties, this version will be published with the distribution tag `next`. Please check <a href="https://www.npmjs.com/package/@carbon/charts-svelte?activeTab=versions">here</a> to see if a version for `latest` has been published.
```
nstuyvesant, line 45: As a future version that uses runes mode would normally be a breaking change, we can use `next` as the distribution tag so only users who want it will access the newer version.

packages/svelte/src/lib/AlluvialChart.svelte
```diff
@@ -1,4 +1,4 @@
-<script lang="ts">
+<script lang="typescript">
```
nstuyvesant, line 1: Switched from vitePreprocess to sveltePreprocess to ensure output is JavaScript. The latter requires the lang attribute to be `typescript` instead of `ts`

packages/svelte/svelte.config.js
```diff
@@ -1,9 +1,9 @@
 import adapter from '@sveltejs/adapter-static'
-import { vitePreprocess } from '@sveltejs/vite-plugin-svelte'
+import { sveltePreprocess } from 'svelte-preprocess'
```
nstuyvesant, line 2: The former was emitting TypeScript Svelte components instead of JavaScript. This ensures the output is JavaScript to maximize compatibility.

packages/angular/tests/dom-regression.spec.ts-snapshots/charts-chromium-darwin.html
nstuyvesant, line 1: New test baseline

AlanGreene:
@nstuyvesant It looks like this could potentially resolve https://github.com/carbon-design-system/carbon-charts/issues/1956 if I'm understanding the changes.
Is that correct?

nstuyvesant:
> @nstuyvesant It looks like this could potentially resolve #1956 if I'm understanding the changes. Is that correct?

Funny... I just posted to the Issue a second ago the rationale for the removal of JSX in this PR to improve compatibility across React 16 to 19. Yes, that was the intention behind this PR (as well as docs website and @carbon/charts-svelte compatibility).

theiliad (approved):

PR #2014, opened 2025-07-17
WIP: feat(radar): add possibility to display all spoke labels

max-jonas:


theiliad:
Hi @max-jonas, could you pls add an example of this to the demo data? that way it's show up in our demos

PR #2032, opened 2025-09-16
Updated public api, added exports for Legend/Roles enums and tools.

vrg84:
#### Core
- Exposed enums `LegendEvent` and `Roles`
- Exposed helper functions from `/packages/core/src/tools.ts`.

### DIFF
--- a/packages/core/CHANGELOG.md
+++ b/packages/core/CHANGELOG.md
@@ -1,4 +1,11 @@
 # Change Log
+## [Unreleased]
+
+Updated public api.
+- Exposed enums `LegendEvent` and `Roles`
+- Exposed helper functions from `/packages/core/src/tools.ts`.
+
+# Change Log
 
 All notable changes to this project will be documented in this file. See
 [Conventional Commits](https://conventionalcommits.org) for commit guidelines.
--- a/packages/core/src/index.ts
+++ b/packages/core/src/index.ts
@@ -148,6 +148,7 @@ export {
 	LayoutAlignItems,
 	LayoutDirection,
 	LayoutGrowth,
+	LegendEvent,
 	LegendItemType,
 	LegendOrientations,
 	LegendPositions,
@@ -158,6 +159,7 @@ export {
 	Projection,
 	RadarEvent,
 	RenderTypes,
+	Roles,
 	ScaleTypes,
 	ScatterEvent,
 	Skeletons,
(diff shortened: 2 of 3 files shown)


vrg84:
Recheck

VictorGerritsen:
> > The library doesn’t provide a grouped + stacked bar chart. Since we need this feature, we extended `Bar` and followed the approach in `bar-grouped.ts` to a version that suits our needs.
> 
> @vrg84 why not contribute the grouped stacked bar chart instead?

I would if i could, i just dont think i have the time to take it to where it needs to be. The version we are using right now is a bare minimum to make it work within our apps.

vrg84:
> > The library doesn’t provide a grouped + stacked bar chart. Since we need this feature, we extended `Bar` and followed the approach in `bar-grouped.ts` to a version that suits our needs.
> 
> @vrg84 why not contribute the grouped stacked bar chart instead?

I wish i could, i just dont have the time right now take it to where it needs to be. Its at a bare minimum to fit our needs. 

Having said that, i do realize the exposed tools are only ever used in creating a new chart / extending the functionality... i'll see if i can make time to create a proper issue for it and start/share the work, maybe more people can contribute.

vrg84:
> > > > The library doesn’t provide a grouped + stacked bar chart. Since we need this feature, we extended `Bar` and followed the approach in `bar-grouped.ts` to a version that suits our needs.
> > > 
> > > 
> > > @vrg84 why not contribute the grouped stacked bar chart instead?
> > 
> > 
> > I wish i could, i just dont have the time right now take it to where it needs to be. Its at a bare minimum to fit our needs.
> > Having said that, i do realize the exposed tools are only ever used in creating a new chart / extending the functionality... i'll see if i can make time to create a proper issue for it instead and start/share the work, maybe more people can contribute.
> 
> would you like to share your current code that achieves the grouped+stacked bar functionality? and maybe I could get it into the library, no promises though 🙂

Yes! I have no expectations since it seems like theres still a lot of work to do, but I ported it into my fork:
https://github.com/vrg84/carbon-charts/tree/grouped-stacked-bar-chart

I've taken some steps to make it run with a simple example (located in 
https://github.com/vrg84/carbon-charts/blob/grouped-stacked-bar-chart/packages/docs/src/lib/bar/examplesGroupedStacked.ts

image

But like i said, this implementation was tailored to our needs (it was the best i can do in a few days), this is still missing from the top of my head:
- Extended tooltip, its not clear now which "stack" you are looking at. We've customized it with the customHTML tooltip but i guess there should be a default for it
- Table view has not been implemented
- CSV export has not been implemented
- Only the vertical orientation works like it should
- Only the labels scaleType has been tested

It was fun diving into the library and d3.js for a bit 😄thanks for your time!

Issue #2081, opened 2026-04-20
[Enhancement]: Set the maximum score for radar chart


mo-alaa:
### Application/Team

CSR

### Summary

Request an enhancement to allow consumers of the Radar chart to explicitly set a maximum score value, instead of relying on the auto-calculated maximum. This is a development enhancement.

<img width="518" height="578" alt="Image" src="https://github.com/user-attachments/assets/bbd7d2be-0318-4dca-93b5-228aebfcad8a" />

### Justification

Auto-calculated maximum scores can cause inconsistencies when comparing results across users or time periods. A fixed, configurable maximum improves consistency, interpretability, and aligns better with structured assessment and learning frameworks.

### Desired UX and success metrics

Consumers of the Radar chart can optionally provide a max score value via configuration. When set, all chart axes consistently use this value.

Success metrics:
- Consistent visual scaling across users and datasets

### "Must have" functionality

- Ability to explicitly set a maximum score for the Radar chart
- Backward compatibility: default behavior remains auto-calculated
- Support for common numeric scales (e.g., 5, 10, 100)

### Specific timeline issues / requests

This enhancement would be valuable to have when capacity allows, particularly to reduce custom workarounds in upcoming assessment-related features.

### Available extra resources

No dedicated development resources are available at this time.

### What priority level would this issue be in your opinion?

P1

PR #2082, opened 2026-04-20
feat(radar): add custom maxValue option

mo-alaa:
## Issue
https://github.com/carbon-design-system/carbon-charts/issues/2081

## Summary
Adds a custom `maxValue` option to radar charts, allowing users to set a fixed maximum scale value instead of using the auto-calculated maximum from data.

## Changes Made
- Added `maxValue?: number` option to `RadarChartOptions` interface
- Updated radar chart configuration with default `maxValue: undefined`
- Modified radar rendering logic to use custom max when provided
- Added demo example "Radar - Custom Max Score (100)"
- Updated CHANGELOG.md

## Motivation
This feature enables:
- Standardized comparisons across multiple radar charts
- Fixed scoring systems (e.g., 0-100 scale)
- Consistent visualization regardless of data range

## Usage Example
```typescript
const options = {
  radar: {
    axes: { angle: 'feature', value: 'score' },
    maxValue: 100  // Custom max score
  }
}
```
### Demo screenshot
image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,17 @@
 All notable changes to this project will be documented in this file. See
 [Conventional Commits](https://conventionalcommits.org) for commit guidelines.
 
+## [Unreleased]
+
+### Features
+
+- **radar:** add custom `maxValue` option to allow setting a fixed maximum scale value instead of auto-calculated max from data
+
+# Change Log
+
+All notable changes to this project will be documented in this file. See
+[Conventional Commits](https://conventionalcommits.org) for commit guidelines.
+
 ## 1.27.6 (2026-04-16)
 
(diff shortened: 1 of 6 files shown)


amal-k-joy (commented):
Thanks for the contribution . 
Can you sign the DCO?

amal-k-joy (approved):

mo-alaa:

szinta (commented):
Good work. Just a small observation, consider adding a validation when `maxValue` is less than the actual data maximum. Currently, if a user sets `maxValue: 50` but data contains values like `92`, the chart may render incorrectly with clipped data points.

mo-alaa:
@szinta Good catch, thank you, I've just added a validation for that.

szinta (approved):

PR #2109, opened 2026-07-13
fix: tooltip not showing for truncated labels in single-series charts

Triparna-146:
Closes #2108 

### Updates

- Fixed tooltip not showing for truncated labels in single-series charts
- Fixed keyboard space bar allowing toggle of single-series charts (preventing empty chart state)
- Added `.single-series` CSS class to legend container
- Added `.has-truncated-label` CSS class to truncated legend items
- Pointer cursor now indicates tooltip availability for truncated labels

image

### DIFF
--- a/packages/core/src/components/essentials/legend.ts
+++ b/packages/core/src/components/essentials/legend.ts
@@ -62,7 +62,9 @@ export class Legend extends Component {
 		})
 
 		const legendClickable = getProperty(this.getOptions(), 'legend', 'clickable')
+		const isSingleSeries = dataGroups.length === 1
 		svg.classed('clickable', legendClickable && dataGroups.length > 1)
+		svg.classed('single-series', isSingleSeries)
 
 		const checkboxRadius = legendConfigs.checkbox.radius
 
@@ -164,9 +166,7 @@ export class Legend extends Component {
 		// Remove old elements as needed.
 		legendItems.exit().on('mouseover', null).on('click', null).on('mouseout', null).remove()
 
-		if (legendClickable && addedLegendItems.size() > 1) {
-			this.addEventListeners()
-		}
+		this.addEventListeners()
 	}
 
 	sortDataGroups(dataGroups: any, legendOrder: any) {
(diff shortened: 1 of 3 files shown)


Triparna-146:

Triparna-146:
recheck

packages/core/src/components/essentials/legend.ts
```diff
@@ -160,9 +162,7 @@ export class Legend extends Component {
 		// Remove old elements as needed.
 		legendItems.exit().on('mouseover', null).on('click', null).on('mouseout', null).remove()
 
-		if (legendClickable && addedLegendItems.size() > 1) {
-			this.addEventListeners()
-		}
+		this.addEventListeners()
```
amal-k-joy, line 165: Let’s not remove the guard entirely, since we don’t need to register all events in every scenario. Instead, we can make it more targeted and apply it only for this specific condition.

Something along these lines:

```js
if (legendClickable && dataGroups.length > 1) {
  this.addEventListeners();
} else {
  this.addTooltipListeners(); // only mouseover/mouseout for tooltip thats suits your scenaio
}
```

packages/core/src/components/essentials/legend.ts
```diff
@@ -396,6 +407,11 @@ export class Legend extends Component {
 				}
 			})
 			.on('click', function () {
+				// Prevent toggling data visibility for single series charts
```
amal-k-joy, line 410: can we remove this check , as we have guarded the event registration already?

packages/core/src/components/essentials/legend.ts
```diff
@@ -437,7 +453,10 @@ export class Legend extends Component {
 			.on('keydown', function (event: KeyboardEvent, d: any) {
 				if (event.key && event.key === ' ') {
 					event.preventDefault()
-					self.model.toggleDataLabel(d.name)
+					// Only allow toggling for multi-series charts
```
amal-k-joy, line 456: can we remove this as well as keydown wouldn't be registered?

amal-k-joy (changes requested):
Thanks @Triparna-146  for the contribution.
Everything looks good. 
Added few comments

packages/core/src/components/essentials/legend.ts
```diff
@@ -396,6 +407,11 @@ export class Legend extends Component {
 				}
 			})
 			.on('click', function () {
+				// Prevent toggling data visibility for single series charts
+				if (isSingleSeries) {
```
makafsal, line 411: In all handlers, read `dataGroups` it live inside each handler from `this.model` instead of capturing it as a closure variable
```suggestion
				if (self.model.getDataGroups().length === 1) {
```

makafsal (changes requested):

packages/core/src/components/essentials/legend.ts
Triparna-146: Addressed — removed the isSingleSeries closure variable entirely. Single-series protection is now enforced at registration time (which method gets called), so no handler needs to read getDataGroups() at all.

packages/core/src/components/essentials/legend.ts
Triparna-146: Done — extracted addTooltipListeners() for mouseover/mouseout/keyup and updated the call site in render() to call addEventListeners() for multi-series clickable charts and addTooltipListeners() otherwise.

packages/core/src/components/essentials/legend.ts
Triparna-146: Removed — the guard is no longer needed since click is only registered inside addEventListeners(), which is only called for multi-series charts.

packages/core/src/components/essentials/legend.ts
Triparna-146: Removed — same reason, keydown is only registered in addEventListeners().

Triparna-146:
@amal-k-joy @makafsal — all review comments addressed, please re-review when you get a chance. Thanks!

PR #2127, opened 2026-08-07
fix(ruler): remove tooltip lag for large data sets

### LINKED ISSUE #1107: Line Chart Tooltips lag with > 100 points (due to debounce)
#### I am submitting a...
- [ ] Design defect

#### charts version:
0.41.80

### Issue description
Tooltip lags with > 100 points on line chart when "points" (the circles on the line at each datapoint) are disabled.  When points are enabled the tooltip appears instantly on mouse over of point.  When points are disabled, they only appear immediately when 100 or less points are displayed. 

Create line chart with > 100 data points but with "points" (the circles) display disabled.

#### Current behavior
ruler.js  debounceWithD3MousePosition causes a lag (the debounce time for me is ~600 ms).  This gives a feeling that the chart is lagging when moving the mouse.   Commenting out the "if (displayData.length > 100) ... " block results in instant tooltip updates like I expect.

#### Expected behavior
Tooltips should still appear immediately as they do when "points" are displayed.

222jkm:
Fixes #1107

### Updates

- Remove the data-count debounce from the ruler.
- Index data and chart symbols once per render.
- Process the latest pointer position once per animation frame.
- Add ruler tests.

### Checks

- `yarn workspace @carbon/charts test --run src/components/axes/ruler.spec.ts`
- `yarn workspace @carbon/charts build`
- Changed-file ESLint: no errors
- 8,991 symbols, 50 moves: about 3.5 s to 64 ms

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,12 @@
 All notable changes to this project will be documented in this file. See
 [Conventional Commits](https://conventionalcommits.org) for commit guidelines.
 
+## [Unreleased]
+
+### Fixed
+
+- **Ruler**: Removed tooltip lag on charts with more than 100 data points.
+
 ## 1.27.18 (2026-07-31)
 
 **Note:** Version bump only for package @carbon/charts-monorepo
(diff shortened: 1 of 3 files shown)


222jkm:

222jkm:
recheck

PR #2131, opened 2026-08-19
fix(toolbar): move control ARIA onto the button, resolving nested-interactive

### LINKED ISSUE #2130: [Accessibility]: Toolbar controls fail WCAG 4.1.2 (nested-interactive)
### Application/Team

Not IBM, external consumer

### Environment

Automated testing tool and ruleset, Browser

### What happened? What did you expect to happen instead?

Toolbar controls will render a `<div role="button">` wrapping the real `<button>` element. Nested button roles are a WCAG2 SC 4.1.2 failure.

Axe DevTools (and axe-core running in automated tests), ARC Toolkit, Firefox Accessibility inspector all agree, with slightly different wording:

- **axe-core 4.12.1**: `nested-interactive`, impact: serious (["Interactive controls must not be nested"](https://dequeuniversity.com/rules/axe/4.10/nested-interactive))
- **ARC Toolkit**: "A focusable element was found within an ARIA role that defines its children as presentational" (ARC directly refer to https://www.w3.org/TR/WCAG21/#name-role-value)

Since `toolbar.enabled` defaults to `true` for all charts, this is reproducible in the docs as well https://carbon-design-system.github.io/carbon-charts/ (see screenshot)

Image

------------

#1761 fixed #1724 by adding `role="button"` to the wrapper. That made the pre-existing `aria-label` **valid**, but introduced the nested button role.

I think the wrapper div `toolbar-control` carries ARIA it should not carry. It is not the interactive element; the inner `<button>` element is (which also, already has `aria-label`. The wrapper has no `tabindex`, no event handlers, and no styling is targetting `.toolbar-control`.

**Suggested fix:** Drop `role` and `aria-label` from the wrapper div `toolbar-control`, and move `aria-disabled` onto the inner `<button>` element. I can open a PR with a proposal, if this looks right.

### What WCAG 2.1 checkpoint does the issue violate?

4.1.2 Name, Role, Value (EN 301 549 §9.4.1.2 and W3C ACT rule 307n5z)

### Version

`@carbon/charts-svelte@1.27.17`

### Data & options used

```js

```

### Relevant log output

```js

```

### StackBlitz repro

_No response_

### What priority level would this issue be in your opinion?

P1

henriksonnergard:
Closes #2130 

### Updates

- Fixes the `nested-interactive` violation on chart toolbar controls
- Removes `role="button"`, `aria-label` and `aria-disabled` from the `.toolbar-control` wrapper `<div>`
- Adds `aria-disabled` onto the inner `<button>`

**Before change** (note the nested button roles in the accessibility tree):
Screenshot 2026-08-19 at 15 50 29

**After change** (no visual change, mouse and keyboard interactivity as before):
Screenshot 2026-08-19 at 15 49 53

### DIFF
--- a/packages/core/src/components/axes/toolbar.ts
+++ b/packages/core/src/components/axes/toolbar.ts
@@ -83,9 +83,6 @@ export class Toolbar extends Component {
 			enteringToolbarControls
 				.merge(toolbarControls as any)
 				.classed('disabled', (d: any) => d.shouldBeDisabled())
-				.attr('role', 'button')
-				.attr('aria-disabled', (d: any) => d.shouldBeDisabled())
-				.attr('aria-label', (d: any) => d.title)
 				.html((d: any) => {
 					return `
 			<button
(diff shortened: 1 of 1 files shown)


henriksonnergard:

